PR #16, opened 2015-12-01
onboarding slider [stage-1]

ghislaineguerin:
@alex-deaconu @ezequielc could you review. thanks!

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -19,7 +19,7 @@
     "angular-mocks": "~1.3.0",
     "common-header": "https://github.com/Rise-Vision/common-header.git#v1.6.3",
     "rv-common-svg": "https://github.com/Rise-Vision/common-svg.git",
-    "rv-common-style": "https://github.com/Rise-Vision/common-style.git#1.3.21",
+    "rv-common-style": "https://github.com/Rise-Vision/common-style.git#v1.3.35",
     "rv-common-app-components": "https://github.com/Rise-Vision/common-app-components.git"
   },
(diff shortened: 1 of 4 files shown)


ezequielc:
@ghislaineguerin :+1:

alex-deaconu:
@ezequielc are we going to allow the inline script and the `$(document).ready` in a partial? Thanks.

ghislaineguerin:
@alex-deaconu where should I move that too?

ezequielc:
@alex-deaconu As it is an experiment, and the code would be reworked I wasn't strict. But I agree, if it won't be too complicated, we should make it better.

ezequielc:
@ghislaineguerin maybe you can use https://github.com/jonahbron/angular-owl-carousel.

alex-deaconu:
@ezequielc @ghislaineguerin Thanks for looking into this. I'm wondering what the best course of action is here. Maybe we need to have a development step for this experiment? Thoughts? Thanks.

ezequielc:
@alex-deaconu @ghislaineguerin  Yes, I think it will be much faster if we do that. Cheers

ghislaineguerin:
@alex-deaconu @ezequielc I'll move the card to the sprint backlog?

ghislaineguerin:
@alex-deaconu @ezequielc can you review this please? thanks!

ezequielc:
@ghislaineguerin Looks good. :+1: Cheers

PR #21, opened 2015-12-09
Merging Displays and Editor code

alex-deaconu:
Unit and e2e tests

[stage-2]


ezequielc:
@alex-deaconu I was playing with the UI and I found an issue. It is not opening Priority Support nor Send Us a Note.
Also, as a suggestions, I think we could move editor as the first App in the menu (after Home), just like RVA.
Cheers

alex-deaconu:
@ezequielc it seems those links don't work on staging. The reason is: http://screencast.com/t/t1AvmuYoS. This is a CH issue and I don't think it's related to the PR.

Thanks.

ezequielc:
@alex-deaconu Yes, I think you are right. I checked https://apps-stage-1.risevision.com/ and it is working, but maybe it is using a different config. Other apps also don't work on staging. Thanks

alex-deaconu:
@ezequielc moved the Editor link to the second spot as suggested. Thanks.

ezequielc:
@alex-deaconu Looks great. Good job! :+1: 
We talk tomorrow about deploying it. Thanks

PR #31, opened 2015-12-22
Render Text Widget

ezequielc:
@alex-deaconu Please review.

I modified stage config so we can add Text widget to presentations. We should really update test data.
Thanks!

### DIFF
--- a/web/scripts/editor/services/svc-widget-renderer.js
+++ b/web/scripts/editor/services/svc-widget-renderer.js
@@ -1,0 +1,118 @@
+'use strict';
+
+angular.module('risevision.editor.services')
+  .value('IFRAME_PREFIX', 'if_')
+  .value('RENDER_WIDGETS', [
+    'ba0da120-7c67-437f-9caf-73585bd30c74',
+    '64cc543c-c2c6-49ab-a4e9-40ceba48a253'
+  ])
+  .factory('widgetRenderer', ['gadgetsApi', '$window', 'IFRAME_PREFIX',
+    'RENDER_WIDGETS', 'userState',
+    function (gadgetsApi, $window, IFRAME_PREFIX, RENDER_WIDGETS, userState) {
+      var factory = {};
+
+      factory._placeholders = {};
+
+      var _isRenderingAllowed = function(objectReference) {
+        return RENDER_WIDGETS.indexOf(objectReference) !== -1;
+      }
+
+      factory.register = function (placeholder, element) {
+        if (placeholder.items && placeholder.items[0] && _isRenderingAllowed(placeholder.items[0].objectReference)) {
+          factory._placeholders[placeholder.id] = placeholder;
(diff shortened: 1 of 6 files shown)


alex-deaconu:
@ezequielc I've updated the Text widget product code such that it can be added/modified in the Test environment. Can you please test it out and let me know if it works? Thanks!

web/scripts/editor/services/svc-widget-renderer.js
```diff
@@ -0,0 +1,113 @@
+'use strict';
+
+angular.module('risevision.editor.services')
+  .value('IFRAME_PREFIX', 'if_')
+  .value('TEXT_WIDGET_ID', 'ba0da120-7c67-437f-9caf-73585bd30c74')
```
alex-deaconu: @ezequielc I'm assuming this would have to be a configuration setting. The Text widget Id for the test server is `64cc543c-c2c6-49ab-a4e9-40ceba48a253`. Thanks.

web/scripts/editor/services/svc-widget-renderer.js
```diff
@@ -0,0 +1,113 @@
...
+        delete factory._placeholders[placeholder.id];
+        var frameName = IFRAME_PREFIX + placeholder.id;
+        gadgetsApi.rpc.removeReceiver(frameName);
+        element.find('#' + frameName).remove();
+      };
+
+      factory.notifyResize = function (placeholder, element) {
+        if (factory._placeholders[placeholder.id]) {
+          factory.unregister(placeholder, element);
+          factory.register(placeholder, element);
```
alex-deaconu: @ezequielc curious, can't we just re-size the iframe without re-rendering the content? Thanks.

ezequielc:
@alex-deaconu it works locally. I had to change the widget id I was using, so I will use a list of allowed widget ids instead of a single value. This will be reused when rendering other widgets too.

alex-deaconu:
@ezequielc a list of IDs works as well (I just suggested using a config value in the code). 

The PR looks good, just a couple of questions and I will do another review when the IDs list is implemented. 

Thanks.

web/scripts/editor/services/svc-widget-renderer.js
ezequielc: @alex-deaconu I used a list of values as it will be reused when rendering other widgets. Thanks

web/scripts/editor/services/svc-widget-renderer.js
ezequielc: @alex-deaconu The iframe has 100% size but the widget receives width and height from url and it sets div size from that. So we need to reload in order to give the new size to the widget. Cheers!

ezequielc:
@alex-deaconu Changes are ready. Please review. Cheers

alex-deaconu:
@ezequielc looks good :+1: .Thanks!

PR #50, opened 2016-01-15
Redirect to original request after signing in

ezequielc:


ezequielc:
@alex-deaconu I changed the implementation to keep the same url when driving user to the sign in page.
Instead of having a landing page for each app, every app will now redirect to the unauthorized/sign in page. I did that change as it is required for the "Launcher page for non sign in users" card and to avoid rewriting e2e tests now that would end up being removed for that card.
Please review. Thanks!

alex-deaconu:
@ezequielc works very nicely. I haven't been able to find any issues.

One thing, I think you can remove the old states here:
https://github.com/Rise-Vision/rise-vision-apps/blob/feature/login-redirect/web/scripts/app.js#L381.

Otherwise, looks good :+1: .

Thanks!

web/scripts/app.js
```diff
@@ -378,6 +371,7 @@ angular.module('risevision.apps', [
   ])
   .run(['$rootScope', '$state', 'userState',
     function ($rootScope, $state, userState) {
+
       $rootScope.$on('risevision.user.signedOut', function () {
         $state.go('apps.launcher.home');
```
alex-deaconu: @ezequielc should we change this to `$state.go('apps.launcher.unauthorized');`? Thoughts? Thanks.

ezequielc:
Good point @alex-deaconu . I will remove them. Thanks!

web/scripts/app.js
ezequielc: @alex-deaconu I think we can leave as it is as it will redirect to '/' after signing out. To me, it makes more sense than to keep current url. Thoughts? Thanks.

alex-deaconu: @ezequielc I'm good with that. Thanks.

ezequielc:
@alex-deaconu I added `apps.displays.alerts` to the check. Please review. Thanks

alex-deaconu:
@ezequielc looks good :+1: , thanks for fixing. Cheers!

PR #56, opened 2016-01-22
Improving Browser Upgrade settings

ezequielc:
- Only show Browser Upgrade field for Displays that are not 'Auto Upgrade'
- 'Current' and 'Previous' options are changed to 'User Managed'
- Changed field labels

### DIFF
--- a/test/unit/displays/controllers/ctr-display-details.tests.js
+++ b/test/unit/displays/controllers/ctr-display-details.tests.js
@@ -212,4 +212,25 @@ describe('controller: display details', function() {
     });
   });
 
+  describe('browserUpgradeMode:',function(){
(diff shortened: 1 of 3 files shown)


alex-deaconu:
@ezequielc looks good :+1: . 

Actually was curious and did a BQ search for the `browserUpgradeMode` in the Core and there are no Displays with the options `2` or `3` selected. About 16% of Displays have `1` selected.

Thanks.

alex-deaconu:
Correction 1.6% of Displays have `1` selected. Thanks.

ezequielc:
@alex-deaconu Only 1.6%? That's good to know. It might simplify a future upgrade. I will write in the card to let Byron know about it. Thanks!

PR #59, opened 2016-01-25
Temporarily hide Schedule Item PUD field

alex-deaconu:
The field will only show if the option is checked; which it is not since this was never deployed. However, this keeps backwards compatibility for stage where the field may have been checked at one point.

[stage-2]


alex-deaconu:
Build took about an hour to finally complete correctly.

ezequielc:
@alex-deaconu :+1:

PR #61, opened 2016-01-27
Show login page for non signed in users

ezequielc:
- CH is hidden when displaying login page
- Sign Up route does not open sign up modal as the same content is shown in login page
- Fix E2E tests to sign in using login page instead of CH


test/e2e/displays/cases/alerts.js
```diff
@@ -1,6 +1,7 @@
 'use strict';
 var expect = require('rv-common-e2e').expect;
 var HomePage = require('./../../launcher/pages/homepage.js');
```
alex-deaconu: @ezequielc are the `**pages/homepage.js` pages, for this and other apps, relevant anymore. I believe the only thing needed is the `get()` function. Should we clean those up? Thanks.

web/scripts/app.js
```diff
@@ -73,14 +72,21 @@ angular.module('risevision.apps', [
...
+          ]
+        }
       })
 
       .state('apps.launcher.signup', {
         url: '/signup',
         templateProvider: ['$templateCache', function ($templateCache) {
           return $templateCache.get(
-            'partials/launcher/app-launcher.html');
+            'partials/launcher/login.html');
```
alex-deaconu: @ezequielc this state is a bit wonky: https://apps-stage-0.risevision.com/signup. Also, I'm wondering if its still needed anymore, can't we just default to `apps.launcher.home`? Thanks.

web/scripts/app.js
```diff
@@ -89,7 +95,7 @@ angular.module('risevision.apps', [
         url: '/signin',
         templateProvider: ['$templateCache', function ($templateCache) {
           return $templateCache.get(
-            'partials/launcher/app-launcher.html');
+            'partials/launcher/login.html');
```
alex-deaconu: @ezequielc this is a transition state. When I try to access it https://apps-stage-0.risevision.com/signin, it's showing `login.html` page w/ CH attached. I'm wondering if we can fix that. 

My suggestion for fixing this is above in `index.html`. 

Thanks.

web/index.html
```diff
@@ -54,7 +54,7 @@
 <body>
 
   <!-- Common Header -->
-  <common-header nav-options="navOptions" nav-selected="navSelected" ng-controller="AppCtrl"></common-header>
+  <common-header nav-options="navOptions" nav-selected="navSelected" ng-controller="AppCtrl" ng-hide="hideCH"></common-header>
```
alex-deaconu: @ezequielc I'm wondering if we can do `ng-show="showCh"` so that the CH is hidden by default. This may prevent issues below where the CH is showing together with `login.html` while the page is resolving.

Thanks.

web/scripts/app.js
```diff
@@ -62,9 +62,8 @@ angular.module('risevision.apps', [
       .state('apps.launcher.unauthorized', {
         templateProvider: ['$templateCache', function ($templateCache) {
           return $templateCache.get(
-            'partials/launcher/app-launcher.html');
-        }],
-        controller: 'HomeCtrl'
+            'partials/launcher/login.html');
+        }]
```
alex-deaconu: @ezequielc do we still need this state? I think we can use `apps.launcher.home` for the same purpose. Thanks.

alex-deaconu:
@ezequielc reviewed. 

Noted an issue with the CH and `login.html` page showing at the same time while some pages are loading.
Also related, I'm suggesting some cleanup of the states that show the new `login.html` page. I think we have too many states right now, and I'm wondering if we can simplify it a bit.
Finally, a cleanup issue with e2e pages.

Otherwise, looks good.

Thanks.

test/e2e/displays/cases/alerts.js
ezequielc: @alex-deaconu Agreed. Any suggestions on where I should move `get()` to? Thanks!

web/scripts/app.js
ezequielc: @alex-deaconu I think  we need to keep the url as it is being used outside Apps, but we can simplify and just redirect it to `apps.launcher.home`. Thanks

web/scripts/app.js
ezequielc: @alex-deaconu Thanks! I will fix it.

ezequielc:
Thanks for your comments @alex-deaconu . I will work on them. Cheers!

test/e2e/displays/cases/alerts.js
alex-deaconu: @ezequielc good question.

One place I would put it is in the `*.cases.js` files. This way you can move the `before` statements w/ `homepage.get();` to that file for each set of cases. 

However, that may be too much work for this PR. Another place I was thinking maybe the `launcher/pages/homepage.js`? 

Or you can just leave it in place, and remove all the other un-used variables from the homepages.

Thanks.

web/scripts/app.js
alex-deaconu: @ezequielc yup, we cannot get rid of `/signup` (or `/signin`) for that reason. Why not just remove `apps.launcher.home` and just redirect to `/signup` instead? Otherwise, I'm good with leaving as is, as long as the header issue is fixed. Thanks.

test/e2e/displays/cases/alerts.js
ezequielc: @alex-deaconu Ok, I'll see what can be done. Thanks

web/scripts/app.js
ezequielc: @alex-deaconu I think we will need `home` for the url as well. I will try some options. Cheers

ezequielc:
@alex-deaconu I refactored E2E tests to use launcher home only. I also fixed the issue and simplified a bit signup and signin states. Please review. Thanks!

alex-deaconu:
@ezequielc looks good. Thanks for the extra work. :+1:

ezequielc:
Thanks @alex-deaconu !

PR #71, opened 2016-02-03
Fix Issue #46: Modal for timeline doesn't close when clicking outside

ezequielc:
@alex-deaconu I had staged at stage-0 but it is not available anymore. I am staging at stage-1. 
I will need to update CH and common-app-components version once they are merged.
Please review. Thanks


alex-deaconu:
@ezequielc I don't think we need a PR for this. Just rebuild master and we should be good. Thanks!

ezequielc:
@alex-deaconu Agreed, just making sure it works. Thanks!

PR #79, opened 2016-02-11
Refactored event tracking

alex-deaconu:
Editor:
Removed tracking from partials; added to factory
Single tracked event for all actions
Renamed addFromTemplate to addPresentationModal
Refactored editorFactory initialization to distinguish from newPresentation

Remove isTemplate flag when a Template is copied

Open player download links in new tab

Schedules:
Removed duplicate event from controllers
Add Schedule event in factory (refactor init)

Removed duplicate Display tracking events

[stage-4]

### DIFF
--- a/test/unit/displays/controllers/ctr-display-add.tests.js
+++ b/test/unit/displays/controllers/ctr-display-add.tests.js
@@ -116,7 +116,6 @@ describe('controller: display add', function() {
     $scope.displayDetails.$valid = true;
     $scope.display = {id:123};
     $scope.save();
-    expect(trackerCalled).to.equal('Save Display');
     expect($scope.savingDisplay).to.be.true;
     $scope.$digest();
     $loadingStartSpy.should.have.been.calledWith('displays-loader');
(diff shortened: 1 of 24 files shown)


alex-deaconu:
@ezequielc despite our conversation from yesterday, I chose to track events (save, delete, etc) after the API is called rather than before. The only reason for this is when adding a new item, the API returns the item's ID, which is added to the tracker. If you feel that's not a good enough reason I can move it back out to track before the API call (and track errors as discussed).

Displays has very little refactoring due to the fact that we don't use a `Factory` there. Once we add that in we can move the events inside.

One more thing. Do you think this event is needed for the dropdown: https://github.com/Rise-Vision/rise-vision-apps/blob/master/web/partials/schedules/schedule-fields.html#L36.

Otherwise, please review. Thanks!

ezequielc:
@alex-deaconu I am good with tracking after API calls, it can be valuable to track item's ID. 
About the dropdown, I will defer to @ghislaineguerin as she originally requested it. Maybe it is not used anymore.  @ghislaineguerin Can we remove it? 
I am reviewing the code. Cheers!

ezequielc:
@alex-deaconu Looks good :+1: Thanks!

alex-deaconu:
@ezequielc spoke with @ghislaineguerin and she said the extra event is not necessary. Will remove in a second PR. Thanks!

PR #80, opened 2016-02-11
Shared templates

ezequielc:


web/scripts/editor/controllers/ctr-shared-templates-modal.js
```diff
@@ -0,0 +1,39 @@
+'use strict';
+angular.module('risevision.editor.controllers')
+  .controller('SharedTemplatesModalController', ['$scope',
+    'ScrollingListService', 'template', 'editorFactory', '$loading',
+    '$filter', '$modalInstance',
+    function ($scope, ScrollingListService, template, editorFactory,
+      $loading, $filter, $modalInstance) {
+      $scope.search = {
+        sortBy: 'publish',
```
ezequielc: @alex-deaconu I am getting a Core error when sorting by `name` (Column not Found). Should we fix it now or queue a card for Debt Reduction epic?

alex-deaconu: @ezequielc curious, did you try to leave the `sortBy` field blank? I believe the default sort is by `name`, so it may work that way. This way we can fix later (in Debt Reduction). 

Otherwise, I would just add a card and ask the team what they think the priority is. Personally, I would queue it for this release; the list looks awkward when sorted by `publish`.

Thanks.

ezequielc: @alex-deaconu Yes, I tried with blank field and got the same error. It works with any of these fields except name: https://github.com/Rise-Vision/core/blob/master/src/main/java/com/risevision/core/data/Presentation.java#L54

I will bring this issue to stand up tomorrow morning to the team.

Thanks

ezequielc:

alex-deaconu:
@ezequielc looks good :+1: . Don't forget to update `bower.json` once dependencies have been merged. Thanks.

ezequielc:
Thanks @alex-deaconu

PR #89, opened 2016-02-18
Fix Gadgets 'reloading' page.

ezequielc:
They were actually opening a full page iframe for their settings, but as they don't have settings page, it was showing a new Editor page in the frame.

To fix, we should not show Gadget settings.

### DIFF
--- a/web/scripts/editor/controllers/ctr-playlist-item-modal.js
+++ b/web/scripts/editor/controllers/ctr-playlist-item-modal.js
@@ -17,15 +17,14 @@ angular.module('risevision.editor.controllers')
       if (!item.objectReference && item.settingsUrl) {
         $scope.widgetName = item.name;
       } else {
-        if (item.objectReference && (item.type === 'widget' || item.type ===
-            'gadget')) {
+        if (item.objectReference && item.type === 'widget') {
           gadgetFactory.getGadget(item.objectReference).then(function (gadget) {
             $scope.widgetName = gadget.name;
           });
         }
       }
 
-      if (showWidgetModal) {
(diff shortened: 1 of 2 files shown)


ezequielc:
@alex-deaconu You can validate using the RSS gadget in Stage environment. Cheers

web/scripts/editor/controllers/ctr-playlist-item-modal.js
```diff
@@ -17,15 +17,14 @@ angular.module('risevision.editor.controllers')
...
-            'gadget')) {
+        if (item.objectReference && item.type === 'widget') {
           gadgetFactory.getGadget(item.objectReference).then(function (gadget) {
             $scope.widgetName = gadget.name;
           });
         }
       }
 
-      if (showWidgetModal) {
+      if (showWidgetModal && item.type !== 'gadget' ) {
```
alex-deaconu: @ezequielc I would suggest we check if `item.type === 'widget'`, in case we ever encounter a 3rd type of item. Thoughts? Thanks.

alex-deaconu:
@ezequielc looks good :+1: . Just one small suggestion. Thanks.

web/scripts/editor/controllers/ctr-playlist-item-modal.js
ezequielc: @alex-deaconu Makes sense. I will update. Thanks

ezequielc:
@alex-deaconu I implemented your suggestion. Please review. Thanks!

alex-deaconu:
@ezequielc thanks for fixing. :+1:

PR #91, opened 2016-02-18
Moved Template purchase check to Core

alex-deaconu:
No longer check Widget Authorization API
Removed all related code

Presentation GET API now verifies Subscription Status
It uses the companyId which is now passed to the Core

Depending on API response, copy the Template or show notification
Retrieve Product Details if not available for the notification UX

[stage-4]


alex-deaconu:
Requires specific Core version: http://template-subscription.rvacore-test.appspot.com/.

PR #96, opened 2016-02-24
Text Widget ID was changed in Production

ezequielc:
@rodrigopavezi Please review. I left the old ID in the list to allow rendering the widgets of existing/old Presentations. The ID is stored in Playlist Item `objectReference`.

Thanks!


rodrigopavezi:
@ezequielc looks good :+1: cheers

PR #99, opened 2016-03-01
Fix #84 - When saving, &quot; gets replaced by " in Core

### LINKED ISSUE #84: Pasting a spreadsheet into the text widget causes presentation to go blank and inaccessible in editor.risevision.com
**How to recreate**
1. Using editor.risevision.com create a new presentation
2. Within that presentation, add 3 placeholders.
3. In two of those placeholders, add content of any type (I've used images for simplicity).
4. In the third placeholder, add the Text widget.
5. In a Google sheet that contains content, select a range of cells and copy.
6. Paste this into the Text Widget
7. Save presentation

**What happens**
The presentation goes completely blank, and the presentation is inaccessible from the Presentations screen, clicking the Presentation does not open anything.

**What should happen**
The presentation should save, all content should remain without being deleted, and the presentation should be fully accessible and editable. 

**Sample presentation**
http://bit.ly/1R8FqY5

**Video overview**
http://www.screencast.com/t/SdCy4n0n4fX5

**Notes**
- I am unable to replicate this in the RVA editor, I was able to paste into a Text Widget and save the presentation with no trouble.
- It appears this is caused by pasting a range of cells from a Spreadsheet, pasting from a Doc, website, or text editor does not replicate this issue.

ezequielc:
…, invalidating JS code.

@rodrigopavezi The issue occur when copying and pasting Google Spreadsheet content to a Text Widget and saving the Presentation. 
Pasted content contains `&quot;`, which gets replaced by `"` in Core, breaking JS code of the Presentation.
Core does that due to a call to `unescapeHtml()`  to fix layout decoding: https://github.com/Rise-Vision/core/commit/14361b8bf28905ad7422156a1e057b4a68237587 

The simplest solution was to replace the `&quot;` in the editor and escape it before sending to Core.
Please review. Thanks!


rodrigopavezi:
@ezequielc looks good :+1: cheers

PR #103, opened 2016-03-07
Apps Funnel BigQuery logging

ezequielc:

### DIFF
--- a/web/scripts/displays/services/svc-display-tracker.js
+++ b/web/scripts/displays/services/svc-display-tracker.js
@@ -2,7 +2,8 @@
 
 angular.module('risevision.displays.services')
   .factory('displayTracker', ['userState', 'segmentAnalytics',
-    function (userState, segmentAnalytics) {
+    'bigQueryLogging',
+    function (userState, segmentAnalytics, bigQueryLogging) {
       return function (eventName, displayId, displayName, downloadType) {
         if (eventName) {
           segmentAnalytics.track(eventName, {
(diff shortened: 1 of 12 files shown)


web/scripts/displays/services/svc-display-tracker.js
```diff
@@ -2,7 +2,8 @@
 
 angular.module('risevision.displays.services')
   .factory('displayTracker', ['userState', 'segmentAnalytics',
-    function (userState, segmentAnalytics) {
+    'bigQueryLogging',
+    function (userState, segmentAnalytics, bigQueryLogging) {
       return function (eventName, displayId, displayName, downloadType) {
```
alex-deaconu: @ezequielc why not add an extra optional parameter which will enable logging to BQ? I'd say it's much cleaner than hardcoding the `eventName` below. Thanks.

alex-deaconu:
@ezequielc looks good. Just one suggestion in the code. Thanks.

web/scripts/displays/services/svc-display-tracker.js
ezequielc: @alex-deaconu It is a good idea, but then we will need to modify every call we want to track to BQ. This way we have a single place to enable/disable BQ logging. I was actually thinking about having a list of tracked events instead of hardcoding. But maybe we don't need that :) Thoughts? Thanks

alex-deaconu: @ezequielc hmm good point. I think we can leave it as is, and move to constants once/if we add extra events. Thanks.

alex-deaconu:
@ezequielc :+1:

PR #105, opened 2016-03-10
Log Schedules and Displays creation to BQ

ezequielc:
@alex-deaconu Please review.
I also tried to add a trick to fix e2e tests failing for `presentation-add.js`. It seems to have helped.


alex-deaconu:
@ezequielc thanks for fixing. PR looks good :+1: .

PR #107, opened 2016-03-16
Fix e2e test

ezequielc:
@alex-deaconu I run the build three times without failures, so I assume it worked :) 
Please review. Thanks


alex-deaconu:
@ezequielc looks good :+1: . Thanks for fixing. Fingers crossed.

PR #116, opened 2016-03-23
Improved unit tests & some refactoring

alex-deaconu:

### DIFF
--- a/test/unit/storage/services/svc-file-selector-factory.tests.js
+++ b/test/unit/storage/services/svc-file-selector-factory.tests.js
@@ -8,15 +8,22 @@ describe('service: fileSelectorFactory:', function() {
     });
 
     $provide.service('gadgetsApi',function(){
-      return {
+      return gadgetsApi = {
         rpc: {
           call: function() {}
         }
       };
     });
-
+    $provide.service('userState', function() {
+      return {
+        getSelectedCompanyId: function() {
+          return 'companyId';
+        },
+        _restoreState: function() {}
+      };
+    });
   }));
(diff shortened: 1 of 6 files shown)


test/unit/storage/services/svc-file-selector-factory.tests.js
```diff
@@ -29,45 +36,134 @@ describe('service: fileSelectorFactory:', function() {
...
+    it('should not do anything if a file is selected', function() {
+      fileSelectorFactory.folderSelect(filesFactory.filesDetails.files[1]);
+      
+      expect(filesFactory.filesDetails.files[1].isChecked).to.be.false;
+      expect(filesFactory.filesDetails.checkedCount).to.be.equal(0);
+      expect(filesFactory.filesDetails.folderCheckedCount).to.be.equal(0);
+      expect(filesFactory.filesDetails.checkedItemsCount).to.be.equal(0);
+    });
+
+    it('should not do anything if file(s) selector', function() {
```
ezequielc: @alex-deaconu Curious, what is the difference from this and the previous test? Thanks

ezequielc: @alex-deaconu Nevermind, I got. It is a folder, right? Thanks

alex-deaconu: @ezequielc yup. First one selects a file which should not be selected because it checks for folders. Second test selects a folder, but the selector type is 'single-file' by default, so it will not select it.

Good question though, I will update the test descriptions as they are unclear. 

Thanks.

ezequielc:
@alex-deaconu Code look good. :+1: Thanks!

PR #117, opened 2016-03-23
File upload

ezequielc:
@alex-deaconu Please review. 

I am not totally happy with the code, I would like to refactor and add more unit tests, but I think we are taking too long on this task. Maybe we should add a card for refactoring in our Debt Reduction epic.

Thanks!

### DIFF
--- a/test/e2e/storage/storage.cases.js
+++ b/test/e2e/storage/storage.cases.js
@@ -1,3 +1,7 @@
 StorageSelectorScenarios = require('./cases/storage-selector.js');
+NewFolderModalScenarios = require('./cases/new-folder.js');
+UploadScenarios = require('./cases/upload.js');
 
 storageSelectorScenarios = new StorageSelectorScenarios();
+newFolderModalScenarios = new NewFolderModalScenarios();
+uploadScenarios = new UploadScenarios();
--- a/web/partials/storage/storage-modal.html
+++ b/web/partials/storage/storage-modal.html
@@ -21,14 +21,17 @@
 
 	  <div class="global-actions">
 	  	<button id="newFolderButton" class="btn btn-lg btn-default" ng-click="storageFactory.addFolder()" ng-disabled="isTrashFolder()">
-        <span translate="storage-client.new-folder"></span>
+        <span translate>storage-client.new-folder</span>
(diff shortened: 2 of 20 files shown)


test/e2e/storage/storage.cases.js
```diff
@@ -1,3 +1,7 @@
 StorageSelectorScenarios = require('./cases/storage-selector.js');
+NewFolderModalScenarios = require('./cases/new-folder.js');
```
alex-deaconu: @ezequielc I'm assuming this was missed in the New Folder PR? Thanks.

ezequielc: @alex-deaconu Exactly, I missed adding it to the commit. Cheers

web/partials/storage/storage-modal.html
```diff
@@ -21,14 +21,69 @@ <h1 id="title" translate="storage-client.select-{{storageFactory.selectorType}}"
...
-	      Upload
+	    <label id="uploadButton" for="upload-files" class="btn btn-lg btn-default" ng-hide="statusDetails.code!==200 || fileListStatus.code===404" ng-disabled="isTrashFolder()">
+        <span translate>storage-client.upload-files</span> 
 	      <i class="fa fa-cloud-upload fa-lg icon-right"></i> 
-	    </button>
+	    </label>
 	  </div>
+
+<div ng-controller="UploadController">
+  <div class="content-box animated fadeInDown upload-panel" ng-show="uploader.queue.length > 0">
```
alex-deaconu: @ezequielc would you be against splitting this out into its own partial? Just to keep the size of this file manageable. Thanks.

alex-deaconu:
@ezequielc looks good :+1: . Yes, I do see room for refactoring, so agreed re: card. However, I would keep that card in the Storage epic. Thanks.

web/partials/storage/storage-modal.html
ezequielc: @alex-deaconu Makes sense. I will split it. Thanks.

ezequielc:
@alex-deaconu I split the upload panel into its own partial. Please review. Thanks!

alex-deaconu:
@ezequielc looks good. Thanks for fixing. :+1:

Issue #112, opened 2016-03-19
Placeholder with CSS animations inside a Div breaks HTML view


sheadarlison:
I added the below code and some CSS animations to the HTML view of a presentation: 

```
<div style="width:815px;height:900px;left:560px;top:55px;z-index:3;position:absolute;box-shadow: 5px 10px 15px 5px rgba(0,0,0,0.4);overflow:hidden;>
<div id="Image-Slideshow" class="image" placeholder="true" style="width: 815px; height: 900px; z-index: 3; visibility: visible;"></div>
</div>
```

And now I can't access the HTML view any more. 

Presentation:
https://apps.risevision.com/editor/workspace/63f52f47-7990-444b-8e6f-ba27281ba275/?cid=7fa5ee92-7deb-450b-a8d5-e5ed648c575f

Preview:
http://preview.risevision.com/?type=presentation&id=63f52f47-7990-444b-8e6f-ba27281ba275&showui=false

![html-view-bug](https://cloud.githubusercontent.com/assets/9273546/13900490/60edd6ba-ede6-11e5-9f1f-4290025749f8.gif)

ezequielc:
@sheadarlison A fix has been released. Could you validate it is working? Thanks!

ezequielc:
Validated in the card https://trello.com/c/pIHTJqBo/1258-issue-112-placeholder-with-css-animations-inside-a-div-breaks-html-view-2

PR #121, opened 2016-03-24
Fix parsing issue when missing top or left

ezequielc:
Two issues when handling the sample code from https://github.com/Rise-Vision/rise-vision-apps/issues/112
- not handling missing position (top/left)
- not handling whitespace in parameter name

[stage-3]

### DIFF
--- a/web/scripts/editor/services/svc-presentation-parser.js
+++ b/web/scripts/editor/services/svc-presentation-parser.js
@@ -231,7 +231,7 @@ angular.module('risevision.editor.services')
         var tokens = htmlString.split(';');
         for (var x = 0; x < tokens.length; x++) {
           if (tokens[x].indexOf(':') !== -1) {
-            var param = tokens[x].split(':')[0];
+            var param = tokens[x].split(':')[0].trim();
             var value = tokens[x].substring(tokens[x].indexOf(':') + 1).trim();
 
             if (param.equalsIgnoreCase(widthParam)) {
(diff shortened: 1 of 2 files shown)


web/scripts/editor/services/svc-presentation-parser.js
```diff
@@ -477,10 +477,18 @@ angular.module('risevision.editor.services')
...
           placeholder.height, placeholder.heightUnits), styleString);
-        styleString = htmlParser.updateStyle(leftParam, htmlParser.getUnitString(
-          placeholder.left, placeholder.leftUnits), styleString);
-        styleString = htmlParser.updateStyle(topParam, htmlParser.getUnitString(
-          placeholder.top, placeholder.topUnits), styleString);
+        if (placeholder.left) {
+          styleString = htmlParser.updateStyle(leftParam, htmlParser.getUnitString(
+            placeholder.left, placeholder.leftUnits), styleString);
+        } else {
+          styleString = htmlParser.updateStyle(leftParam, '0px', styleString);
```
alex-deaconu: @ezequielc I don't think we should be adding `0px` by default here. If the user wants to use relative positioning/size or define it in a CSS class, this would overwrite it. Thoughts? Thanks.

web/scripts/editor/services/svc-presentation-parser.js
```diff
@@ -477,10 +477,18 @@ angular.module('risevision.editor.services')
           placeholder.width, placeholder.widthUnits), styleString);
```
alex-deaconu: @ezequielc curious, why not add the same conditions `if (placeholder.width) {...}` for width and height as well? Thanks.

web/scripts/editor/services/svc-presentation-parser.js
ezequielc: @alex-deaconu This is the same behavior I found in RVA. Also, if we leave it blank, we will have errors in placeholder settings. I could not think of any other simple solution. Suggestions? Thanks

web/scripts/editor/services/svc-presentation-parser.js
ezequielc: @alex-deaconu I wasn't sure about it, but if you don't see any problems as well I will add for both. Thanks

web/scripts/editor/services/svc-presentation-parser.js
alex-deaconu: @ezequielc hmm, looked and I can't find the same logic in the RVA code. Can you point me to it? Thanks.

web/scripts/editor/services/svc-presentation-parser.js
alex-deaconu: Nope, can't think of any issues that would come up. Thanks.

web/scripts/editor/services/svc-presentation-parser.js
ezequielc: @alex-deaconu I didn't look at the code, but if you paste a HTML without top or left:
```
<div style="width:815px;height:900px;left:560px;top:55px;z-index:3;position:absolute;box-shadow: 5px 10px 15px 5px rgba(0,0,0,0.4);overflow:hidden;">
<div id="Image-Slideshow" class="image" placeholder="true" style="width: 815px; height: 900px; z-index: 3; visibility: visible;"></div>
</div>
```
and go to Design and back to HTML, it will show 0px for both properties. Cheers

alex-deaconu: @ezequielc I guess it's probably from java `int` values defaulting to `0`. I think we're good with `0px` default. Thanks.

ezequielc: @alex-deaconu It might be the case. I will leave as `0px` then. Thanks!

ezequielc:
@alex-deaconu I added the same conditional to width, height and z-index in order to behave like RVA. Please review. Thanks

alex-deaconu:
:+1:

PR #126, opened 2016-03-31
E2E tests parallelism

ezequielc:
- Run E2E tests in parallel.
- Workaround to retrieve files from CCI command.

@alex-deaconu  Please review. Thanks

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -16,6 +16,7 @@ var gutil       = require("gulp-util");
 var rename      = require('gulp-rename');
 var runSequence = require('run-sequence');
 var factory     = require("widget-tester").gulpTaskFactory;
+var fs          = require('fs');
 
 //--------------------- Variables --------------------------------------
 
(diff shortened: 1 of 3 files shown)


gulpfile.js
```diff
@@ -236,13 +237,19 @@ gulp.task("server", factory.testServer({
...
 gulp.task("test:webdrive_update", factory.webdriveUpdate());
-gulp.task("test:e2e:core", ["test:webdrive_update"], factory.testE2EAngular({
+gulp.task("test:e2e:core", ["test:webdrive_update"],factory.testE2EAngular({
   browser: "chrome",
   loginUser: process.env.E2E_USER,
   loginPass: process.env.E2E_PASS,
-  testFiles: process.env.TEST_FILES
+  testFiles: function(){ 
+    try{
+      return JSON.parse(fs.readFileSync('/tmp/testFiles.txt').toString())
```
alex-deaconu: @ezequielc not sure if you've tried, but if I create a file named `/tmp/testFiles.txt` I could potentially use this feature locally too, to run specific tests? Thanks.

alex-deaconu:
@ezequielc looks good :+1: . Thanks.

gulpfile.js
ezequielc: @alex-deaconu I didn't try it, but it should work. Although it needs to be one of our `*.cases.js` file. It will not work with our individual test files because the way we structured them. Cheers!

alex-deaconu: Yes, something like `test/e2e/storage/storage.cases.js`. Cool!

PR #127, opened 2016-03-31
Fix #401 - Remove link does not work when uploading multiple files in Storage

ezequielc:


alex-deaconu:
:+1:

PR #131, opened 2016-04-04
Improving styles for Storage file selector modal.

ezequielc:
- Files List refactoring 
- Modal header for Storage file selector

### DIFF
--- a/web/partials/storage/storage-modal.html
+++ b/web/partials/storage/storage-modal.html
@@ -1,11 +1,13 @@
-
-<div class="modal-body storage-full" stop-event="touchend" id="storageSelectorModal" ng-controller="FilesListController">
-  <button type="button" id="closeButton" class="close" ng-click="dismiss()" aria-hidden="true">
+<div id="storageSelectorModal" ng-controller="FilesListController">
+<div class="modal-header">
+  <button type="button" id="closeButton" class="close" ng-click="dismiss()" data-dismiss="modal" aria-hidden="true">
     <i class="fa fa-times"></i>
   </button>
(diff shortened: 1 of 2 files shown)


web/partials/storage/storage-modal.html
```diff
@@ -13,12 +15,8 @@
...
 
 	<div class="padding-large-horizontal padding-large-bottom">
-	<div class="app-header half-top">
-	  <div class="title-block">
-	    <!-- App Title -->
-	    <h1 id="title" translate="storage-client.select-{{storageFactory.selectorType}}"></h1>
-    </div>
-
+	<div class="app-header">
+    <div class="title-block"></div>
```
alex-deaconu: @ezequielc I don't think this `div` is necessary anymore. And the one above it may not be as well. Can you double check? Thanks!

alex-deaconu:
@ezequielc just one comment, otherwise looks good. Thanks.

web/partials/storage/storage-modal.html
ezequielc: @alex-deaconu I tried to remove both of them but I was having style issues. `app-header` is necessary for dropdown and `title-block` fills the space in the div so that buttons are placed in the right side. I tried to use `pull-right` but they were not being placed correctly. We can ask Ghislaine or Adrian for help or leave like that for now. Thoughts? Thanks!

alex-deaconu: @ezequielc no worries. If the styling is finicky we can leave them in. I'll leave it up to you if you want to pursue it further. Thanks.

alex-deaconu:
@ezequielc :+1: . Thanks!

PR #136, opened 2016-04-11
Send only single file through RPC

alex-deaconu:
@ezequielc looks like RVA wants a single file, not an array.

Verify at http://1-08-035.rva-test.appspot.com/#/PRESENTATION_MANAGE. Will work on sub-company switching issue separately.

Please review. Thanks.


ezequielc:
@alex-deaconu :+1:

PR #141, opened 2016-04-15
common wait for presentation properties

alex-deaconu:
wait for transition


alex-deaconu:
@ezequielc finally a build that passed. This fixes the Presentation Properties modal which is responsible for most of the failures. I think we should work off this PR to fix other smaller issues.

Thanks.

ezequielc:
@alex-deaconu Agreed. :+1: Thanks!

PR #142, opened 2016-04-15
Fix download test

alex-deaconu:
Fix storage list spinner key
Unify loader, list and search into list page


ezequielc:
@alex-deaconu :+1:

PR #144, opened 2016-04-15
Storage empty state

ezequielc:
- Adding empty state for Storage modal
- Fixed missing svg icon
- Fixed missing Store status
- Fixed issue https://github.com/Rise-Vision/storage-client/issues/402


web/partials/storage/storage-modal.html
```diff
@@ -23,13 +23,33 @@ <h3 id="title" class="modal-title" translate="storage-client.select-{{storageFac
 
 	</div><!--app-header-->
 
-  <search-filter filter-config="filterConfig" search="search" do-search="search.doSearch" ng-show="!trialAvailable"></search-filter>
+  <div class="container ng-scope" id="fileinspector_noFiles" ng-show="statusDetails.code!==202 && !filesDetails.bucketExists && !trialAvailable && fileUploader.queue.length === 0">    
+    <div class="product-graphic">
+      <svg-icon p="riseStorage"></svg-icon>
```
alex-deaconu: @ezequielc there's a hover effect over the Storage icon. Is that intentional? I think that may be misleading to the user thinking there's an action associated with the icon. Can you please take a look? Thanks.

alex-deaconu:
@ezequielc couple of questions on the PR. Otherwise looks good. Thanks.

web/partials/storage/storage-modal.html
ezequielc: @alex-deaconu I investigated this issue and it is a CSS issue. I believe it is related to this line: https://github.com/Rise-Vision/common-style/blob/6404858d27a372d5a4ceb05fdf6f7a174d0a7630/src/core/utilities.scss#L64

It happens in the old iframe as well.
Should we ask UX for help, or add a new card to avoid context switching? Cheers!

alex-deaconu: @ezequielc we can add a new card or log an issue (which will add a card). Agreed re- not fixing this right now. Thanks.

alex-deaconu:
@ezequielc looks like I missed adding my second question. 

To clarify, the empty state is not added to "full screen" Storage. Is that intentional? Thanks.

ezequielc:
@alex-deaconu Yes, it is intentional, although it is debatable. 
Full screen Storage will show Trash, so it will not be really empty. 
This is the behavior of old Storage, so I kept the same, but we I am not convinced it is the best option. 
Cheers

alex-deaconu:
@ezequielc sounds good. Do you think we should add a card for this decision and the appropriate fix in our Debt Reduction epic? I think this call to action is useful in Full Screen storage as well. Thanks.

web/partials/storage/storage-modal.html
ezequielc: Thanks @alex-deaconu I added a new card for it https://trello.com/c/D15gz7R4/1335-storage-hover-empty-state-icon-changes-its-color. Cheers

alex-deaconu: @ezequielc thanks!

ezequielc:
@alex-deaconu I added a card: https://trello.com/c/xnGA8SYs/1336-storage-empty-state-for-storage-home

Are we good to merge? Thanks!

alex-deaconu:
@ezequielc yes, looks good to merge 👍 . Thanks!

PR #146, opened 2016-04-18
Fixing folder selection right after upload

alex-deaconu:
Fixes issue brought up here: https://trello.com/c/OpMms5DC/1093-dev-in-app-storage-file-selection-2.

When a new folder is added, set the `kind` property. Otherwise, when the URL is formed the new folder will be interpreted as a file: https://github.com/Rise-Vision/rise-vision-apps/blob/master/web/scripts/storage/services/svc-file-selector-factory.js#L90.


ezequielc:
@alex-deaconu Thanks for fixing :+1:

PR #150, opened 2016-04-20
Disable New Folder and Upload when not subscribed

alex-deaconu:
Also disable instead of ng-hide to prevent flashing buttons

[stage-4]


alex-deaconu:

ezequielc:
@alex-deaconu :+1: Thanks!

PR #153, opened 2016-04-21
Allow Storage use with Cancelled but Active Subscription

alex-deaconu:
Updated CH and subscription-status-component versions

Some cleanup

[stage-4]


alex-deaconu:
@ezequielc last piece in the series. 
Relies on updated CH, Subscription Status and Storage server.

Please review. Thanks.

ezequielc:
@alex-deaconu Lookg good :+1: Cheers

Issue #69, opened 2016-02-02
'Delete' key moves Editor back in browser history (behind modal)


alex-deaconu:
Also happens when there is no modal. Chrome/OSX

https://cloud.githubusercontent.com/assets/11826611/11129867/c850e618-894f-11e5-96ee-c06a9145b78f.gif

ezequielc:
@alex-deaconu Issue should be fixed. Could you validate and close? Thanks!

alex-deaconu:
@ezequielc works perfectly. Thanks for fixing!

PR #156, opened 2016-04-27
Fix issue #69 - Hide modals on state change

ezequielc:
[stage-2]

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -19,7 +19,7 @@
     "angular-bootstrap-colorpicker": "~3.0.19",
     "angular-ui-codemirror": "~0.3.0",
     "angular-vertilize": "~1.0.1",
-    "common-header": "https://github.com/Rise-Vision/common-header.git#v1.10.5",
+    "common-header": "https://github.com/Rise-Vision/common-header.git#v1.10.6",
     "component-storage-selector": "https://github.com/Rise-Vision/component-storage-selector.git",
     "rv-common-i18n": "https://github.com/Rise-Vision/common-i18n.git",
     "rv-common-style": "https://github.com/Rise-Vision/common-style.git#v1.3.61",
--- a/test/e2e/storage/cases/dismiss-modal.js
+++ b/test/e2e/storage/cases/dismiss-modal.js
@@ -1,0 +1,54 @@
+'use strict';
+var expect = require('rv-common-e2e').expect;
+var helper = require('rv-common-e2e').helper;
+var StorageSelectorModalPage = require('./../pages/storageSelectorModalPage.js');
+var NewFolderModalPage = require('./../pages/newFolderModalPage.js');
+var HomePage = require('./../../launcher/pages/homepage.js');
+var CommonHeaderPage = require('rv-common-e2e').commonHeaderPage;
+var StorageHelper = require('./../pages/helper.js');
+var StorageHomePage = require('./../pages/storageHomePage.js');
+var FilesListPage = require('./../pages/filesListPage.js');
+
+
+var DismissModalScenarios = function() {
+
+  browser.driver.manage().window().setSize(1400, 900);
+  describe("As a user, I would like modals to be dismissed when going back in history", function () {
+    var storageSelectorModalPage = new StorageSelectorModalPage();
+    var newFolderModalPage = new NewFolderModalPage();
+    var homePage = new HomePage();
+    var commonHeaderPage = new CommonHeaderPage();
+    var storageHomePage = new StorageHomePage();
+    var filesListPage = new FilesListPage();
+
+    var describeNewFolder = function() {
+
+      it('should open New Folder modal', function(){
+        storageSelectorModalPage.getNewFolderButton().click();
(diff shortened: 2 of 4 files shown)


alex-deaconu:
@ezequielc interesting solution! I didn't know that was possible - @olegrise FYI, you may want to take a look as well.

However, I've had issues where I was pressing the Delete key, thinking I was deleting from a textbox and I back navigated instead (losing my form data). Should we maybe disable Delete key navigation completely?

Also, do you think we should move this code to common-header so its common to all our projects? 

Thanks!

ezequielc:
@alex-deaconu Thanks! 
As we have a card for notifying unsaved changes, I think it is redundant to disable Delete key. Thoughts? 
Regarding moving the code to CH, I think it is a good suggestion. I will do that. 
Cheers!

alex-deaconu:
@ezequielc sounds good on both points. Thanks!

olegrise:
@alex-deaconu @ezequielc interesting solution. I didn't know about that. 
It might be useful in some situations, but not everywhere. For example we had few situations when a user would open a Payment Form in the Store, start filling it in, then at some point click on the Country dropdown, select a country and then push Delete key. Not sure what they expected the Delete key to do, but definitely not to close the modal and loose all the data they just entered. Cheers.

alex-deaconu:
@olegrise interesting point. @ezequielc FYI.

I dug up the Store PR which fixes the similar issue: https://github.com/Rise-Vision/store/pull/1098/files.

Personally, I'm not a fan of adding the specific code to every modal. However, I think with this handler (in the CH) we would deal with all situations, and the Store modal specific handler 'should' override this one, right?

Also, one thing I noticed, Store doesn't un-register its handlers on controller unload... may be an issue for backspace navigation elsewhere in the Store after one of those modals is opened.

Thanks.

olegrise:
@alex-deaconu what do you mean "_Store doesn't un-register its handlers on controller unload_"? 

@alex-deaconu @ezequielc I thought about this a little more. When I put myself into user shoes, I don't know if I would hit the Delete key in order to close the module. You see what I mean? They probably expect something else to happen. Closing the module or showing  a message "Do you want to save your changes" would be completely unexpected. Maybe I don't understand your use case.

alex-deaconu:
@olegrise http://stackoverflow.com/questions/26983696/angularjs-does-destroy-remove-event-listeners. Disregard that statement, according to the article I believe we're ok. Thanks!

olegrise:
@alex-deaconu LOL. I was just looking at the same article.

alex-deaconu:
@olegrise @ezequielc wondering if my original suggestion of disabling Delete Key navigation specifically  would work for us. Sounds like quite a hack, however (http://stackoverflow.com/questions/1495219/how-can-i-prevent-the-backspace-key-from-navigating-back). 

Thanks!

olegrise:
@ezequielc @alex-deaconu curious, have you guys discussed the UX for this use case with @RV-Adrian and @ghislaineguerin ?

ezequielc:
@olegrise @alex-deaconu I think we have two different scenarios here.

Scenario 1: When starting a new presentation, Presentation Properties modal opens. If I go back in history (either Delete key or browser back button), I will go back to Presentations List but the modal will still appear. This is the issue I fixed.

Scenario 2: When filling a form in a modal, I accidentally pressed Delete and I was sent back in navigation history. For this case, I agree that we should disable Delete key.

Thoughts? Thanks

olegrise:
In my opinion disabling Delete key will work well for both scenarios @alex-deaconu @ezequielc

RV-Adrian:
@olegrise @ezequielc @alex-deaconu 
Apologies if this is obvious --
if the user is filling in a form, and 'Delete' is disabled... how do I delete text?
I'm thinking that if text input fields enable 'Delete' while they have focus, we should be good?

ezequielc:
@RV-Adrian Yes, we need to handle that case and enable Delete for input fields.

@alex-deaconu @olegrise @RV-Adrian
To summarize, I will move this code to CH and add a code to disable Delete key. Do you guys agree?

Thanks!

alex-deaconu:
@ezequielc works for me. Thanks!

RV-Adrian:
@ezequielc Great 👍

olegrise:
@ezequielc works for me. Thanks!

ezequielc:
@alex-deaconu @RV-Adrian I just had a conversation with @olegrise and we defined that:

1)  we will disable Delete key from CH code
2)  we will dismiss modal when navigating back in history from Apps code, as it is a specific need and it could potentially affect current behavior elsewhere.

Do you guys agree? Thanks!

olegrise:
@ezequielc agreed. Thanks!

alex-deaconu:
@ezequielc works for me, thanks!

RV-Adrian:
@ezequielc awesome, thanks

ezequielc:
@alex-deaconu Please review. Staged at https://apps-stage-2.risevision.com. Thanks

alex-deaconu:
@ezequielc code looks good 👍 . Just waiting for the CH updates that @olegrise suggested. Thanks!

PR #158, opened 2016-05-05
Fix CH add user cache

ezequielc:
- Apply changes from Rise-Vision/ng-core-api-client#27


alex-deaconu:
@ezequielc 👍

PR #170, opened 2016-05-20
Priority Support banner

ezequielc:
Stage link: https://apps-stage-1.risevision.com/

### DIFF
--- a/test/unit/editor/controllers/ctr-presentation-list.tests.js
+++ b/test/unit/editor/controllers/ctr-presentation-list.tests.js
@@ -70,6 +70,10 @@ describe('controller: Presentation List', function() {
     expect($scope.search).to.have.property('reverse');
     expect($scope.search.count).to.equal(5);
   });
+
+  it('should attach presentation list to editorFactory',function(){
+    expect($scope.editorFactory.presentations).to.be.ok;
+  });
   
   describe('$loading: ', function() {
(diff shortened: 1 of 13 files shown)


alex-deaconu:
@ezequielc looks good 👍 . 

One thing I'd like to bring up. When checking the Priority Support subscription status, we use the Selected Company Id. Wondering if we should check instead if the User's Company is subscribed? Thoughts?

Thanks!

ezequielc:
@alex-deaconu Good question. I think it is OK to use Selected Company Id as the subscription is Company related. I guess we will cover some use cases and miss others.
I checked Priority Support modal and it uses Selected Company as well: https://github.com/Rise-Vision/common-header/blob/master/src/js/svc-support-factory.js#L71 So at least we are consistent.
Thoughts? Thanks

alex-deaconu:
@ezequielc agreed re: consistency with the modal. 

However, as a user who has purchased support for their company, I should be able to access priority support in my sub-companies as well. Currently, I'd have to switch back to my original company to open the support modal. It would be an inconvenience, and I'm wondering if it could lead to someone buying it twice by mistake.

Either way, this is just something to keep in mind for the future. I don't think it affects this PR. 

Thanks!

ezequielc:
@alex-deaconu Makes sense. Thanks!

PR #176, opened 2016-06-01
Auto create first Schedule

ezequielc:


web/scripts/editor/services/svc-editor-factory.js
```diff
@@ -156,8 +156,22 @@ angular.module('risevision.editor.services')
 
               $state.go('apps.editor.workspace.artboard', {
                 presentationId: resp.item.id
+              }).then(function () {
```
ezequielc: I added it after $state.go() as we close modals when changing state.

ezequielc:

ezequielc:
@alex-deaconu You can hold on the review, I found an issue when clicking Save and Preview. I'll let you know when I fix it. Thanks

alex-deaconu:
@ezequielc sounds good. Thanks!

ezequielc:
@alex-deaconu I added a fix for the Save and Preview issue. It is not an elegant solution but we can revisit it later. Please review. Thanks

alex-deaconu:
@ezequielc reviewed, looks good 👍 .  Thanks!

ezequielc:
@alex-deaconu Sorry, I missed your comment, I already squashed.
But as it breaks the functionality of auto creating the Schedule, I think it should be deployed with this PR.
Thoughts? Thanks

alex-deaconu:
@ezequielc no worries. We should be good. Thanks.

PR #179, opened 2016-06-03
Change Template - Editor UI

ezequielc:
[stage-1]

@alex-deaconu Please review. 
We could deploy this change alone, but we also could wait for detecting changes in the Presentation to be implemented. Users might click in `Change Template` and lose their work. Thoughts?
Thanks!

### DIFF
--- a/test/e2e/displays/cases/displayadd.js
+++ b/test/e2e/displays/cases/displayadd.js
@@ -77,6 +77,7 @@ var DisplayAddScenarios = function() {
     });
     
     it('should show download buttons', function() {
+      helper.wait(displayAddPage.getDownloadWindows64Button(), 'Download Button');
       expect(displayAddPage.getDownloadWindows64Button().isDisplayed()).to.eventually.be.true;
     });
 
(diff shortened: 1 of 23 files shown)


alex-deaconu:
@ezequielc to confirm, the `Change Template` button also shows when a New Presentation is selected (not only a Template). Is that intended behaviour?

Code looks good 👍 . Regarding releasing the PR to production, I'm good either way. Thanks!

ezequielc:
@alex-deaconu Good point. We call it Blank Template, so I assume we should show the button. @RV-Adrian What do you recommend? Thanks

RV-Adrian:
@ezequielc @alex-deaconu 
- I am good with 'Blank' retaining the 'Change Template' button
- My current thought is we wait until until the 'Save' confirmation dialog is also deployed, because 'Change Template' implies the data will remain intact while the visuals change. On the flip side, this is a function only new users will see, so the destructive parts should be minimal. Let's confirm at standup, thanks.

ezequielc:
@alex-deaconu Squashed and rebased. I had to fix a E2E test (https://github.com/Rise-Vision/rise-vision-apps/pull/179/files#diff-739e9dd962e23dbf6fd1a596247e8ef0R80). If you agree, we are good to release. Thanks

alex-deaconu:

PR #182, opened 2016-06-07
Fix auto created Schedule

### LINKED ISSUE #181: Schedules created for a brand new user on first Presentation/Display creation don't work
<b>How it works now:</b>
1. Create a <b>brand new Gmail account</b>, this will not work an existing account that has already created anything.
2. Complete the sign in process, and on the CTA page, press Add Presentation.
3. Select the Full Screen Image template Presentation, name it anything, add a Placeholder, and add any Widget.
4. Do whatever you need to do to the Widget to get to show in a Presentation. Press Apply.
5. Hit 'Save and Preview', you should see the Widget. Close Preview tab and return to Editor.
6. You should see a 'Congrats' popup. Select 'Add Display'. Enter a Display name, and use that Display ID on a Player.
7. Notice that Player only shows a loading animation. Restart or reboot the Player, it only shows that loading animation.
8. Delete the Schedule and recreate it with the same Presentation and Display distribution (All). 
9. Restart Player on the Displays, and the Presentation begins player as expected.

This was tested on both Electron and V2 Players on Windows 10. It acts the same way.

<b>How it should work</b>
You should be able to see the Presentation that is auto created by that first Schedule called All Displays - 24/7 instead of having to delete that one and create a new one for All Displays.

ezequielc:
Fixes #181 
I replicated the default properties of a manually created Schedule. Actually only one change. It seems to be working fine now. 
@alex-deaconu Please review.

### DIFF
--- a/test/unit/schedules/services/svc-schedule-factory.tests.js
+++ b/test/unit/schedules/services/svc-schedule-factory.tests.js
@@ -305,7 +305,7 @@ describe('service: scheduleFactory:', function() {
       content: [{
         name: 'presentationName',
         objectReference: 'presentationId',
-        playUntilDone: true,
+        duration: 10,
         timeDefined: false,
         type: 'presentation'
       }],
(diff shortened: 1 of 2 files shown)


alex-deaconu:

PR #183, opened 2016-06-08
Feature/unsaved changes

ezequielc:

### DIFF
--- a/web/scripts/editor/controllers/ctr-workspace.js
+++ b/web/scripts/editor/controllers/ctr-workspace.js
@@ -3,13 +3,65 @@
 angular.module('risevision.editor.controllers')
   .controller('WorkspaceController', ['$scope', 'editorFactory',
     'placeholderFactory', 'userState', '$modal', '$templateCache',
-    '$location', '$stateParams', '$window', 'RVA_URL',
+    '$location', '$stateParams', '$window', 'RVA_URL', '$timeout', '$state',
+    '$filter',
     function ($scope, editorFactory, placeholderFactory, userState, $modal,
-      $templateCache, $location, $stateParams, $window, RVA_URL) {
+      $templateCache, $location, $stateParams, $window, RVA_URL, $timeout,
+      $state, $filter) {
       $scope.factory = editorFactory;
       $scope.placeholderFactory = placeholderFactory;
       $scope.isSubcompanySelected = userState.isSubcompanySelected;
       $scope.isTestCompanySelected = userState.isTestCompanySelected;
+      $scope.hasUnsavedChanges = false;
+
+      var _initializing = true;
+      $scope.$watch('factory.presentation', function (newValue, oldValue) {
+        if (_initializing) {
+          $timeout(function () {
+            _initializing = false;
+          });
+        } else {
+          $scope.hasUnsavedChanges = true;
+        }
+      }, true);
+
+      $scope.$on('presentationUpdated', function () {
+        $timeout(function () {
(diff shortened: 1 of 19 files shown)


web/scripts/editor/controllers/ctr-workspace.js
```diff
@@ -3,13 +3,65 @@
...
+          $timeout(function () {
+            _initializing = false;
+          });
+        } else {
+          $scope.hasUnsavedChanges = true;
+        }
+      }, true);
+
+      $scope.$on('presentationUpdated', function () {
+        $timeout(function () {
```
alex-deaconu: @ezequielc curious, why do we need a `$timeout` here? Thanks.

web/scripts/editor/controllers/ctr-workspace.js
```diff
@@ -3,13 +3,65 @@
...
       $scope.placeholderFactory = placeholderFactory;
       $scope.isSubcompanySelected = userState.isSubcompanySelected;
       $scope.isTestCompanySelected = userState.isTestCompanySelected;
+      $scope.hasUnsavedChanges = false;
+
+      var _initializing = true;
+      $scope.$watch('factory.presentation', function (newValue, oldValue) {
+        if (_initializing) {
+          $timeout(function () {
+            _initializing = false;
```
alex-deaconu: @ezequielc why do we need a `$timeout` here as well? Thanks.

alex-deaconu:
@ezequielc one thing I noticed (http://screencast.com/t/y0IsyHFR):

When I add a new Presentation, click Apply on the Presentation settings Modal, and then Cancel, the Save notification doesn't appear. 
However, when I open the Presentation settings and click Apply on an existing Presentation the Save notification does appear. 

Is that because existing Presentations may not have all fields filled in? I don't think it's a big deal, just wanted to bring it up.

Thanks.

alex-deaconu:
@ezequielc same thing happens when I switch between HTML/Layout views.

web/scripts/editor/controllers/ctr-workspace.js
ezequielc: @alex-deaconu `$timeout` code will run just after the `$digest` cycle finishes. When loading an existing presentation, somewhere in the loop the presentation object was being updated, causing the `hasUnsavedChanges` to be flagged.
Using the `$timeout` will make sure `_initializing` is set to false only after the cycle finishes. Thanks

ezequielc:
@alex-deaconu You are right, both cases happen because an existing Presentation does not have all the fields, specially the ones that are purposely not saved as they are used only to help parsing/editor (backgroundStyle, etc). 
I tried to fix it, but I didn't spend much time trying to figure it out as we want to finish Miranda. Should I spend more time and fix it or should we add a card as an improvement for later? Thanks

alex-deaconu:
@ezequielc I don't think either are major issues; just wanted to make sure we're aware of them. We could add a card to prevent the modal from appearing for these particular cases, for our cleanup epic. Maybe a before/after comparison of the Presentation object could help. Either way, not a big deal on either. Thanks.

alex-deaconu:
@ezequielc I'm good with the code in the PR. Also, functionality works as expected, other than the things noted above. Nice work 👍 . Thanks!

ezequielc:
@alex-deaconu I implemented @RV-Adrian 's  suggestion to not notify users when changing template. Please review. Thanks

alex-deaconu:
@ezequielc looks good 👍 . Thanks.

PR #184, opened 2016-06-09
Split out display factory from controllers

alex-deaconu:
[stage-2]

### DIFF
--- a/web/scripts/app.js
+++ b/web/scripts/app.js
@@ -268,9 +268,11 @@ angular.module('risevision.apps', [
         }],
         controller: 'displayDetails',
         resolve: {
-          canAccessApps: ['canAccessApps',
-            function (canAccessApps) {
-              return canAccessApps();
+          displayId: ['canAccessApps', '$stateParams',
+            function (canAccessApps, $stateParams) {
+              return canAccessApps().then(function () {
+                return $stateParams.displayId;
+              });
             }
           ]
         }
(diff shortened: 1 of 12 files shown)


alex-deaconu:
@ezequielc PR to prepare for the Add Display Instruction Email & Modal card. 

Please review. Thanks!

web/scripts/app.js
```diff
@@ -284,9 +286,9 @@ angular.module('risevision.apps', [
         }],
         controller: 'displayAdd',
         resolve: {
-          canAccessApps: ['canAccessApps',
-            function (canAccessApps) {
-              return canAccessApps();
+          displayInfo: ['canAccessApps', 'displayFactory',
```
ezequielc: @alex-deaconu Curious, why does it need the `displayInfo` here? I think it is not being used as you inject the displayFactory to the controller. Thanks

alex-deaconu: @ezequielc it's consistent with what we have for Schedules: https://github.com/Rise-Vision/rise-vision-apps/blob/master/web/scripts/app.js#L196 (and in some ways Editor). Basically this is a dummy variable that allows us to check `canAccessApps()` and initialize the `displayFactory.newDisplay`. Any suggestions for refactoring this? Thanks.

ezequielc: @alex-deaconu Ok, I thought that maybe `canAccessApps()` alone would work, but I am good with the way it is. Thanks

ezequielc:
@alex-deaconu Looks good :+1: Thanks!

web/scripts/app.js
alex-deaconu: @ezequielc that was how I first implemented it. However I noticed that when adding a new Display, the previous Display's data would appear. So we have to reset the factory on a new Display. Thanks.

PR #186, opened 2016-06-09
Save model updates

ezequielc:

### DIFF
--- a/web/scripts/editor/controllers/ctr-workspace.js
+++ b/web/scripts/editor/controllers/ctr-workspace.js
@@ -16,12 +16,26 @@ angular.module('risevision.editor.controllers')
 
       var _initializing = true;
       $scope.$watch('factory.presentation', function (newValue, oldValue) {
+        if ($scope.hasUnsavedChanges) {
+          return;
+        }
         if (_initializing) {
           $timeout(function () {
             _initializing = false;
           });
         } else {
-          $scope.hasUnsavedChanges = true;
+          var newObject = angular.copy(newValue);
+          var oldObject = angular.copy(oldValue);
+          var ignoreFields = ['revisionStatusName', 'changeDate',
+            'changedBy', 'backgroundStyle', 'backgroundScaleToFit'
+          ];
+          for (var k in ignoreFields) {
+            delete newObject[ignoreFields[k]];
+            delete oldObject[ignoreFields[k]];
+          }
+          if (!angular.equals(newObject, oldObject)) {
+            $scope.hasUnsavedChanges = true;
+          }
         }
       }, true);
 
@@ -63,7 +77,7 @@ angular.module('risevision.editor.controllers')
         $window.onbeforeunload = undefined;
       });
 
-      $scope.changeTemplate = function() {
+      $scope.changeTemplate = function () {
         $scope.hasUnsavedChanges = false;
         $state.go('apps.editor.add');
       };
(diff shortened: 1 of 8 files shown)


ezequielc:
@RV-Adrian Could you also confirm it works as expected: 
https://apps-stage-0.risevision.com 
Thanks

alex-deaconu:
@ezequielc one thing I've noticed is that after pressing Publish the Save button is enabled. I think it should be disabled. Thanks.

ezequielc:
@alex-deaconu I checked the code for publishing Presentations and it actually doesn't update the Presentation content, it only sets the status to Published of the Presentation in the server. So the Save button should still be enabled as there are unsaved changes in the Editor.

But I don't think this is clear to the users. We should have a conversation with @RV-Adrian to improve that. I think Restore is also not very clear and I remember users complaining about it. Thoughts? Thanks

alex-deaconu:
@ezequielc to confirm, this is happening even if I don't make any changes to the Presentation. I just open a Revised Presentation and press Publish. In this case there are no changes in the Editor (other than the Published flag). I think this may be confusing for the user as it leads them into a Save & Publish loop. Thanks.

ezequielc:
@alex-deaconu Ok, I will review this case. Thanks

RV-Adrian:
@ezequielc @alex-deaconu Right now it sounds like a technical issue? (Save enabled after Publish, even with no changes present)
Let me know if not and we'll discuss.

web/scripts/editor/controllers/ctr-workspace.js
```diff
@@ -21,7 +21,18 @@ angular.module('risevision.editor.controllers')
             _initializing = false;
           });
         } else {
-          $scope.hasUnsavedChanges = true;
+          var newObject = angular.copy(newValue);
+          var oldObject = angular.copy(oldValue);
+          var ignoreFields = ['revisionStatusName', 'changeDate',
```
ezequielc: @alex-deaconu Sorry, I squashed the update. I improved the detection of changes to ignore fields that are not related to the content itself. So now it should not enable Save after publishing. 
It also fixes the issue we mentioned in the previous PR, that Save was being enabled after opening Presentation Properties. Please review. Thanks

alex-deaconu: @ezequielc I see. To clarify, ignoring `backgroundStyle` and `backgroundScaleToFit` resolves the issues with opening/closing the settings?

Wondering how efficient this code is however in terms of performance (copy object & compare could slow down the digest process). 

Otherwise, looks good. Thanks.

ezequielc: @alex-deaconu Yes, `backgroundStyle` and `backgroundScaleToFit` fix that issue.

Regarding performance, I share the same concern. I updated the code to avoid checking again when `hasUnsavedChanges` is already set: https://github.com/Rise-Vision/rise-vision-apps/pull/186/files#diff-ca5a6d5167138827bf78a94b1f30d6a4R19
As most of the time the Presentation will have unsaved changes, it will significantly reduce processing.  

Thanks!

alex-deaconu: @ezequielc good idea. Should work well this way! Thanks for looking into it.

alex-deaconu:
@ezequielc 👍

PR #190, opened 2016-06-10
Update Publish and Restore behavior

ezequielc:

### DIFF
--- a/test/e2e/editor/cases/presentation-add.js
+++ b/test/e2e/editor/cases/presentation-add.js
@@ -51,9 +51,9 @@ var PresentationAddScenarios = function() {
       expect(workspacePage.getSaveAndPreviewButton().isPresent()).to.eventually.be.false;
     });
 
-    it('should not show Publish/Restore Buttons', function () {
-      expect(workspacePage.getPublishButton().isDisplayed()).to.eventually.be.false;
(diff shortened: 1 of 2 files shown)


alex-deaconu:
@ezequielc looks good. Couple of things I just noticed:
A brand new, blank Presentation has the `Save` button disabled until I make a change.
When making a change on a Published Presentation, the `Restore` button is enabled. At this point making a server trip to Restore is useless (since the user hasn't saved), so maybe we should disable it?

Neither are major points, just raising for discussion.

Thanks.

ezequielc:
@alex-deaconu I think Restore action is ok, because it will replace the unsaved content in the editor with the Published version.
About Save, we could let it always enabled for a new Presentation. @RV-Adrian Thoughts? Thanks

RV-Adrian:
@ezequielc @alex-deaconu 
If it's extra work to enable Save for an un-modified Blank Presentation, I say leave it.
Why would they want to save something completely empty? Same could even be said for an un-modified Template I think.

ezequielc:
Thanks @RV-Adrian. It is not a lot of work, but it is extra work, including tests, etc.
@alex-deaconu If you agree, I think we are good to merge. Thanks

alex-deaconu:
@ezequielc agreed. 👍 . Thanks.

PR #193, opened 2016-06-14
Fix #189

### LINKED ISSUE #189: Editor goes in a ‘Saving’ loop when you make changes to a Presentation, don’t Save and select to Delete the Presentation
**How to recreate**
1. Create a new Presentation in the new Editor
2. Make a change to the Presentation such as adding an Image Widget
3. Press Save.  
4. Go back to the Presentation, make a change.  Don't press Save.
5. Select to Delete the Presentation

**What happens**
After you confirm that you want the Presentation deleted, you see the prompt “Do you want to save changes before leaving?”. 

If you press ‘Don’t Save’ your presentation is deleted.  

If you select ‘Save’ or 'Cancel', the screen is stuck on the ‘Saving’.  

If you go back to ‘Presentations’ on the same Window, you are prompted to save changes again.  

**What should happen**
Being able to delete a Presentation without having to save it.

ezequielc:

### DIFF
--- a/test/unit/editor/controllers/ctr-workspace.tests.js
+++ b/test/unit/editor/controllers/ctr-workspace.tests.js
@@ -88,6 +88,14 @@ describe('controller: Workspace', function() {
     expect($scope.hasUnsavedChanges).to.be.false;
   });
 
+  it('should clear unsaved changes when deleting the presentation',function(){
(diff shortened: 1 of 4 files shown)


alex-deaconu:

PR #194, opened 2016-06-22
Display Add Modal

alex-deaconu:
Show Player Install & How to show Content videos


alex-deaconu:
Took quite a long time to get this build to complete. Especially had issues with the test for setting the timezone, you can see the workarounds for it.

Also, regarding this PR, I've updated all "Add Display" scenarios to use the modal; removed the add display endpoint. However, after adding a Display, the user may just close the modal and continue doing whatever they were doing before. 

To note, we don't update the Displays list once the Display is added. Maybe we should ocnsider doing that. Thoughts?

ezequielc:
@alex-deaconu I agree with you, we should update the list. My initial idea is to broadcast an event whenever the display is created so we can refresh the list. Thanks

ezequielc:
@alex-deaconu Code looks good. As we talked in standup, we will wait for the list refresh before merging. Thanks

alex-deaconu:
@ezequielc see second commit. Once a Display is created, the list will refresh. Also refactored the list. Please review. Thanks!

ezequielc:
@alex-deaconu Code looks good 👍  Thanks for the refactoring, nice job!

PR #195, opened 2016-06-23
Configure sender name for Display emails

alex-deaconu:
New parameter as per:
https://github.com/Rise-Vision/core/pull/251

[stage-3]


alex-deaconu:

PR #196, opened 2016-06-23
UX impprovements for Distribution selector

ezequielc:
Apply changes from https://github.com/Rise-Vision/common-app-components/pull/58

### DIFF
--- a/web/scripts/app.js
+++ b/web/scripts/app.js
@@ -430,8 +430,12 @@ angular.module('risevision.apps', [
 
     }
   ])
-  .run(['$rootScope', '$state', '$modalStack',
-    function ($rootScope, $state, $modalStack) {
+  .run(['$rootScope', '$state', '$modalStack', 'displayFactory',
+    function ($rootScope, $state, $modalStack, displayFactory) {
+     
+      $rootScope.$on('distributionSelector.addDisplay', function () {
+        displayFactory.addDisplayModal();
+      });
 
       $rootScope.$on('$stateChangeStart', function () {
         $modalStack.dismissAll();
--- a/test/e2e/schedules/cases/add-distribution.js
+++ b/test/e2e/schedules/cases/add-distribution.js
@@ -68,13 +68,14 @@ var AddDistributionScenarios = function() {
         });
 
         it('should show a table for listing displays', function () {
-          expect(distributionModalPage.getDistributionListTable().isDisplayed()).to.eventually.be.true;
+          helper.waitDisappear(distributionModalPage.getDistributionListLoader()).then(function () {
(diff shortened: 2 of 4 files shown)


alex-deaconu:

ezequielc:
@alex-deaconu I added a rootscope event listener for Add Display modal. Please review. Thanks

web/scripts/app.js
```diff
@@ -430,8 +430,12 @@ angular.module('risevision.apps', [
...
     }
   ])
-  .run(['$rootScope', '$state', '$modalStack',
-    function ($rootScope, $state, $modalStack) {
+  .run(['$rootScope', '$state', '$modalStack', 'displayFactory',
+    function ($rootScope, $state, $modalStack, displayFactory) {
+      
+      $rootScope.$on('distributionSelector.addDisplay', function () {
+        displayFactory.addDisplayModal();
+      });
```
alex-deaconu: @ezequielc curious why you opted to add the listener in app.js rather than the `displayFactory` itself? Thanks.

alex-deaconu:
@ezequielc looks good. Just one question regarding listener code. Thanks.

web/scripts/app.js
ezequielc: @alex-deaconu I added in app.js because we have many other listeners here. But thinking again, it makes sense to have it in the factory. I will update. Thanks

web/scripts/app.js
```diff
@@ -430,9 +430,9 @@ angular.module('risevision.apps', [
 
     }
   ])
-  .run(['$rootScope', '$state', '$modalStack',
-    function ($rootScope, $state, $modalStack) {
-
+  .run(['$rootScope', '$state', '$modalStack', 'displayFactory',
+    function ($rootScope, $state, $modalStack, displayFactory) {
+
```
alex-deaconu: @ezequielc just a heads up, I think this file needs to be cleaned up. Thanks.

ezequielc: @alex-deaconu Thanks!

ezequielc:
@alex-deaconu Implemented as suggested. Please review. Thanks

alex-deaconu:
@ezequielc 👍

ezequielc:
@alex-deaconu I will need to revert the listener back to app.js. Having the listener in displayFacotry only works if the factory is used/injected before the event is raised. So, it works fine if we load displays list before opening the schedule page, but it will not work if my first page is the schedule. Cheers

ezequielc:
@alex-deaconu Listener is back to app.js. Please review. Thanks

alex-deaconu:

PR #206, opened 2016-07-07
Fix #205 - Not able to Save presentation after deleting Widgets

### LINKED ISSUE #205: Not able to Save presentation after deleting multiple Widgets in a Placeholder
**How to recreate**
1. Open a Presentation that was previously created 
2. Make sure that there are multiple Widgets located within a Placeholder
3. Delete at least one Widget 

**What happens**
When I delete at least one Widget, the 'Save' button is still a light green color, indicating that I cannot save it.

See http://g.recordit.co/rxTgW6Wzdy.gif

**What should happen**
When I delete, I should be able to save the Presentation

ezequielc:
Arrays where handled as regular objects


alex-deaconu:

PR #210, opened 2016-07-14
Center new placeholders

alex-deaconu:
Take scroll into consideration
Note: Offsets are based on current css values

JShint fixes

[stage-4]

### DIFF
--- a/test/e2e/editor/cases/artboard-placeholders.js
+++ b/test/e2e/editor/cases/artboard-placeholders.js
@@ -54,6 +54,14 @@ var ArtboardPlaceholdersScenarios = function() {
         it('should select placeholder after adding it', function () {
           expect(artboardPage.getPlaceholderContainer().getAttribute('class')).to.eventually.contain('edit-mode');
         });
+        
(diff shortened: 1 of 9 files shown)


alex-deaconu:

ezequielc:
@alex-deaconu Code looks good. Just one question about functionality. I had the impression we were going to fix the case where placeholders were added on top of each other when creating multiples placeholders. We would add an offset every time we create a new placeholder. Is this another card? Thanks

alex-deaconu:
@ezequielc hmm, ok. I guess I missed that; will try to revisit. Thanks.

ezequielc:
@alex-deaconu I think you should confirm in the card, maybe I am wrong. Cheers

alex-deaconu:
@ezequielc discussed on the card and during standup. I will add a `20px` offset as suggested so Placeholders don't overlap. Thanks.

ezequielc:
@alex-deaconu Looks good 👍 Thanks!

alex-deaconu:
@ezequielc thanks!

PR #215, opened 2016-07-27
Add Image/Video by URL option

alex-deaconu:
By URL button added to the Storage selector
Button shows even when Storage Trial is not started

Image widget settings are opened; if the user Saves a new Image Widget item is added
If Cancel is pressed, no item is added

Text widget logic works the same, if Cancel is pressed, no item is added

[stage-4]


alex-deaconu:

ezequielc:
@alex-deaconu Code looks good 👍 Thanks!

PR #216, opened 2016-07-27
Add Content no longer opens Playlist Item Settings

alex-deaconu:
Move subscriptionStatus check to when an item is added
Added special case for non-widgets to open Playlist Item Settings

[stage-4]


alex-deaconu:

ezequielc:
@alex-deaconu 👍 Cheers!

PR #220, opened 2016-07-29
Implemented Image/Video types filter

alex-deaconu:
[stage-4]

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -21,7 +21,7 @@
     "angular-vertilize": "~1.0.1",
     "common-header": "https://github.com/Rise-Vision/common-header.git#v1.10.10",
     "component-storage-selector": "https://github.com/Rise-Vision/component-storage-selector.git",
-    "rv-common-i18n": "https://github.com/Rise-Vision/common-i18n.git",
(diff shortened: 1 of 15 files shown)


alex-deaconu:

ezequielc:
@alex-deaconu Code looks good. 👍
Just wondering, did you consider using content type instead of file extension for filtering? Do you think it would help? Cheers

alex-deaconu:
@ezequielc good question. I did actually however I opted for this particular implementation.

Reasons? Well, actually I didn't know if the content-type header was reliable, ie. I thought it was based on the extension. However, double checking the code it seems that header is not arbitrarily set, but actually obtained from the file object in the browser (https://github.com/Rise-Vision/rise-vision-apps/blob/master/web/scripts/storage/services/svc-upload-uri.js#L32)...

Having said that, I agree it would be a more reliable method for filtering the files, in case someone gets their extensions mixed up. Do you think I should update?

Thanks.

ezequielc:
@alex-deaconu I don't think it is worth the effort right now. In the future, if we have problems we can revisit the code. 👍 Cheers!

alex-deaconu:
@ezequielc sounds good. Thanks.

PR #227, opened 2016-08-05
Responsive breadcrumbs

ezequielc:
Breadcrumbs are responsive based on the width of parent container.
It will always show root, current and parent folders. It will hide in between folders if there is not enough space. Also, every folder name has a limit of 20 chars.
@alex-deaconu Please review.

### DIFF
--- a/web/partials/storage/files-list.html
+++ b/web/partials/storage/files-list.html
@@ -1,4 +1,4 @@
-<div class="content-box" ng-show="!trialAvailable">
+<div class="content-box half-top" ng-show="!trialAvailable">
   <div class="scrollable-list" 
   rv-spinner rv-spinner-key="storage-selector-loader"
   rv-spinner-start-active="1" >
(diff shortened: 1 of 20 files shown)


web/partials/storage/files-list.html
```diff
@@ -62,7 +62,7 @@
         <tr class="clickable-row"
         ng-click="fileClick(file);"
         ng-class="{'active': file.isChecked, 'blocked-file': file.isThrottled, 'back-btn': file.currentFolder, 'no-select-row': !storageFactory.canSelect(file), 'disabled-row': storageFactory.isDisabled(file) }"
-        ng-repeat="file in filesDetails.files | filter:search.query | orderBy:orderByAttribute:reverseSort track by $index" ng-if="!storageFactory.fileIsTrash(file) || storageFactory.storageFull">
+        ng-repeat="file in filesDetails.files | filter:search.query | orderBy:orderByAttribute:reverseSort track by $index" ng-if="file.name !== storageFactory.folderPath && (!storageFactory.fileIsTrash(file) || storageFactory.storageFull)">
```
alex-deaconu: @ezequielc wondering if instead of filtering out the `folderPath` folder we should just prevent it from being created here: https://github.com/Rise-Vision/rise-vision-apps/blob/master/web/scripts/storage/services/svc-files-factory.js#L88. Thanks.

alex-deaconu:
@ezequielc looks good 👍 . Just one suggestion in the code. Thanks.

web/partials/storage/files-list.html
ezequielc: @alex-deaconu I thought about it but when I read the comment about Trash (https://github.com/Rise-Vision/rise-vision-apps/blob/master/web/scripts/storage/services/svc-files-factory.js#L104) I didn't try. But it is a good suggestion. I will try and see how it works. Thanks

ezequielc:
@alex-deaconu I removed the code to add current folder to the list, so it will not appear anymore. Also, Storage Home and Selector layouts have been updated. I squashed the code as we had too many commits here. We can merge and we are OK to release. Please review. Thanks!

alex-deaconu:
@ezequielc looks good 👍 . Thanks for updating!

PR #244, opened 2016-08-24
draggable placeholder playlist [stage-1]

settinghead:


settinghead:
@alex-deaconu @ezequielc please review.

Notes:
1. I spent quite some time trying to get the drag and drop to work in E2E tests, but it seems it's a problem that's still unsolved for Chrome Driver (see [bug report here](https://bugs.chromium.org/p/chromedriver/issues/detail?id=841))
2. I removed all logic regarding move up and down button, because it's not part of UX. Let me know if that's the desired thing to do, since I remember seeing @alex-deaconu just adding this feature very recently

Cheers

alex-deaconu:
@settinghead your stage environment is missing for the last commit. I'm going to overwrite stage-0 with @ezequielc's PR for Tooltips which I am reviewing right now. Thanks.

settinghead:
@alex-deaconu my bad. It should be staged now.

web/partials/editor/placeholder-playlist.html
```diff
@@ -16,16 +16,12 @@
...
+<div class="editor-playlist scrollable-list add-top sortable-list">
   <!-- Item Playlist -->
-  <div class="editor-playlist-item" ng-repeat="item in factory.getItems() track by $index" ng-show="factory.getItems().length">
-    <div class="arrows">
-      <button class="btn btn-block btn-xs btn-default" id="moveUpButton" ng-click="factory.movePlaylistItemUp(item)" ng-disabled="!factory.canPlaylistItemMoveUp(item)"><i class="fa fa-arrow-up"></i></button>
-      <button class="btn btn-block btn-xs btn-default" id="moveDownButton" ng-click="factory.movePlaylistItemDown(item)" ng-disabled="!factory.canPlaylistItemMoveDown(item)"><i class="fa fa-arrow-down"></i></button>
-      
-      <!-- <i class="fa fa-grip-large fa-rotate-90"></i> -->
+  <div class="editor-playlist-item draggable-item" ng-repeat="item in factory.getItems() track by $index" ng-show="factory.getItems().length">
+    <div class="list-grip e2e--drag-handle">
```
alex-deaconu: @settinghead interesting approach for setting and labelling an e2e specific class. This is a new pattern. @ezequielc do you think this is something we should be doing? Thanks.

web/scripts/editor/directives/dtv-placeholder-playlist.js
```diff
@@ -42,6 +42,27 @@ angular.module('risevision.editor.directives')
             });
           };
 
+          applySortable();
+
+          function applySortable() {
```
alex-deaconu: @settinghead I would prefer if we had a `sortable` directive, which took one parameter for the sort function (`placeholderPlaylistFactory.moveItem`). Also you would probably have to modify `moveItem` to parameters to `(oldIndex, newIndex)`, but that's ok. This way this functionality can then be re-used. Thanks.

alex-deaconu:
@settinghead there's one issue which I discovered with the staged version. Not sure what the cause is but I couldn't figure out a way to recover from this state: http://screencast.com/t/lA69BgIOGG. Can you take a look? Thanks.

web/scripts/editor/directives/dtv-placeholder-playlist.js
settinghead: @alex-deaconu good idea. Will do.

web/partials/editor/placeholder-playlist.html
ezequielc: @alex-deaconu @settinghead It is an interesting approach, but I don't see many benefits in comparison to using the original class. We are still using CSS selector and by using the original class, we may catch issues like a missing class, etc. Plus, we are not adding something new to the code. Do you see any other benefits by using e2e classes? Thanks

settinghead: @alex-deaconu @ezequielc I'll argue for two benefits:
1. Separation of concern. if a class is multi-purpose (say it changes, width, color, and in this case is also an E2E element identifier), it might create difficulty in maintenance: if we want to change the definition of this class, because it's range of purpose is bloated, we won't have much confidence that our changes won't break other parts of the code. 
2. Visibility of purpose. If we have an explicit label, it makes both the developer and designer aware that there are E2E tests associated with the element, which reminds us to be careful when we change the UX DOM structure.

However, I agree that the benefit for introducing `e2e-` class isn't obvious in this case, because `list-grip` happens to serve as a clear identifier on its own. 

This pattern is probably most useful if we have dedicated QA's writing test cases, that way developers and QA's won't interfere with each other's work.

That said, I'm OK with not introducing this pattern for the moment, until next time I find a stronger case for it.

ezequielc: @settinghead Sounds good to me. Thanks!

settinghead:
@alex-deaconu I wasn't able to reproduce the bug on Stage 1. What browser are you using?

alex-deaconu:
@settinghead I'm using Chrome. Here it is again: http://screencast.com/t/b5JxaIm8Xx. Thanks.

settinghead:
@alex-deaconu @ezequielc 
Update on PR:

- Removed `e2e-` class
- Created `rv-sortable` directive. I tried to add tests, but manual event triggering for Sortable library proved to be difficult, so I had to give up
- Changed signature of `moveItem` from `(item, newIndex)` to `(oldIndex, newIndex)`
- Fixed scroll issue @alex-deaconu found

Currently staged on https://apps-stage-0.risevision.com/

alex-deaconu:
@settinghead re: unit tests for the `rv-Sortable` directive, I think you can mock the `Sortable` object and test that the `create` and `destroy` functions being called correctly. Thanks.

alex-deaconu:
@settinghead reviewed, code looks good and I can confirm the issue is fixed. Just a suggestion on the unit tests. Thanks!

web/scripts/editor/directives/dtv-placeholder-playlist.js
```diff
@@ -55,6 +55,13 @@ angular.module('risevision.editor.directives')
             });
           };
 
+          $scope.sortItem = function (evt) {
+            var oldIndex = evt.oldIndex;
```
alex-deaconu: @settinghead also just noticed this function is not unit tested. Thanks!

settinghead:
@alex-deaconu @ezequielc Unit tests for `sortItem` and `rv-sortable` directive added. Please review. Thanks!

alex-deaconu:
@settinghead looks good 👍 . Please squash before merging. Thanks!

PR #245, opened 2016-08-25
Playlist tooltips

ezequielc:
- Tooltips for Playlist Items: Duplicate, Delete and Duration
- Artboard Empty State (help image) for a new blank Presentation (not from Template)
- Tooltip pops up when closing presentation properties:
  - On Add Placeholder button for a new blank Presentation
  - On first Placeholder when new Presentation from a Template

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -22,7 +22,7 @@
     "common-header": "https://github.com/Rise-Vision/common-header.git#v1.11.1",
     "component-storage-selector": "https://github.com/Rise-Vision/component-storage-selector.git",
     "rv-common-i18n": "https://github.com/Rise-Vision/common-i18n.git",
-    "rv-common-style": "https://github.com/Rise-Vision/common-style.git#v1.3.80",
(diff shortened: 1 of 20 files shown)


alex-deaconu:
@ezequielc just playing with the deployed version, the tooltip popup seems to sometimes block clicking on the actual button. I think the animation + container from the tooltip is receiving the click event rather than the button itself: http://screencast.com/t/oUqqO4zC. Can you take a look? Thanks.

ezequielc:
@alex-deaconu Thanks, I will take a look. Maybe we should bounce only for the first time, right after we close Presentation Properties, on mouse enter, there is no need to bounce. If you agree, I will try to implement that. Thanks

alex-deaconu:
@ezequielc one bounce works for me. Thanks.

ezequielc:
@alex-deaconu Updated the behaviour, now it only bounces by the event of closing Presentation Properties. Please review. Thanks
@settinghead fyi

alex-deaconu:
@ezequielc looks good. Thanks for fixing 👍

PR #252, opened 2016-09-01
Fix #248 : Unsaved changes dialog should appear 

### LINKED ISSUE #248: If you are in a Presentation in your Parent Company and you switch to a Sub Company, the Editor stays on the Presentation in the Parent Company.
<b>How it works now:</b>
1. Open a Presentation in your Parent Company.2
2. Switch to a Sub Company. Notice that the Presentation remains on the Presentation from your Parent Company.

Also:
1. Make a change to the Presentation in the Parent Company and do not save.
2. Switch to the Sub Company, and click on Presentations. Notice that it does not ask you to save, and it removes you from the Sub Company and takes you back to the Parent Company.

For further clarification: https://goo.gl/k5Kw8h

<b>How it should work:</b>
If you switch to a Sub Company while you are in a Presentation in the Parent, it should tell you that Presentation is not available in the Sub Company and spit you out to the Presentations page.

ezequielc:
and Presentation list should remain in sub company

When changing to a subcompany, CID is changed in search string. When calling event.presentDefault on stateChange to show Unsaved Changes dialog, a new state change was triggered as the cid has changed since last state change. Adding the flag `reloadOnSearch` prevents the reload and the subsequent trigger of stateChange. 
I could not think of a scenario where this could cause issues as this change is restricted to the workspace states (artboard and htmleditor). 

@alex-deaconu  Please review. Thanks

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ tmp
 web/tmp
 web/scripts/config/config.js
 test/e2e/config/config.json
+.DS_Store
--- a/test/unit/app.tests.js
+++ b/test/unit/app.tests.js
@@ -91,6 +91,27 @@ describe('app:', function() {
     });
   });
 
+  describe('state apps.editor.workspace.artboard:',function(){
+    it('should register state',function(){
+      var state = $state.get('apps.editor.workspace.artboard')
+      expect(state).to.be.ok;
+      expect(state.url).to.equal('');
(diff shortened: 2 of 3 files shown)


alex-deaconu:
@ezequielc looks good 👍 . Fix is nice and clean!

Rebased and added a file to `.gitignore`. Merging. Thanks!

PR #265, opened 2016-10-12
Fix #261 - Mi'cros'fot was being handled as a ChromeOs

### LINKED ISSUE #261: Landscape rotation setting showing for non-chrome os devices
See here, http://www.screencast.com/t/nU6O8ieGvhHk

This should only appear for Chrome OS devices.

ezequielc:
Our condition to check for chrome os was interpreting Mi**cros**oft label as Chrome OS a device.

@alex-deaconu @settinghead Please review. Thanks


alex-deaconu:
@ezequielc interesting issue. Thanks for fixing 👍

PR #267, opened 2016-10-13
Fix #264 - First file in the list was being duplicated on re-upload

### LINKED ISSUE #264: Files are not overwriting in Storage when you upload one with the same name
<b>How it works now:</b>
1. Take a jpg image file and upload it to Storage.
2. Take another jpg image file, rename it the same as the first image, and upload it, to replace the original.
3. Notice that it does not overwrite the original, instead it adds another image. Looking at the thumbnails, they appear as the same image.

For further clarification: https://goo.gl/f5CnuA

<b>How it should work:</b>
The file should be overwritten with the  newly uploaded version.

ezequielc:
When `existingFileNameIndex` was 0, it was not being handled as an existing file.
Change to default and check for -1.

@settinghead @alex-deaconu Please review. Thanks!

### DIFF
--- a/test/unit/storage/services/svc-files-factory.tests.js
+++ b/test/unit/storage/services/svc-files-factory.tests.js
@@ -145,6 +145,14 @@ describe('service: filesFactory:', function() {
         expect(filesFactory.filesDetails.files.length).to.equal(6);
       });
     });
+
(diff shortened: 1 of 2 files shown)


alex-deaconu:
@ezequielc 👍

PR #286, opened 2016-11-24
If the display is online, consider it as installed and don't prompt to download it

fjvallarino:
This change hides the download buttons if the display is online but its basic information has not yet been updated (mainly, playerVersion). I'm wondering if we should display something else to the user, since having a bunch of empty fields does not make a lot of sense. The change for running the Core task every 5 minutes was handled in a Core PR.

### DIFF
--- a/web/partials/displays/display-fields.html
+++ b/web/partials/displays/display-fields.html
@@ -15,7 +15,7 @@
 </div>
 
 <!--      INSTALL PLAYER      -->
-<div ng-show="display.id && !display.playerVersion">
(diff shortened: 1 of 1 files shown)


alex-deaconu:
@fjvallarino curious, would this cause the page to flicker (hiding the fields) while waiting for the response from the Messaging Service? Thanks.

fjvallarino:
@alex-deaconu yes, it will show the Download buttons until the online status is determined. As soon as the display is considered online, that section will be hidden and the fields will be displayed.

alex-deaconu:
@fjvallarino okay, thanks for the clarification. 👍

PR #288, opened 2016-12-01
Update menu items

fjvallarino:
Git hashes will be replaced with proper version numbers when the related components are released/tagged

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -19,7 +19,7 @@
     "angular-bootstrap-colorpicker": "~3.0.19",
     "angular-ui-codemirror": "~0.3.0",
     "angular-vertilize": "~1.0.1",
-    "common-header": "https://github.com/Rise-Vision/common-header.git#v1.11.5",
+    "common-header": "https://github.com/Rise-Vision/common-header.git#v1.11.6",
     "component-storage-selector": "https://github.com/Rise-Vision/component-storage-selector.git",
     "rv-common-i18n": "https://github.com/Rise-Vision/common-i18n.git",
     "rv-common-style": "https://github.com/Rise-Vision/common-style.git#v1.3.88",
--- a/web/scripts/config/dev.js
+++ b/web/scripts/config/dev.js
@@ -32,8 +32,9 @@
     .value('ALERTS_WS_URL',
       'https://rvacore-test.appspot.com/alerts/cap')
     .value('STORAGE_ENDPOINT_URL',
-       'https://storage-dot-rvacore-test.appspot.com/_ah/api')
+      'https://storage-dot-rvacore-test.appspot.com/_ah/api')
     .value('MESSAGING_URL',
-       'https://display-messaging-staging.risevision.com');
+      'https://display-messaging-staging.risevision.com')
+    .value('APPS_URL', '');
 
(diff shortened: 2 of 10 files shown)


web/scripts/config/dev.js
```diff
@@ -32,8 +32,9 @@
     .value('ALERTS_WS_URL',
       'https://rvacore-test.appspot.com/alerts/cap')
     .value('STORAGE_ENDPOINT_URL',
-       'https://storage-dot-rvacore-test.appspot.com/_ah/api')
+      'https://storage-dot-rvacore-test.appspot.com/_ah/api')
     .value('MESSAGING_URL',
-       'https://display-messaging-staging.risevision.com');
+      'https://display-messaging-staging.risevision.com')
+    .value('APPS_URL', '');
```
alex-deaconu: @fjvallarino wondering why this is blank. Also, same goes for `test.js` below. Thanks.

fjvallarino: @alex-deaconu to make the urls relative to the current application. For dev I was not sure what port was being used, or if it's dynamic. For test (what CircleCI runs), the same happens.

alex-deaconu: @fjvallarino okay, sounds good. Thanks.

bower.json
```diff
@@ -19,7 +19,7 @@
     "angular-bootstrap-colorpicker": "~3.0.19",
     "angular-ui-codemirror": "~0.3.0",
     "angular-vertilize": "~1.0.1",
-    "common-header": "https://github.com/Rise-Vision/common-header.git#v1.11.5",
+    "common-header": "https://github.com/Rise-Vision/common-header.git#2fd604a",
```
alex-deaconu: @fjvallarino my suggestion is to use the branch name here rather than commit hash. The deployed version still has the old CH from this commit. The User/Alert icons are swapped. Thanks.

alex-deaconu:
@fjvallarino one suggestion regarding the bower config for the commit. Otherwise looks good. Thanks.

fjvallarino:
@alex-deaconu yes, I agree. I usually do that, I don't know why I used hashes in this case. It's updated now; I hadn't requested review again because I had not updated the reference.

@ezequielc FYI

web/scripts/config/dev.js
fjvallarino: @alex-deaconu this was updated too

alex-deaconu:
@fjvallarino not sure if it's an issue, but when I click on alerts settings it seems to reload the page (ie. it doesn't just navigate through ui-router). Not a big deal, just wondering why that is.

Otherwise, looks good 👍 . Thanks!

fjvallarino:
@alex-deaconu I guess it's because the navigation comes from a direct link created by CH and not by the navigation rules handled by the main Apps' handler (it puzzles me a bit too, to be honest)

alex-deaconu:
@fjvallarino if it becomes an issue I think we can address later. Otherwise, I think we're good to go, unless @ezequielc disagrees. Thanks.

ezequielc:
@alex-deaconu @fjvallarino I am good with that. Thanks!

PR #290, opened 2016-12-08
remove param value showui=false from scheduleFactory.getPreviewUrl()

settinghead:


ezequielc:
@settinghead I believe your tests are failing for the same reason as @fjvallarino found out yesterday. Angular was updated to 1.6 and it removed $http.success(). A quick workaround is to force angular version to 1.5. Cheers!

alex-deaconu:
@ezequielc do you think we should consider fixing instead? Thanks.

ezequielc:
@alex-deaconu I think we should add a separate card for the fix. It might require updating other repos and I am not sure if @settinghead has the availability to do that right now. Thoughts?

alex-deaconu:
@ezequielc okay.

settinghead:
@alex-deaconu @ezequielc @fjvallarino I'll do an update shortly, because I will need this update to get the PR merged. Thanks.

settinghead:
@Rise-Vision/apps the build is now fixed. @fjvallarino Thanks for the tip.

alex-deaconu:
@settinghead 👍

PR #291, opened 2016-12-09
Add last connection time support

fjvallarino:

### DIFF
--- a/web/partials/displays/displays-list.html
+++ b/web/partials/displays/displays-list.html
@@ -36,11 +36,11 @@
           <th class="visible-lg visible-md" translate>
             displays-app.list.heading.address
           </th>
-          <th id="tableHeaderStatus" class="col-sm-2 text-center clickable" ng-click="displays.sortBy('onlineStatus')">
+          <th id="tableHeaderLastConnection" class="col-sm-2 text-center">
+            {{ 'displays-app.list.heading.lastConnection' | translate }}
+          </th>
+          <th id="tableHeaderStatus" class="col-sm-2 text-center">
             {{ 'displays-app.list.heading.status' | translate }}
-            <!-- ngIf: search.sortBy == 'onlineStatus' -->
-            <i ng-if="search.sortBy == 'onlineStatus'" class="fa" ng-class="{false: 'fa-long-arrow-up', true: 'fa-long-arrow-down'}[search.reverse]"></i>
-            <!-- end ngIf: search.sortBy == 'onlineStatus' -->
           </th>
         </tr>
       </thead>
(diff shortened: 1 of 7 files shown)


ezequielc:
@fjvallarino Your branch is not staged anymore, so I could not confirm, but by looking at the code, it seems that the last connection time would be visible while the request to messaging service is in progress. This means that the value could change after the request is finished, causing the displayed value to change for the user. 
I think we should hide the field until the request is finished, just like we do for status. Thanks!

fjvallarino:
@ezequielc no, the fields are different. Core sends lastActivityDate, but the list and details page display lastConnectionTime, which is only updated when the Messaging Service sends the response (even for Legacy players)

ezequielc:
@fjvallarino Awesome. Thanks for the confirmation!

web/partials/displays/displays-list.html
```diff
@@ -36,18 +36,19 @@ <h1 id="title" translate>displays-app.title</h1>
...
-            <!-- end ngIf: search.sortBy == 'onlineStatus' -->
           </th>
         </tr>
       </thead>
       <tbody>
         <tr class="clickable-row" ng-class="{ 'bg-warning' : selectedCompayId !== display.companyId }" ng-repeat="display in displays.items.list" ui-sref="apps.displays.change({displayId: display.id, companyId: display.companyId})">
           <td class="col-sm-5 display-name"><a ui-sref="apps.displays.change({displayId: display.id, companyId: display.companyId})"><span class="ellipsis"><strong>{{display.name}}</strong></span></a></td>
           <td class="nowrap display-address visible-lg visible-md"><a ui-sref="apps.displays.change({displayId: display.id, companyId: display.companyId})"><span class="text-muted ellipsis">{{display.address}}</span></a></td>
+          <td class="nowrap display-last-connection visible-lg visible-md"><a ui-sref="apps.displays.change({displayId: display.id, companyId: display.companyId})"><span class="text-muted ellipsis">{{display.lastConnectionTime | date:'dd/MM/yyyy hh:mm:ss a'}}</span></a></td>
```
alex-deaconu: @fjvallarino wondering if you can use the original filter for this and in display-fields.html: 
date:'d-MMM-yyyy h:mm a'

(https://github.com/Rise-Vision/rise-vision-apps/commit/d1d0bc8b6fd3ae5e10d6da42dae3865e78e44043#diff-49403f2e7a2f5a4d2a3def0f1ea34bf4L58)

Thanks.

alex-deaconu: @fjvallarino okay. I was mainly suggesting because of consistency with wherever else we use Date/Time values (https://github.com/Rise-Vision/common-app-components/blob/master/src/last-modified/last-modified.html). Thanks.

web/scripts/displays/services/svc-display.js
```diff
@@ -237,23 +237,27 @@
     ]);
 
   function _mergeStatuses(items, statuses) {
-    var lookup = {};
+    items.forEach(function(item) {
+      item.lastConnectionTime = item.lastActivityDate;
+    });
```
fjvallarino: @alex-deaconu the main reason is to avoid showing the field until the Messaging Service answers. It comes from a discussion with @ezequielc about avoiding flickering/changing values and confusing the user.

alex-deaconu: @fjvallarino aha, makes sense. Thanks for the clarification.

fjvallarino:
@Rise-Vision/apps updated

alex-deaconu:
@fjvallarino PR looks good 👍 . Thanks for fixing. 

However, I have to mention that you keep overriding your staging versions. I still have not been able to see this particular PR staged on stage-0.

PR #295, opened 2016-12-14
Add download player popup

fjvallarino:


alex-deaconu:
@fjvallarino apps is not es6 enabled. Also, not sure how to access the new functionality, I guess that's part of another card. Otherwise, looks good to me 👍 . Thanks!

PR #296, opened 2016-12-15
Align last connection time to the right

fjvallarino:
Not submitted earlier because Circle was having connectivity issues. Can be checked at: https://apps-stage-0.risevision.com


ezequielc:
@fjvallarino 👍

PR #302, opened 2017-01-10
Fix initial value for user state select

ezequielc:
Complementary PR for https://github.com/Rise-Vision/common-header/pull/645


alex-deaconu:
@ezequielc 👍

PR #304, opened 2017-01-11
Add screenshot support

fjvallarino:


web/scripts/displays/controllers/ctr-display-details.js
```diff
@@ -96,6 +98,67 @@ angular.module('risevision.displays.controllers')
...
+      };
+
+      $scope.screenshotState = function(display) {
+        var statusFilter = $filter('status');
+
+        if(!display || $scope.displayService.statusLoading || $scope.displayService.statusLoading) {
+          return '';
+        }
+        else if(display.os && display.os.indexOf('cros') >= 0) {
+          return 'os-not-supported';
```
alex-deaconu: @fjvallarino wondering if we should use constants for these states. Thanks.

web/scripts/displays/services/svc-display.js
```diff
@@ -263,4 +302,32 @@
 
     item.lastConnectionTime = !isNaN(lookup.lastConnectionTime) ? new Date(lookup.lastConnectionTime) : (item.lastActivityDate || '');
   }
+
+  function loadImage(imageUrl, cb) {
```
alex-deaconu: @fjvallarino I would recommend a separate service for this, as it is unrelated to svc-display. Thanks.

web/scripts/displays/services/svc-primus.js
```diff
@@ -1,46 +1,97 @@
...
-        var primus = new Primus(
-          MESSAGING_URL, {
-            reconnect: {
-              retries: 0,
-            },
+  .factory('getDisplayStatus', ['loadPrimus', '$q', '$timeout', function (loadPrimus, $q, $timeout) {
+    return function (displayIds) {
+      var deferred = $q.defer();
+
+      loadPrimus.create()
```
alex-deaconu: @fjvallarino it seems we open a new Primus connection for every request. Is that ok in terms of usage for the service? Should we be re-using the same connection if available? Or, should we be closing old connections once we received the result? Thanks.

web/scripts/displays/controllers/ctr-display-details.js
fjvallarino: @alex-deaconu I thought about it, but they're only used in that function and on the html partial (where I can't access them, I understand), so I don't think it adds much value.

web/scripts/displays/services/svc-display.js
fjvallarino: @alex-deaconu makes sense

web/scripts/displays/controllers/ctr-display-details.js
alex-deaconu: @fjvallarino if you put them on the $scope they will be available in the partial. I don't feel strongly about this however, so either way works.

web/scripts/displays/services/svc-primus.js
fjvallarino: @alex-deaconu I changed the way the connections are handled and they are being closed now after a request is done, so resources are being freed. I will re-think the way we use the connection in order to keep it open and reuse it for both services (definitely makes sense if we're planning on having more interaction with the Messaging Service); I opted for this solution because it was simpler in terms of event handling and, mainly, because we rarely use the connection.

alex-deaconu: @fjvallarino sounds good. But definitely something to consider if we start using the service for more tasks. Thanks!

fjvallarino:
@Rise-Vision/apps changes applied

alex-deaconu:
@fjvallarino FYI there are some commit conflicts to resolve. Thanks!

fjvallarino:
@alex-deaconu rebased and pushed. I reverted the beautify commit, since it makes reviewing the changes really hard (comparing with master, basically). I will apply it before merging.

web/partials/displays/display-fields.html
```diff
@@ -24,7 +24,7 @@
           </button>
         </div>
         <div class="col-sm-4">
-          <button id="downloadPlayer" type="button" class="btn btn-default btn-block half-bottom" ng-click="showDownloadButtons = true">
+          <button id="downloadPlayer" type="button" class="btn btn-default btn-block half-bottom" ng-click="showDownloadButtons = true"I can add it, no problem. >
```
alex-deaconu: @fjvallarino typo here. Thanks.

fjvallarino: @alex-deaconu good catch. I have to fix the copy/paste mechanism on my machine. It's driving me crazy

web/scripts/common/services/svc-resource-loader.js
```diff
@@ -0,0 +1,35 @@
+'use strict';
+
+angular.module('risevision.apps.services')
+  .service('imageBlobLoader', ['$q', function ($q) {
+    return function (imageUrl) {
+      var deferred = $q.defer();
+      var xhr = new XMLHttpRequest();
```
alex-deaconu: @fjvallarino curious why did you opt to use the `XMLHttpRequest` rather than the angular js `$http` service? Just wondering if there were any technical issues, I'm not suggesting you update this. Thanks!

alex-deaconu:
@fjvallarino just one question. Otherwise PR looks good. Also double checked staging and it looks good as well 👍 . Thanks!

web/scripts/common/services/svc-resource-loader.js
fjvallarino: @alex-deaconu I thought about $http when this was already working, so I decided not to change it. It should be possible to do the same, I think.

alex-deaconu: @fjvallarino okay, thanks.

fjvallarino:
@alex-deaconu updated with latest common-style tag and formatted code.

Staged at: https://apps-stage-0.risevision.com

@RV-Adrian @ezequielc do you have any code pending merge to Apps? This formatting commit may cause issues, so it's probably better I merge after you

ezequielc:
@fjvallarino Yes, your staged version has some issues with tables. @RV-Adrian I think we need to merge your PR first. Thanks!

RV-Adrian:
@fjvallarino @ezequielc 
Yes, it looks like fix/table-classes-3 should deploy first. Hoping to have all PRs concluded by EOD..

fjvallarino:
@Rise-Vision/apps this is again staged, after being rebased. This is staged at stage-4. I'll have to modify the circle.yml script so the environment is selected based only on commits from the current branch (otherwise it does not go to stage-0 if nothing is selected, but whatever finds on a master commit). Maybe even going back to the original way of per commit environment selection would be better.

https://apps-stage-4.risevision.com

alex-deaconu:
@fjvallarino looks good. Checked the staging link and didn't find any issues. My suggestion is however not to Squash the 2 commits when merging; it will be easier to look back at the changes in the future. Thanks.

fjvallarino:
@alex-deaconu so, do we want to keep the formatting changes?

alex-deaconu:
@fjvallarino yes, just don't Squash so we can see the 2 separate commits.

RV-Adrian:
@fjvallarino Looks good so far. Do you have a direct link where I can see a cached screenshot?

ezequielc:
@fjvallarino You should ping Support room about this change. Thanks!

fjvallarino:
@ezequielc I will as soon as I tested everything works fine (I have a new PR pending review, fixing a small issue)

ezequielc:
@fjvallarino Great! Thanks!

ezequielc:
@fjvallarino I might have found an issue actually. A win10 display is showing as OS not supported.
https://apps.risevision.com/displays/details/89YMNGW3J8SE?cid=fee1f642-cdd1-4bc4-8f93-1fc97cb00d55
Could you review? Thanks!

fjvallarino:
@ezequielc that's the issue I'm waiting for review: https://github.com/Rise-Vision/rise-vision-apps/pull/316

PR #307, opened 2017-01-17
Fix/table classes

RV-Adrian:


web/partials/displays/display-fields.html
```diff
@@ -1,212 +1,193 @@
 <div class="form-group" ng-class="{'has-error': !displayDetails.name.$pristine && displayDetails.name.$invalid}" show-errors>
-  <label class="control-label">{{'displays-app.fields.name.name' | translate}} *</label>
```
alex-deaconu: @RV-Adrian lots of spacing changes in this file. Not sure if there are any actual code changes in here as it is almost impossible to review, can you confirm? Also, are the formatting updates necessary? Thanks.

web/partials/editor/presentation-list.html
```diff
@@ -20,41 +20,41 @@ <h1 id="title" translate>editor-app.list.title</h1>
...
       rv-spinner-start-active="1"
       >
-      <table id="presentationListTable" class="table-2 table-hover">
-        <thead>
-          <tr class="clickable">
-            <th id="tableHeaderName" ng-click="factory.sortBy('name')">
+      <table id="presentationListTable" class="table table--hover">
+        <thead class="table-header">
+          <tr class="table-header__row clickable">
+            <th id="tableHeaderName" class="table-header__cell clickable" ng-click="factory.sortBy('name')">
```
alex-deaconu: @RV-Adrian curious, is the `clickable` class necessary here as the `<tr>` already has it? Thanks.

RV-Adrian: @alex-deaconu Good point.  I ended up adding 'clickable' to rows only on every table, and that seems to have done the trick and cleaned up class usage a bit.

web/partials/displays/display-fields.html
RV-Adrian: @alex-deaconu It appears to be only formatting. Not sure why the spacing got thrown in. Is there a standard Apps usually uses to get on the same page with indentation etc? (ie. a certain Sublime package)

alex-deaconu: @RV-Adrian no we don't use a standard, but generally formatting changes are very hard to review and mostly unnecessary in a commit. Are you using a plugin that made these changes automatically? Thanks.

RV-Adrian: @alex-deaconu I use a prettifier shortcut in Sublime. I may have accidentally hit it in this HTML file. Formatting adjusted now, should be easier to scan.

alex-deaconu:
@RV-Adrian looks good to me 👍 . Thanks!

PR #312, opened 2017-01-20
Issue Fix / Table Classes

RV-Adrian:


alex-deaconu:
@RV-Adrian how did you manage to modify every file in this repo? I don't think that is acceptable, can you please fix? Thanks.

PR #316, opened 2017-01-24
Validate Chrome OS name only at the beginning of the string (avoid rejecting Microsoft)

fjvallarino:


ezequielc:
@fjvallarino 👍 Thanks for fixing it!

alex-deaconu:
👍

PR #319, opened 2017-01-27
Feature/revenue experiment

RV-Adrian:


RV-Adrian:
Please review PR for https://trello.com/c/ni2X6pJQ/2005-revenue-experiment-contextualized-ads-on-list-pages-1

Staged on https://apps-stage-3.risevision.com

web/partials/displays/display-details.html
```diff
@@ -1,4 +1,5 @@
 <div class="app-header">
+  <div class="banner-block"></div>
```
alex-deaconu: @RV-Adrian curious, why are you adding black `banner-block` divs here and for schedule-details.html? Thanks.

alex-deaconu:
@RV-Adrian reviewed. In addition to the inline comment, you seem to have other layout changes that don't seem to be related to this PR. Are those intentional? Thanks.

web/partials/displays/display-details.html
RV-Adrian: @alex-deaconu 
The reason for including it (despite being empty) was because .app-header uses a static column size layout across the app, and I didn't want to push that too far. Removing .banner-block results in the appearance below (for Display details)

image

RV-Adrian:
@alex-deaconu The spacing is unintentional except in app-header.html, where gaps were causing underlined blank space (upon link hover). I'll clean up the rest now. 👍

web/partials/displays/display-details.html
alex-deaconu: @RV-Adrian okay, that's what I thought. So basically this version of common-style breaks the layout of all the `.app-header` divs' spacing unless they include `.banner-block`?

alex-deaconu:
@RV-Adrian I wasn't referring to the spacing necessarily. I was referring to the additional changes required for the Storage and Alerts header. See inline comment/question. Thanks.

web/scripts/displays/services/svc-display-tracker.js
```diff
@@ -3,7 +3,8 @@
 angular.module('risevision.displays.services')
   .value('DISPLAY_EVENTS_TO_BQ', [
     'Display Created',
-    'Player Download'
+    'Player Download',
+    'Clicked Experiment Jan2017'
```
RV-Adrian: @ezequielc Is this needed by BQ?

web/partials/displays/display-details.html
RV-Adrian: @alex-deaconu That's right

web/scripts/displays/services/svc-display-tracker.js
ezequielc: @RV-Adrian We don't need to send these events to BQ, so this change is not necessary. We will track it on Google Analytics. Thanks

web/partials/displays/display-details.html
alex-deaconu: @RV-Adrian I don't think that's good. These `.app-header` classes can be used elsewhere and break unknown components/apps. Can't this be avoided? @ezequielc thoughts? Thanks.

web/partials/displays/alerts.html
```diff
@@ -1,12 +1,14 @@
...
+		<div class="banner-block">
+			<span class="add-right half-top text-muted" ng-if="!termsAccepted" translate>displays-app.alerts.acceptTerms</span>
+		</div>
 		<div class="title-block">
 			<!-- App Title -->
 			<h1 id="title" translate>displays-app.alerts.title</h1>
 		</div>
 
 		<div class="global-actions">
-			<span class="add-right text-muted" ng-if="!termsAccepted" translate>displays-app.alerts.acceptTerms</span>
```
ezequielc: @RV-Adrian Any reason why we are removing this? Thanks

RV-Adrian: @ezequielc I moved it into the center column because it did not fit properly in a 3-column layout.

web/partials/displays/display-details.html
RV-Adrian: @alex-deaconu @ezequielc This is where a seemingly small change hits a wall with an established static layout. Other options could be fluid columns or positioned elements, all work with further impact. Let me know your thoughts, thanks.

alex-deaconu: @RV-Adrian if we can't work around this with `.app-header`, I'd rather have a separate 3-column `.app-header-2` class then these issues & required workarounds. Seems less risky.

ezequielc: @RV-Adrian Yeah, it was supposed to be a quick experiment... :P What if we go back to your original idea, of a row on top(https://s3.amazonaws.com/uploads.hipchat.com/106443/1986381/21g3cQaZ97eqipj/upload.png)? Maybe if we reduce the margin it would look a bit better and we can go with it?
@alex-deaconu fyi.

RV-Adrian: @ezequielc @alex-deaconu  The row seems the safest bet (and to me, looks a little neater). But using the *-2 convention could also work if we're going to see what sticks and then judge after the experiment. (plus, bulk of work is done)
I'm fairly confident i got all .app-header instances covered, but didn't go into components etc.
Good with either approach 👍

ezequielc: @RV-Adrian I recommend going for the extra row. Using  *-2 will add technical debt. If the experiment succeed we can revisit all .app-headers or even stick with the extra row. Thank

RV-Adrian: @ezequielc True! Sounds like a plan

Issue #323, opened 2017-02-02
Can't copy DisplayID from Display Settings page


alanclayton:
Copying display ID's from the Display Settings page is a common function in Support to avoid having to manually type IDs.

Something changed recently where the table data that includes display id can no longer be copied.

RV-Adrian:
@alanclayton Issue is fixed and deployed, please confirm if ok? Thanks!

alanclayton:
@RV-Adrian Looks good. Thanks

PR #327, opened 2017-02-07
Fix/text tables

RV-Adrian:


RV-Adrian:
@alex-deaconu @ezequielc @fjvallarino 
Please validate fix for Issue https://github.com/Rise-Vision/rise-vision-apps/issues/323 
on https://apps-stage-3.risevision.com/displays/details/VH3CKAHZQ5P7

The Display ID cell should show a Text cursor on hover, and be selectable with a double-click.

ezequielc:
@RV-Adrian I am wondering if we should have the same behaviour for the whole row as we might need to select other cells there too. Thoughts?

RV-Adrian:
@ezequielc I was planning to change the behaviour of the Pointer cursor on the other table columns (not a bug necessarily, but seems odd). Can we tackle Table text selection outside of this Issue? (since it did focus on Display ID) Thx

ezequielc:
@RV-Adrian Sounds good, we can release it as it will fix the issue, but we should add a card to our tech debt. Thanks!

ezequielc:
@RV-Adrian 👍

RV-Adrian:
@ezequielc Makes sense! Thanks

alex-deaconu:
@RV-Adrian my question is why those table rows are non-selectable in general? I think you should address that as well as the cursor in the future work. Thanks.

RV-Adrian:
@alex-deaconu Sorry, yes that's what I meant -- why they're indicating a 'click' function and not typical HTML text 👍

PR #328, opened 2017-02-10
Fix sourcemaps

fjvallarino:
After trying way too many options which partially worked or completely failed, the best solution I found is in the latest commit. The interaction of sourcemaps and uglify cause issues, unless compression is disabled. If "compression" is set to false, the source map works fine. I tried a more fine-grained solution, and with the following set of options it still seems correct (I can reduce the number of options and only leave the non-default ones). This is staged at: https://apps-stage-1.risevision.com


alex-deaconu:
@fjvallarino the size of the website.min.js file in your version is 2.5mb vs 375kb from Production. I don't think we can go forward with this as it would introduce longer load times for the app. Thanks.

fjvallarino:
@alex-deaconu sorry, during one of the tests I removed the '.' from sourcemaps.write, which caused the files to be merged together, apparently (I tested size at the beginning, then forgot about it). It should be fine now.

I also realized that storage-selector.min.js was not being generated. I had to refactor the task a bit, but it is now being built.

alex-deaconu:
@fjvallarino thanks for fixing, good catch re: storage-selector.min.js, I had not verified that file. I think you need to rebase your branch off master, otherwise looks good. Thanks!

PR #344, opened 2017-02-24
Fix store-url config in CH

alex-deaconu:
[stage-4]


alex-deaconu:
Staged at https://apps-stage-0.risevision.com/. 

We definitely have an issue with staging to a specific url. This was supposed to be on stage-4 as you can see in the commit message. I will add a card.

@ezequielc @fjvallarino please review. Thanks.

ezequielc:
@alex-deaconu 👍 Thanks for fixing

PR #345, opened 2017-02-24
Handle errors in presentation js code

fjvallarino:
This is on stage-2

@apps please review


web/scripts/editor/services/svc-editor-factory.js
```diff
@@ -128,29 +129,42 @@ angular.module('risevision.editor.services')
...
 
         distributionParser.updateDistribution(factory.presentation);
 
         _updateEmbeddedIds(factory.presentation);
+
+        return parseResult;
+      };
+
+      factory.validatePresentation = function () {
+        return presentationParser.parsePresentation(factory.presentation);
```
alex-deaconu: @fjvallarino why not show the messageBox error here and not duplicate that code everywhere else. Thanks.

web/scripts/editor/services/svc-presentation-parser.js
```diff
@@ -205,15 +205,23 @@ angular.module('risevision.editor.services')
...
             json = json.replace('\\\'', '\'');
 
-            var dataObject = JSON.parse(json);
+            try {
+              dataObject = JSON.parse(json);
+            }
+            catch (err) {
+              $log.error('parse presentation error', err);
+              $log.error('failing json', json);
+              return { jsonParseError: true };
```
alex-deaconu: @fjvallarino what's the point of returning this weird object? I think `true`/`false` return would suffice. Thanks.

web/scripts/editor/services/svc-presentation-parser.js
```diff
@@ -391,16 +401,21 @@ angular.module('risevision.editor.services')
...
         end = htmlString.toLowerCase().indexOf(bodyEndTag, start);
         if (start === -1 || end === -1) {
-          return;
+          return {};
         }
 
         // process data
-        factory.parsePresentationData(presentation);
+        if(factory.parsePresentationData(presentation).jsonParseError) {
+          $log.debug('error parsing presentation', presentation);
```
alex-deaconu: @fjvallarino parse error is logging 3 console messages. I don't think that's necessary. Thanks.

web/scripts/editor/services/svc-presentation-parser.js
alex-deaconu: Or maybe allowing the error to flow up to the calling function and catching it there.

alex-deaconu:
@fjvallarino I made some suggestions in the code, but overall it seems to be touching too many pieces of code for 1 seemingly simple validation.

Also, I would have hoped to be able to re-use some sort of message box component. Maybe we can just simply use an Alert? 

Up to you if you want to follow up on this comment or leave it as is. 

Thanks.

web/scripts/editor/services/svc-presentation-parser.js
fjvallarino: @alex-deaconu agreed, those are debugging statements I forgot to remove.

web/scripts/editor/services/svc-editor-factory.js
fjvallarino: @alex-deaconu mainly because I wanted to keep the data/error flow on addPresentation/updatePresentation as it is, without breaking the promise based mechanism (in those functions, the deferred needs to be returned even if the validation fails). I'll re-think that part a bit, since I should be able to do that process here and still return the promise.

web/scripts/editor/services/svc-presentation-parser.js
fjvallarino: @alex-deaconu I added the object to be explicit about what could be failing, in case other kind of validation was needed. Using true/false sounds good too, but I'd rather avoid bubbling the exception up.

web/scripts/common/services/svc-message-box.js
```diff
@@ -0,0 +1,25 @@
+'use strict';
+
+angular.module('risevision.apps.services')
+  .factory('messageBox', ['$q', '$log', '$modal', '$templateCache',
```
alex-deaconu: @fjvallarino curious, why not include this service in common-app-components too? Thanks.

fjvallarino: @alex-deaconu makes sense. I did not add it because it was some sort of simple utility, but I'll move it there.

web/scripts/editor/services/svc-editor-factory.js
alex-deaconu: @fjvallarino okay, actually, you make a good point. Why does this not return a promise instead of an object?

I think if you open the message box here, and return a promise you can simplify the code down the line: ie.

```
_parseOrUpdatePresentation().then(
  return presentation.add(...)
)
```

fjvallarino:
@Rise-Vision/apps updated to reflect suggestions/comments

ezequielc:
@fjvallarino 👍 Thanks

PR #348, opened 2017-03-02
Fix issue when saving presentation which causes content changes to be reverted

fjvallarino:


ezequielc:
@fjvallarino 👍 Thanks for fixing it. Cheers

PR #349, opened 2017-03-03
Async call was causing empty Trash label. Same for filetype Folder

ezequielc:
Trash label was not showing.
@fjvallarino Please review. Thanks


fjvallarino:
:+1:

PR #351, opened 2017-03-06
Fix Status message

ezequielc:
@fjvallarino Please review. It is a bit hard to test as we don't have the widgets and subscriptions in our test env. 
If you really want to double check it, you can load the branch locally and update your config.js to point to production.
Please review. Thanks!


fjvallarino:
:+1:

PR #352, opened 2017-03-06
Feature/video experiment

RV-Adrian:


RV-Adrian:
@fjvallarino @alex-deaconu 
I am having issues with this branch. I took a look and the Schedule does seem to get created, despite the error. Will bring this up in standup tomorrow, thanks.
https://apps-stage-3.risevision.com

@ezequielc FYI

RV-Adrian:
@alex-deaconu @fjvallarino 
e2e issue fixed, PR good for review now 🎆 
https://apps-stage-3.risevision.com

@ezequielc I believe this PR would also cover https://trello.com/c/VeWVSbQ6/2076-revert-blank-presentation-experiment-1

fjvallarino:
@RV-Adrian looks good! I assume we're not using common-18n because this is an experiment. If that's expected, then :+1: 

Only concern is the video is not in High Definition.

RV-Adrian:
@fjvallarino Good point. I assumed not to use common-i18n as well, but will do so if there's a good reason.

ezequielc:
@RV-Adrian As it is an experiment, we are good to not use common-i18n. We are probably changing this page again next week. Thanks

RV-Adrian:
@ezequielc Sounds good

alex-deaconu:
@RV-Adrian I have a similar concern regarding the inline styles. Can't you re-use a common-style class for the paddings? Thanks.

RV-Adrian:
@alex-deaconu Missed that one. Easily removed (didn't really do much).
Thanks!

alex-deaconu:
@RV-Adrian got it, thanks for fixing! Looks good 👍.

PR #360, opened 2017-03-09
Fix a couple of issues when navigating to a trash subfolder

ezequielc:
Trash subfolders were not being handled as being inside trash, therefore restore and delete forever actions were not available. Also, upload buttons were enabled when they shouldn't be.

[stage-4]


fjvallarino:
:+1:

ezequielc:
@alex-deaconu This got staged at stage-0 somehow, maybe your fix didn't work? Thanks

alex-deaconu:
@ezequielc you're right. Looks like it didn't work this time either. Will investigate. Thanks.

PR #364, opened 2017-03-14
Add rename feature

fjvallarino:
This can be tested on: https://apps-stage-0.risevision.com

### DIFF
--- a/web/partials/storage/break-link-warning-modal.html
+++ b/web/partials/storage/break-link-warning-modal.html
@@ -1,0 +1,32 @@
+<div id="breakLinkModal">
+  <div class="modal-header">
+    <button type="button" class="close" ng-click="cancel()" data-dismiss="modal" aria-hidden="true">
+      <i class="fa fa-times"></i>
+    </button>
+    <h3 class="modal-title">{{ 'storage-client.rename.warning' | translate }}</h3>
+  </div>
+  <div class="modal-body" stop-event="touchend" autofocus="autofocus">
+    <div>
+      <p>{{ 'storage-client.rename.breakingLink1' | translate }}</p>
+      <p>{{ 'storage-client.rename.breakingLink2' | translate }}</p>
+    </div>
+  </div>
+  <div class="modal-footer">
+    <div class="col-sm-6 col-xs-12">
+      <!--
+      <div class="checkbox text-left">
+        <label>
+          <input name="choice" type="checkbox" ng-model="background.useImage" class="ng-pristine ng-untouched ng-valid ng-empty">
+          Do not show this warning again
+        </label>
+      </div>
+      -->
+    </div>
+    <button id="renameFileFolderModalSaveButton" class="btn btn-primary" ng-click="ok()">
+      <span>{{ 'common.proceed' | translate }}</span> <i class="fa fa-white fa-check icon-right"></i>
+    </button>
+    <button id="renameFileFolderModalCancelButton" class="btn btn-default" ng-click="cancel()">
+      <span>{{ 'common.cancel' | translate }}</span> <i class="fa fa-white fa-times icon-right"></i>
+    </button>
+  </div>
+</div>
--- a/web/partials/storage/break-link-warning-modal.html
+++ b/web/partials/storage/break-link-warning-modal.html
@@ -7,20 +7,18 @@
   </div>
   <div class="modal-body" stop-event="touchend" autofocus="autofocus">
     <div>
-      <p>{{ 'storage-client.rename.breakingLink1' | translate }}</p>
-      <p>{{ 'storage-client.rename.breakingLink2' | translate }}</p>
+      <p>{{ 'storage-client.rename.breaking-link1' | translate }}</p>
+      <p>{{ 'storage-client.rename.breaking-link2' | translate }}</p>
     </div>
   </div>
   <div class="modal-footer">
     <div class="col-sm-6 col-xs-12">
-      <!--
       <div class="checkbox text-left">
         <label>
-          <input name="choice" type="checkbox" ng-model="background.useImage" class="ng-pristine ng-untouched ng-valid ng-empty">
-          Do not show this warning again
+          <input type="checkbox" ng-model="hideWarning">
+          {{ 'storage-client.rename.breaking-link-hide-warning' | translate }}
         </label>
       </div>
-      -->
     </div>
     <button id="renameFileFolderModalSaveButton" class="btn btn-primary" ng-click="ok()">
       <span>{{ 'common.proceed' | translate }}</span> <i class="fa fa-white fa-check icon-right"></i>
--- a/web/partials/storage/break-link-warning-modal.html
+++ b/web/partials/storage/break-link-warning-modal.html
@@ -7,8 +7,8 @@
   </div>
   <div class="modal-body" stop-event="touchend" autofocus="autofocus">
     <div>
-      <p>{{ 'storage-client.rename.breaking-link1' | translate }}</p>
-      <p>{{ 'storage-client.rename.breaking-link2' | translate }}</p>
+      <p>{{ infoLine1Key | translate }}</p>
+      <p>{{ infoLine2Key | translate }}</p>
     </div>
   </div>
   <div class="modal-footer">
@@ -16,15 +16,15 @@
       <div class="checkbox text-left">
         <label>
           <input type="checkbox" ng-model="hideWarning">
-          {{ 'storage-client.rename.breaking-link-hide-warning' | translate }}
+          {{ warningKey | translate }}
         </label>
       </div>
     </div>
     <button id="renameFileFolderModalSaveButton" class="btn btn-primary" ng-click="ok()">
-      <span>{{ 'common.proceed' | translate }}</span> <i class="fa fa-white fa-check icon-right"></i>
+      <span>{{ confirmKey | translate }}</span> <i class="fa fa-white fa-check icon-right"></i>
     </button>
     <button id="renameFileFolderModalCancelButton" class="btn btn-default" ng-click="cancel()">
-      <span>{{ 'common.cancel' | translate }}</span> <i class="fa fa-white fa-times icon-right"></i>
+      <span>{{ cancelKey | translate }}</span> <i class="fa fa-white fa-times icon-right"></i>
     </button>
   </div>
 </div>
--- a/web/partials/storage/break-link-warning-modal.html
+++ b/web/partials/storage/break-link-warning-modal.html
@@ -3,12 +3,12 @@
     <button type="button" class="close" ng-click="cancel()" data-dismiss="modal" aria-hidden="true">
       <i class="fa fa-times"></i>
     </button>
-    <h3 class="modal-title">{{ 'storage-client.rename.warning' | translate }}</h3>
+    <h3 class="modal-title" translate>storage-client.rename.warning</h3>
   </div>
   <div class="modal-body" stop-event="touchend" autofocus="autofocus">
     <div>
-      <p>{{ infoLine1Key | translate }}</p>
-      <p>{{ infoLine2Key | translate }}</p>
+      <p translate>{{ infoLine1Key }}</p>
+      <p translate>{{ infoLine2Key }}</p>
     </div>
   </div>
   <div class="modal-footer">
@@ -16,15 +16,15 @@
       <div class="checkbox text-left">
         <label>
           <input type="checkbox" ng-model="hideWarning">
-          {{ warningKey | translate }}
+          <span translate>{{ warningKey }}</span>
         </label>
       </div>
     </div>
     <button id="renameFileFolderModalSaveButton" class="btn btn-primary" ng-click="ok()">
-      <span>{{ confirmKey | translate }}</span> <i class="fa fa-white fa-check icon-right"></i>
+      <span translate>{{ confirmKey }}</span> <i class="fa fa-white fa-check icon-right"></i>
     </button>
     <button id="renameFileFolderModalCancelButton" class="btn btn-default" ng-click="cancel()">
-      <span>{{ cancelKey | translate }}</span> <i class="fa fa-white fa-times icon-right"></i>
+      <span translate>{{ cancelKey }}</span> <i class="fa fa-white fa-times icon-right"></i>
     </button>
   </div>
 </div>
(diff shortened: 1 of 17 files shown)


web/scripts/storage/controllers/ctr-break-link-warning-modal.js
```diff
@@ -0,0 +1,21 @@
+'use strict';
+
+angular.module('risevision.storage.controllers')
+  .controller('BreakLinkWarningModalCtrl', ['$scope', '$modalInstance', 'localStorageService',
+    function ($scope, $modalInstance, localStorageService) {
+      $scope.hideWarning = false;
+
+      $scope.ok = function () {
+        localStorageService.set('breakingLinkWarning.hideWarning', $scope.hideWarning);
```
alex-deaconu: @fjvallarino wondering if we can make this a common component, abstracted from this particular Storage functionality. I can imagine it could be useful for other warnings as well (ie. Deprecated content in a Presentation). Thanks.

web/scripts/storage/controllers/ctr-rename-modal.js
```diff
@@ -0,0 +1,82 @@
...
+    function ($scope, $modalInstance, $q, storage, filesFactory, fileSelectorFactory, source) {
+      $scope.renameName = source.name.replace("/", "");
+      $scope.isProcessing = false;
+
+      function isFile(name) {
+        return name.lastIndexOf("/") !== name.length - 1;
+      }
+
+      function clone(obj) {
+        return JSON.parse(JSON.stringify(obj));
```
alex-deaconu: @fjvallarino is this utility function necessary? Why not just use Angular.copy? Thanks.

web/scripts/storage/controllers/ctr-rename-modal.js
```diff
@@ -0,0 +1,82 @@
+'use strict';
+
+angular.module('risevision.storage.controllers')
+  .controller('RenameModalCtrl', ['$scope', '$modalInstance', '$q', 'storage', 'filesFactory', 'fileSelectorFactory', 'sourceName',
+    function ($scope, $modalInstance, $q, storage, filesFactory, fileSelectorFactory, source) {
+      $scope.renameName = source.name.replace("/", "");
+      $scope.isProcessing = false;
+
+      function isFile(name) {
+        return name.lastIndexOf("/") !== name.length - 1;
```
alex-deaconu: @fjvallarino I believe the StorageFactory already has a similar function: https://github.com/Rise-Vision/rise-vision-apps/blob/master/web/scripts/storage/services/svc-storage-factory.js#L120. Thanks.

web/scripts/storage/controllers/ctr-rename-modal.js
```diff
@@ -0,0 +1,82 @@
...
+      $scope.ok = function () {
+        var suffix = isFile(source.name) ? "" : "/";
+        var renameName = $scope.renameName + suffix;
+        var newFile = clone(source);
+
+        $scope.errorKey = null;
+        $scope.isProcessing = true;
+
+        return storage.rename(source.name, renameName)
+          .then(function(resp) {
```
alex-deaconu: @fjvallarino I don't believe this logic belongs in the controller. We have a FileActionsFactory you can add this to, or create a new factory for this. Thanks.

web/scripts/storage/controllers/ctr-rename-modal.js
```diff
@@ -0,0 +1,82 @@
...
+            $scope.isProcessing = false;
+          });
+      };
+
+      $scope.cancel = function () {
+        $modalInstance.dismiss('cancel');
+      };
+
+      $scope.dismiss = function () {
+        $modalInstance.dismiss();
```
alex-deaconu: @fjvallarino what's the difference between cancel and delete here, do we need 2 functions that do the same thing? Thanks.

alex-deaconu:
@fjvallarino reviewed. Couple of comments.

Preferably we should add as little code as possible that cannot be reused and we should keep business logic out of controllers as much as possible. 

Thanks.

fjvallarino:
@Rise-Vision/apps updated with PR suggestions.

There is a build issue because of a transitive dependency (this is new, apparently something we depend on now requires a newer Node version). I will look into it in the morning.

web/scripts/storage/services/svc-file-actions-factory.js
```diff
@@ -175,6 +175,91 @@ angular.module('risevision.storage.services')
...
+                  fileSelectorFactory.resetSelections();
+
+                  return resp;
+                });
+            }
+          });
+      };
+
+      factory.showBreakLinkWarning = function (infoLine1Key, infoLine2Key, warningKey, confirmKey, cancelKey, localStorageKey) {
+        var hideWarning = localStorageService.get(localStorageKey) === 'true';
```
alex-deaconu: @fjvallarino do you think there's a way to move this logic over to the modal controller as well? If not, disregard. Thanks.

web/scripts/storage/controllers/ctr-rename-modal.js
```diff
@@ -0,0 +1,39 @@
+'use strict';
+
+angular.module('risevision.storage.controllers')
+  .controller('RenameModalCtrl', ['$scope', '$modalInstance', 'fileActionsFactory', 'sourceObject',
+    function ($scope, $modalInstance, fileActionsFactory, sourceObject) {
+      $scope.renameName = sourceObject.name.replace("/", "");
```
alex-deaconu: @fjvallarino quick question, is `sourceObject.name` just the filename, or does it actually contain folder information as well? ie. if the file is in a subfolder would this be `file.jpg` or `folder/sub-folder/file.jpg`? Thanks.

web/partials/storage/break-link-warning-modal.html
```diff
@@ -0,0 +1,30 @@
...
+          <input type="checkbox" ng-model="hideWarning">
+          {{ warningKey | translate }}
+        </label>
+      </div>
+    </div>
+    <button id="renameFileFolderModalSaveButton" class="btn btn-primary" ng-click="ok()">
+      <span>{{ confirmKey | translate }}</span> <i class="fa fa-white fa-check icon-right"></i>
+    </button>
+    <button id="renameFileFolderModalCancelButton" class="btn btn-default" ng-click="cancel()">
+      <span>{{ cancelKey | translate }}</span> <i class="fa fa-white fa-times icon-right"></i>
```
alex-deaconu: @fjvallarino from what I recall, the recommended method for using angular-translate efficiently is via the directive (as opposed to the filter), as it is less resource intensive. Wondering if we can do that here, or there's a specific reason for using the filter. Thanks.

web/scripts/storage/controllers/ctr-rename-modal.js
alex-deaconu: @fjvallarino actually this is a concern. This logic does not work nicely with files within sub-folders, see attached:
screen shot 2017-03-15 at 11 04 49 am

Thanks.

web/scripts/storage/services/svc-file-actions-factory.js
fjvallarino: @alex-deaconu I'm thinking about it, but I guess if it's possible I will need to close the modal as soon as it's open (I don't think I can avoid having the modal window open from the modal controller itself)

alex-deaconu: @fjvallarino I thought so. I'm good with leaving as is, or if you want to wrap that part in a service similar to message-box. Thanks.

web/scripts/storage/controllers/ctr-rename-modal.js
alex-deaconu: @fjvallarino as a solution, I would suggest just showing the file/folder name without the underlaying folder structure. It would reduce confusion, if they want to rename and move they would have to perform 2 separate operations. @ezequielc thoughts? Thanks.

fjvallarino: @alex-deaconu good catch. I'll fix now

web/partials/storage/break-link-warning-modal.html
fjvallarino: @alex-deaconu I think it should work. My only concern is cancelKey is a variable, not a key, but  Angular should expand it before passing the value to the directive.

web/scripts/storage/controllers/ctr-rename-modal.js
fjvallarino: @alex-deaconu yeah, that was my plan. Allowing a folder rename would be a move, which is a different operation. The plan was always showing just the name, but I missed the fact that name included the full path (bad testing, basically)

web/partials/storage/break-link-warning-modal.html
alex-deaconu: @fjvallarino okay; we use the same methodology in message-box without issues (https://github.com/Rise-Vision/common-app-components/blob/master/src/message-box/message-box.html). Just to confirm, I was referring to all i18n strings in this PR, forgot to specify that. Thanks.

web/scripts/storage/controllers/ctr-rename-modal.js
alex-deaconu: @fjvallarino sounds good. Maybe you can use the `fileNameFilter` functionality to prevent duplicate logic. Thanks.

ezequielc: @alex-deaconu @fjvallarino Agreed, we should show only the name. Thanks

ezequielc:
@fjvallarino I reviewed the functionality and apart from the subfolder issue that Alex mentioned, everything else works fine. There are some improvements we can work on, but they are minor improvements and I think we should release without them so we can release now, middle of the week.

Suggested improvements are:
- Autofocus and select filename when Rename modal opens
- Add a spinner on Save button after clicked
- Add a filename validation message when using "/"
- Queue thumbnails when renaming a folder. The first time I enter the folder after renaming it, I don't see thumbnails.

fjvallarino:
@ezequielc my comments:

- I thought about the autofocus, but I'm not sure we're using it anywhere else. If we already have a directive for that, I can just apply it.
- I tried using the file list spinner but it looked awful. I was not sure if we wanted a smaller spinner on the dialog box, that's why I disabled the buttons instead.
- Do you mean, show an error while typing, besides disabling the button?
- The storage server rename actually queues the thumbnail generation, but maybe it's not running as quickly as needed.

ezequielc:
@fjvallarino Thanks

1. I think we had a directive for that, it is called something like focus-me. 
2. I was something more subtle, like replacing the tick in the save button to a spinner. Or we could just rename Save to Saving, like we do when saving Presentations.
3. Yes, a validation error like we do for background images in Presentation Properties.
4. Ok, sounds good. I will double check.

Thank you!

alex-deaconu:
@ezequielc @fjvallarino I think the `focus-me` directive requires a trigger variable (ie. conditional focus). I suggest you just use the `auto-focus` property, for example: https://github.com/Rise-Vision/rise-vision-apps/blob/master/web/partials/storage/copy-url-modal.html. Thanks.

fjvallarino:
@Rise-Vision/apps changes applied. The rebase caused losing history. The new commits begin on "Fix issue with renaming subfolders".

web/scripts/storage/services/svc-file-actions-factory.js
alex-deaconu: @fjvallarino okay, thanks.

web/scripts/storage/services/svc-file-actions-factory.js
```diff
@@ -175,6 +175,91 @@ angular.module('risevision.storage.services')
...
+
+        return storage.rename(sourceObject.name, renameName)
+          .then(function(resp) {
+            if(resp.code !== 200) {
+              return resp;
+            }
+            else {
+              newObject.name = renameName;
+
+              return factory.refreshThumbnail(newObject)
```
alex-deaconu: @fjvallarino curious, does `refreshThumbnail` work correctly for a folder? Also, wondering if it were more beneficial if we moved this functionality to the server side (ie. enqueue an asynchronous task) or made this API call asynchronous. This way the response to the User is not delayed by the extra call. Thanks.

fjvallarino: @alex-deaconu the only thing that method does is force a reload of the file, using a server side method which generates the thumbnail if one does not exist (the endpoint already existed and it was not modified). For a folder it would not generate anything, that's why it just resolves a promise and avoids the server call. Originally this call did not exist; it was suggested by @ezequielc to have the thumbnail refreshed immediately (since the server side task, which is indeed created, may not run as soon as needed).

web/scripts/storage/services/svc-file-actions-factory.js
```diff
@@ -175,6 +175,91 @@ angular.module('risevision.storage.services')
...
+        }).result;
+      };
+
+      factory.showRenameBreakLinkWarning = function() {
+        var prefix = 'storage-client.rename.';
+
+        return factory.showBreakLinkWarning(prefix + 'breaking-link1',
+                                            prefix + 'breaking-link2',
+                                            prefix + 'breaking-link-hide-warning',
+                                            'common.proceed',
```
alex-deaconu: @ezequielc to confirm, we are introducing a new action word here: "Proceed". Is that intentional and is it included in our wording documentation? I would venture to say that "Okay" would do just fine here. @fjvallarino FYI. Thanks.

web/scripts/storage/services/svc-file-actions-factory.js
alex-deaconu: @fjvallarino sounds good. Thanks for the clarification.

alex-deaconu:
@fjvallarino one last comment re: wording. Looks good otherwise, thanks for the fixes!

web/scripts/storage/services/svc-file-actions-factory.js
fjvallarino: @alex-deaconu @ezequielc that's the word used in the prototype: https://github.com/Rise-Vision/rise-vision-apps/pull/331/files#diff-000d492c6ba708165d2e4bf4ad29f7f4R24

ezequielc: @alex-deaconu @fjvallarino Good point. I am good with either. If there are other changes required in the PR we can change this, otherwise we can go as is. Thanks

ezequielc:
@fjvallarino I reviewed the functionality and it looks good. I think we are good to go. Thanks!

PR #367, opened 2017-03-16
update support flow [stage-2]

settinghead:
Related: https://github.com/Rise-Vision/common-header/pull/666


settinghead:
Approved as per comment https://github.com/Rise-Vision/common-header/pull/666#issuecomment-287385759

PR #368, opened 2017-03-17
Avoid calling rename endpoint when provided the same name for source and destination

fjvallarino:


alex-deaconu:
@fjvallarino 👍 . Thanks for fixing!

PR #369, opened 2017-03-17
Feature/apps launcher experiment

ghislaineguerin:
@alex-deaconu 
Can you take a look and let me know your thoughts and how we can proceed from here?
Also, I assume there's a better way to link to the video, image and blank templates?

Thanks!

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -299,7 +299,7 @@ gulp.task("test:e2e", function (cb) {
 gulp.task("metrics", factory.metrics());
 
 gulp.task("test",  function (cb) {
-  runSequence(["config", "html2js"], ["test:unit"], "coveralls", cb);
+  runSequence(["config", "html2js"], ["test:unit", "test:e2e"], "coveralls", cb);
 });
 
 gulp.task("test:ci",  function (cb) {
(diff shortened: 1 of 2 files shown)


ghislaineguerin:
@alex-deaconu I have added your suggestions. Please take a look. thanks

alex-deaconu:
Been working with @ghislaineguerin to fix this integration. Basically, we are setting a flag in Optimizely on the $rootScope and inside the partial we are checking that flag. 

The Optimizely script is quite simple:

```
var $rootScope = angular.element(document.body).injector().get('$rootScope');

$rootScope.presentationTemplateCtaExperiment = true;
```

and as you can see in the code we just use an `ng-if` to select which div to show.

Staging link:
https://apps-stage-4.risevision.com/?optimizely_x8290831837=1&cid=34bc012f-bb3a-4005-9744-27df580ee636

@ezequielc @fjvallarino please review the above. Thanks!

fjvallarino:
@alex-deaconu @ghislaineguerin is there anything special I should be doing to get the alternative page? I tried a bunch of times in Incognito mode (closing the window, not closing it, any combination) and I got the regular "How Simple is Rise Vision?" every time.

ghislaineguerin:
@fjvallarino I see the variation, could you try refreshing many times? or appending this parameter 
optimizely_token=PUBLIC

ezequielc:
@ghislaineguerin @alex-deaconu I can see both versions. Looks good. Thanks! 👍

PR #375, opened 2017-03-24
Update Templates Library ad message

ezequielc:
[stage-1]
Updated ad message and added it to Templates modal.
@fjvallarino Please review. Thanks!


fjvallarino:
:+1:

PR #379, opened 2017-03-27
Open Presentation Properties when clicking presentation name

ezequielc:
@fjvallarino Please review. Thanks

### DIFF
--- a/test/e2e/editor/cases/presentation-properties.js
+++ b/test/e2e/editor/cases/presentation-properties.js
@@ -45,8 +45,19 @@ var PresentationPropertiesScenarios = function() {
       });
 
       describe("Given a user clicks on the presentation properties cog icon", function () {
-        it('should open the properties modal', function () {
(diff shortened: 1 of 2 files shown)


fjvallarino:
@ezequielc would it make sense to switch the cursor to indicate the element is clickable?

ezequielc:
@fjvallarino Yes, I think so. 
I also checked google docs and they keep the same text cursor, but highlight the element. But it is an inline edit in their case.
@ghislaineguerin What to you think? Thanks

ezequielc:
@ghislaineguerin Actually, if you think we should add the pointer, we could merge this PR and add the pointer in your open CS PR. Thoughts? Thanks

ghislaineguerin:
@ezequielc that's a good idea. I will add the pointer to my common styles PR.

fjvallarino:
:+1:

PR #382, opened 2017-03-29
Update to CH 1.13.1 to honor supscription status for the remaining valid period after cancellation [stage-2]

settinghead:


settinghead:
@alex-deaconu @ezequielc or @fjvallarino please review. Thanks.

bower.json
```diff
@@ -16,7 +16,7 @@
     "angular-ui-codemirror": "~0.3.0",
     "angular-vertilize": "~1.0.1",
     "angular-mocks": "1.5.0",
-    "common-header": "https://github.com/Rise-Vision/common-header.git#v1.13.0",
+    "common-header": "https://github.com/Rise-Vision/common-header.git#1.13.1",
```
alex-deaconu: @settinghead this works even without the `v`? Thanks.

settinghead: @alex-deaconu good catch. It's a typo. No idea why the CI passed. Will fix.

alex-deaconu: @settinghead yeah I was wondering the same thing. From what I can tell, it behaves the same with or without the `v`. Interesting. I guess for consistency we should keep it however.

Thanks for fixing!

settinghead: Interesting to know!

PR #383, opened 2017-04-05
Add MAC Address field and UX changes to Displays page

fjvallarino:

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -18,7 +18,7 @@
     "angular-mocks": "1.5.0",
     "common-header": "https://github.com/Rise-Vision/common-header.git#v1.13.1",
     "component-storage-selector": "https://github.com/Rise-Vision/component-storage-selector.git",
-    "rv-common-i18n": "https://github.com/Rise-Vision/common-i18n.git#chore/add-mac-player-data",
+    "rv-common-i18n": "https://github.com/Rise-Vision/common-i18n.git",
     "rv-common-style": "https://github.com/Rise-Vision/common-style.git#v1.4.9",
     "rv-common-app-components": "https://github.com/Rise-Vision/common-app-components.git",
     "rv-loading": "https://github.com/Rise-Vision/rv-loading.git",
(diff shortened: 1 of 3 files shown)


alex-deaconu:
@Rise-Vision/apps the right aligned buttons look so weird to me: https://screencast.com/t/RDALDy7PlP7. Is that intentional? Thoughts? Thanks.

adrianDD:
@alex-deaconu @fjvallarino 
That's as prototyped, just to fall in line with the button locations above.
Full-width didn't add up here. 
I do think the 'Download Rise Player' component can use a re-do at some point.
image

alex-deaconu:
@adrianDD okay, no problem. I just thought right align looked weird & new pattern; currently the fields are right aligned so maybe I'm just used to that pattern. Thanks.

alex-deaconu:
@fjvallarino looks good 👍 .

Not sure if you want to merge or hold off until the "Hide empty table sections" card is implemented.

Thanks.

adrianDD:
@alex-deaconu woops, wrong git account :)

Yeah, I did consider it.. and our most common UI pattern is actually a right-align for related actions.. but I do agree the dropdown of buttons is a little jarring. I think we can do better overall with accessing these player files, and I'll note that 👍 

re: Hide tables, I think there's a bit of discussion that needs to happen on that one but I'll leave that up to Apps :)

fjvallarino:
@alex-deaconu @ezequielc what should I do regarding @RV-Adrian latest comment? 

> Hide tables, I think there's a bit of discussion that needs to happen on that one but I'll leave that up to Apps :)

RV-Adrian:
@fjvallarino @ezequielc 
I'll clarify a bit on the card on 'hide tables'. I think that's best-suited to tackle when the rest of the Offline Play UI is ready to go, because the idea is to hide entire sections when they are simply empty. 
With the UI on this PR, rows could be hidden.. but I think the usefulness is not as great (imo)
Thanks, and let me know if that doesn't add up.
image

ezequielc:
@RV-Adrian @fjvallarino I do think we can hide Serial Number and MAC address fields when empty for now. We can discuss hiding tables whenever we implement the next layout. Thanks

adrianDD:
@ezequielc Sounds good, thanks

fjvallarino:
@Rise-Vision/apps updated

alex-deaconu:
@fjvallarino looks good, thanks for fixing 👍

PR #384, opened 2017-04-06
Chore/storage refactor

alex-deaconu:


alex-deaconu:
@ezequielc @fjvallarino just a quick note, this PR contains the changes I have so far for the refactoring. To confirm, unit tests are disabled as they have not been updated. However, e2e tests are passing which indicates the functionality is working as before (at least what was included in the test coverage).

To summarize the changes up until now:
- Converted a whole bunch of Controller/partial combos into Directives
- Merged the fileSelectorFactory into the filesFactory; this could be seen as not ideal but simplifies the refactor and somehow all functionality in this factory is interrelated anyways
- Extracted all static functions from storageFactory and fileSelectorFactory (old one) into a storageUtils factory
- Moved selectedFolder constant to the filesFactory from the storageFactory
- Changed filesFactory, storageFactory & fileActionsFactory into services which can be instantiated
- Updated FilesListController to instantiate the factories and updated necessary references in partials/sub-controllers/directives
- Created a new storageIFrame controller
- All entry points for Storage (iFrame, $modal & storage full) now pass initialization parameters (type, filter) via $scope to the FilesListController through ng-init; FilesListController in turn initializes the StorageFactory instance with the parameters
- Refactored iframe & $modal controllers to return files on their own (no longer a factory responsibility)

Note:
- For convenience, Directives which do not declare their own scope re-use parent scope references to factory instances (ie. https://github.com/Rise-Vision/rise-vision-apps/pull/384/files#diff-1340b31a406437bb8cc377f4d40c495dR23).
- Because of so many partials (and some directives) that re-use the FileListController's scope variables, I attempted to use the `controller as` pattern (https://toddmotto.com/digging-into-angulars-controller-as-syntax/) to clarify where each function is coming from. However, it turns out the refactor for this would need to be quite extensive so I dropped this.

At this point I will start working on fixing unit tests. Please review & test out the new version at your convenience.

Thanks!

ezequielc:
@alex-deaconu I am starting to test now, one thing I noticed is a ';' appearing at the top of iframe selector: https://apps-stage-4.risevision.com/storage-selector.html.  I couldn't quickly identify where it is coming from. Could you take a look? Thanks

alex-deaconu:
@ezequielc good catch re: `;`, fixed via latest commit. Thanks!

ezequielc:
@alex-deaconu File renaming modal is not appearing. Error: 
Error: [$injector:unpr] Unknown provider: fileActionsFactoryProvider <- fileActionsFactory <- RenameModalCtrl <- RenameModalCtrl

alex-deaconu:
@ezequielc oops, fixed in the last commit. Thanks.

ezequielc:
@alex-deaconu I took a look at the good and it looks good overall. I will defer to @fjvallarino on the specifics. Thanks!

ezequielc:
@alex-deaconu Thanks for the fixes. Let me know when the build works so I can validate. Thanks!

alex-deaconu:
@ezequielc FYI, build finally passed. Thanks.

ezequielc:
@alex-deaconu Looks good. Thanks!

fjvallarino:
@alex-deaconu code looks good and everything works fine!

alex-deaconu:
@ezequielc @fjvallarino it turns out the tests were failing because of this particular piece of code:

```
    $provide.service('$q', function() {
      return Q;
    });
```

in the unit test file. Basically, mocking $q prevents the directive from loading. I actually tested with other directives and that's the case as well. I did not dig any deeper into this, but found it quite weird. In any case, this PR is now ready to go.

Wondering if you can do another quick review of the remaining commits. Thanks!

web/scripts/storage/services/svc-storage-factory.js
```diff
@@ -24,136 +24,98 @@ angular.module('risevision.storage.services')
...
-      factory.fileParent = function (file) {
-        var idx = file.name.length - (factory.fileIsFolder(file) ? 2 : 1);
+        var _fileHasExtension = function (file, allowedExtensions) {
+          var fileType = $filter('fileTypeFilter')(file.name);
+          return allowedExtensions.indexOf(fileType.toLowerCase()) !== -1;
+        };
 
-        return file.name.substr(0, file.name.lastIndexOf('/', idx) + 1);
-      };
+        factory.canSelect = function (file) {
```
ezequielc: @alex-deaconu I am wondering is this function and the next ones should be part of `storageUtils` as we have some similar functions there like `fileIsTrash()`, `fileIsFolder()`, etc.  Or would that cause cyclic dependency? Thanks

ezequielc:
@alex-deaconu I am wondering if you could stage at stage-0 so we can test the iframe mode when using image/video widgets, just to be safe. Looks good other than that, just a minor comment. Thanks!

alex-deaconu:
@ezequielc re-staging now to stage-0. 

Just FYI, you can also test the storage-selector at 
https://apps-stage-4.risevision.com/storage-selector.html
and add the selector-type parameter to obtain different types of selector. For example:
https://apps-stage-4.risevision.com/storage-selector.html#/?cid=b428b4e8-c8b9-41d5-8a10-b4193c789443&selector-type=multiple-file
other types are
single-file (default)
single-folder
multiple-files-folders

Thanks.

web/scripts/storage/services/svc-storage-factory.js
alex-deaconu: @ezequielc main issue is `storageUtils` is static, as opposed to `StorageFactory` instances which depend on two instance variables: type & filter. The remaining functions here are dependent on those two variables, so they couldn't be moved to the static utils class. Thanks.

ezequielc:
@alex-deaconu Thanks. Yeah, I had tested with the suggested URLs but I would like to confirm the postMessage functionality. Cheers!

alex-deaconu:
@ezequielc makes sense. Looks like the build passed & everything is staged. 

I will need to re-build again to rebase off my other commit... so please ignore the orange dot :)

web/scripts/storage/services/svc-storage-factory.js
ezequielc: @alex-deaconu Got it. Thanks for the clarifications!

ezequielc:
@alex-deaconu It is working fine 👍 Thanks!

fjvallarino:
@alex-deaconu :+1: !

PR #390, opened 2017-04-11
Remove gesture detection

ezequielc:

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -16,7 +16,7 @@
     "angular-ui-codemirror": "~0.3.0",
     "angular-vertilize": "~1.0.1",
     "angular-mocks": "1.5.0",
-    "common-header": "https://github.com/Rise-Vision/common-header.git#v1.13.1",
+    "common-header": "https://github.com/Rise-Vision/common-header.git#v1.13.2",
     "component-storage-selector": "https://github.com/Rise-Vision/component-storage-selector.git",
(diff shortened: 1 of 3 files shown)


alex-deaconu:
@ezequielc should probably remove the directive ref from here as well: https://github.com/Rise-Vision/rise-vision-apps/blob/master/web/scripts/storage-selector-app.js#L58. Thanks.

ezequielc:
@alex-deaconu Good catch. I will remove. Thanks

bower.json
```diff
@@ -16,7 +16,7 @@
     "angular-ui-codemirror": "~0.3.0",
     "angular-vertilize": "~1.0.1",
     "angular-mocks": "1.5.0",
-    "common-header": "https://github.com/Rise-Vision/common-header.git#v1.13.1",
+    "common-header": "https://github.com/Rise-Vision/common-header.git#v1.13.1#feature/remove-gesture",
```
alex-deaconu: @ezequielc I think this is a typo, but you're adding the branch name after the existing tag. I don't believe the branch is staged correctly in this case. Shouldn't be an issue in this case, just FYI. Thanks.

ezequielc: @alex-deaconu I will update with the tagged version. Thanks

alex-deaconu:

PR #394, opened 2017-04-12
Add duplicate functionality

fjvallarino:

### DIFF
--- a/web/scripts/storage/services/svc-file-actions-factory.js
+++ b/web/scripts/storage/services/svc-file-actions-factory.js
@@ -185,6 +185,31 @@ angular.module('risevision.storage.services')
             });
         };
 
+        factory.duplicateObject = function (sourceObject) {
+          var newObject = angular.copy(sourceObject);
+
+          pendingOperationsFactory.addPendingOperation(sourceObject, 'duplicate');
+
+          return storage.duplicate(sourceObject.name)
+            .then(function (resp) {
+              if (resp.code !== 200) {
+                return resp;
+              } else {
+                newObject.name = resp.message;
+
+                return factory.refreshThumbnail(newObject)
+                  .then(function (file) {
+                    pendingOperationsFactory.removePendingOperation(sourceObject);
+
+                    filesFactory.addFile(newObject);
+                    filesFactory.resetSelections();
+
+                    return resp;
+                  });
+              }
+            });
+        };
+
         factory.showBreakLinkWarning = function (infoLine1Key, infoLine2Key,
           warningKey, confirmKey, cancelKey, localStorageKey) {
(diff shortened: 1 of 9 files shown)


web/scripts/storage/services/svc-file-actions-factory.js
```diff
@@ -185,6 +185,31 @@ angular.module('risevision.storage.services')
...
 
+        factory.duplicateObject = function (sourceObject) {
+          var newObject = angular.copy(sourceObject);
+
+          pendingOperationsFactory.addPendingOperation(sourceObject, 'duplicate');
+
+          return storage.duplicate(sourceObject.name)
+            .then(function (resp) {
+              if (resp.code !== 200) {
+                return resp;
```
alex-deaconu: @fjvallarino I don't think just returning the `resp` object suffices here. We need to actually display an error in the UI via the pendingOperationsFactory. Thanks.

alex-deaconu:
As discussed in standup today, we can allow multiple items to be duplicated at once; we can implement that in a subsequent PR.

ezequielc:
@fjvallarino Functionality looks good.
Strangely, the sorting mechanism shows 'file (1)' before 'file'. I think it should show in the opposite order. But that is not related to this PR. We can consider it an improvement. Thanks

fjvallarino:
@Rise-Vision/apps updated. This is deployed at: https://apps-stage-0.risevision.com

web/scripts/storage/services/svc-file-actions-factory.js
```diff
@@ -250,6 +277,12 @@ angular.module('risevision.storage.services')
           });
         };
 
+        factory.duplicateButtonClick = function () {
+          return $q.all(filesFactory.getSelectedFiles().map(function(file) {
```
alex-deaconu: @fjvallarino I'm not sure if using `$q.all` here for an indeterminate number of files is the best approach. Maybe a sequential queue of files would be better. @ezequielc FYI. Thoughts? Thanks.

fjvallarino: @alex-deaconu the idea was being able to run the tasks in parallel. The browser should throttle the number of concurrent requests and delay them. I can change it to run sequentially, but it would take much longer to finish.

ezequielc: @alex-deaconu @fjvallarino I don't have a preference here. I tested with a large number of files and it seems that the requests were throttled by the browser. So  I am ok with this. Thanks

alex-deaconu: @ezequielc @fjvallarino sounds good. Thanks.

alex-deaconu:
@fjvallarino 👍

PR #395, opened 2017-04-13
Storage Move functionality [stage-4]

alex-deaconu:
Folder selector updates

Filter out selected folders
Allow root folder selection

Linting and beautify changes

Re-enable coveralls

Fix pendingOperations failure

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -18,7 +18,7 @@
     "angular-mocks": "1.5.0",
     "common-header": "https://github.com/Rise-Vision/common-header.git#v1.13.5",
     "component-storage-selector": "https://github.com/Rise-Vision/component-storage-selector.git",
-    "rv-common-i18n": "https://github.com/Rise-Vision/common-i18n.git#feature/storage-move",
+    "rv-common-i18n": "https://github.com/Rise-Vision/common-i18n.git",
     "rv-common-style": "https://github.com/Rise-Vision/common-style.git#v1.4.9",
     "rv-common-app-components": "https://github.com/Rise-Vision/common-app-components.git",
     "rv-loading": "https://github.com/Rise-Vision/rv-loading.git",
@@ -39,7 +39,6 @@
     "angular-sanitize": "~1.5.0",
     "angular-mocks": "~1.5.0",
     "angular-bootstrap": "^0.13.0",
-    "rv-common-style": "v1.4.9",
-    "rv-common-i18n": "feature/storage-move"
+    "rv-common-style": "v1.4.9"
   }
 }
--- a/bower.json
+++ b/bower.json
@@ -18,7 +18,7 @@
     "angular-mocks": "1.5.0",
     "common-header": "https://github.com/Rise-Vision/common-header.git#v1.13.5",
     "component-storage-selector": "https://github.com/Rise-Vision/component-storage-selector.git",
-    "rv-common-i18n": "https://github.com/Rise-Vision/common-i18n.git#v1.4.18",
+    "rv-common-i18n": "https://github.com/Rise-Vision/common-i18n.git#feature/storage-move",
     "rv-common-style": "https://github.com/Rise-Vision/common-style.git#v1.4.9",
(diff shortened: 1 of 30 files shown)


alex-deaconu:
This is ready for review. I looked/attempted to write e2e tests for this feature, but it requires extra work for testing folder creation & asset cleanup. I propose we add another card for that implementation (and include e2e tests for duplicate as well).

web/scripts/storage/services/svc-file-actions-factory.js
```diff
@@ -100,47 +96,63 @@ angular.module('risevision.storage.services')
...
+                'restricted-role') >= 0) {
+              $translate('storage-client.access-denied')
+                .then(function (msg) {
+                  pendingOperationsFactory.statusDetails.message = msg;
+                });
+            } else {
+              $translate('storage-client.' + resp.message, {
+                username: resp.userEmail
+              }).then(function (msg) {
+                pendingOperationsFactory.statusDetails.message = msg;
```
alex-deaconu: This piece of code requires the String to actually exist, if it doesn't the promise will never return and no message will be shown. I need to switch this to use `$filter`.

ezequielc:
@alex-deaconu I found a small issue: when a move operation fails (I tested moving to the same folder) when I try another move operation, the title of pending operations panel still shows the error message instead of 'Moving...'. Thanks

ezequielc:
@alex-deaconu Also, probably related to your inline comment above, when I try to move and another file already exists with the same name, the title of pending operations becomes empty instead of showing the error message. Thanks

alex-deaconu:
PR is updated. To confirm, I've fixed the issue with pending operations getting stuck on the error message. Also fixed another related issue.

In order to fix the second issue with the error message strings as well as the comment from the i18n PR, I've done some refactoring in common-i18n for Storage. Unfortunately, due to the fact that the Strings are returned from the Server, and used in 2 different error components, I tried a bunch of different things, but I couldn't prevent duplication.

@ezequielc @fjvallarino wondering if you could review the latest commits and let me know if acceptable. Thanks!

web/partials/storage/list-view.html
```diff
@@ -54,7 +54,7 @@
   <tbody class="table-body">
     <tr class="table-body__row"
     ng-click="fileClick(file);"
-    ng-class="{'active': file.isChecked, 'table-body__row--blocked-file': file.isThrottled, 'table-body__row--no-select': !storageFactory.canSelect(file), 'table-body__row--disabled': storageFactory.isDisabled(file) }"
+    ng-class="{'active': file.isChecked, 'table-body__row--error.blocked-file': file.isThrottled, 'table-body__row--no-select': !storageFactory.canSelect(file), 'table-body__row--disabled': storageFactory.isDisabled(file) }"
```
ezequielc: @alex-deaconu Is this change in the class name intentional?  Thanks

alex-deaconu: @ezequielc definitely not. Good catch, fixed!

ezequielc:
@alex-deaconu 👍  Thanks for fixing the issues. I had one minor comment, but I saw you are already fixing it. Thanks

alex-deaconu:
@ezequielc @fjvallarino rebased this and https://github.com/Rise-Vision/common-i18n/pull/222 off of the Duplicate changes. This should be good to go, wondering if you could do a final review. Thanks!

bower.json
```diff
@@ -18,7 +18,7 @@
     "angular-mocks": "1.5.0",
     "common-header": "https://github.com/Rise-Vision/common-header.git#v1.13.4",
     "component-storage-selector": "https://github.com/Rise-Vision/component-storage-selector.git",
-    "rv-common-i18n": "https://github.com/Rise-Vision/common-i18n.git#v1.4.18",
+    "rv-common-i18n": "https://github.com/Rise-Vision/common-i18n.git#feature/storage-move",
```
fjvallarino: @alex-deaconu maybe we should release a new common-i18n version

fjvallarino:
@alex-deaconu looks good! :+1:

PR #396, opened 2017-04-14
Screenshot disclaimer

ezequielc:
[stage-3]
As it is a temporary message (we will remove it once we capture the entire screen), I am not adding it to common-i18n.

@tejohnso @settinghead Could you please review as Apps members are off today? Thanks!


tejohnso:
:+1:

PR #398, opened 2017-04-18
Fix iOS issue with userState [stage-3]

### LINKED ISSUE #287: Storage App does not load on iOS device.
Reproduction Steps 
1. Go to storage.risevision.com
2. Select 'Sign In'
3. Select Account

Expected Results :
Storage App appears

Actual Results :
Loader spinner is endless (video attached)

Note: Android seems successful, while iOS does not.

[bug_ios.mov.zip](https://github.com/Rise-Vision/rise-vision-apps/files/622329/bug_ios.mov.zip)

alex-deaconu:
Fix #287 

@ezequielc @fjvallarino CH update for fixing the iOS issue. Please review.

@olegrise FYI. Verified and this issue occurs in the Store as well. New CH fixes it, so I would suggest updating for Store as well when you have a chance.

Thanks!

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -16,7 +16,7 @@
     "angular-ui-codemirror": "~0.3.0",
     "angular-vertilize": "~1.0.1",
     "angular-mocks": "1.5.0",
-    "common-header": "https://github.com/Rise-Vision/common-header.git#v1.13.3",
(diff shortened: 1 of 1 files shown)


ezequielc:
@alex-deaconu 👍

Issue #403, opened 2017-04-24
Sort in Storage is broken


sheadarlison:
Problem: 
You can only sort by name in Storage and the arrow stays on all the column header instead of disappearing when you try to sort a different column.

How it should work: 
You can sort by any column 

Video: 
https://www.screencast.com/t/h7AzEs0Zw8Ss

alex-deaconu:
@sheadarlison great find! Issue should be fixed now, please verify and close if acceptable. Thanks!

sheadarlison:
Looks good, thanks @alex-deaconu

PR #404, opened 2017-04-24
Fix issue with Storage sorting

alex-deaconu:
Most likely caused by the refactoring and mixed up scopes
Refactored search to match other lists and used scope function for sort


alex-deaconu:
Fix for #403 

fjvallarino:
@alex-deaconu looks good. The only concern is it breaks the fix (hack?) we added for the correct ordering of duplicates.

alex-deaconu:
@fjvallarino it should not, does it? It should be using the same filter function as before... Thanks.

fjvallarino:
@alex-deaconu sorry, it works fine. I don't know what I saw.

:+1:

PR #420, opened 2017-05-09
Increase line width from 80 to 120

fjvallarino:
This PR only contains the configuration change. It is part of the investigation card of an alternative formatting tool, which we plan on using in the future.


alex-deaconu:
@fjvallarino 👍

PR #422, opened 2017-05-09
Show spinner for Widget settings

alex-deaconu:
Spinner will close when getAdditionalParams is called
or after 3 seconds (in case the widget doesn't require additional params)

[stage-4]

### DIFF
--- a/test/e2e/editor/cases/add-content.js
+++ b/test/e2e/editor/cases/add-content.js
@@ -112,6 +112,8 @@ var AddContentScenarios = function() {
       
       it('should open Widget Settings', function () {
         helper.wait(widgetSettingsPage.getWidgetModal(), 'Widget Settings Modal');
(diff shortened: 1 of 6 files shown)


alex-deaconu:
The only communication that the Widget has with Apps is requesting the additionalParams. However, even that is not mandatory. Double checked with Delivery and all known widgets do request them. But, just in case there's one that doesn't, I added a timer to close the spinner if no request is received.

ezequielc:
@alex-deaconu Looks good 👍 
Just a minor comment, I noticed that the spinner slides down from the top, which is inconsistent from the other spinners which just appear. By looking at the code, it seems it is going to be tricky to change that, but it would be a good improvement. Anyway I am fine with deploying as is. Thanks

alex-deaconu:
@ezequielc interesting find. Actually, if I'm not mistaken, I think that transition is from the modal itself. Thanks.

ezequielc:
@alex-deaconu Yes, that is why I think it is not simple to fix. Thanks

PR #426, opened 2017-05-12
Improve upload process

fjvallarino:

### DIFF
--- a/test/unit/storage/services/svc-file-upload.tests.js
+++ b/test/unit/storage/services/svc-file-upload.tests.js
@@ -83,7 +83,27 @@ describe("Services: uploader", function() {
 
     });
   });
-  
+
+  it("removeAll: ", function () {
+    var files = [];
+
+    for(var i = 1; i <= uploader.queueLimit + 5; i++) {
+      files.push({ name: "test" + i + ".txt", webkitRelativePath: "folder/test" + i + ".txt", size: 200, type: "text" });
+    }
+
+    uploader.addToQueue(files);
+
+    expect(uploader.queue.length).to.equal(uploader.queueLimit);
+    expect(uploader.queue[0].file.name).to.equal("folder/test1.txt");
+    
+    uploader.removeAll();
+
+    expect(uploader.queue.length).to.equal(0);
+
+    $timeout.flush(500);
+
+    expect(uploader.queue.length).to.equal(0);
+  });
 
   describe('uploadItem:',function(){
     it("should invoke onBeforeUploadItem", function(done) {
--- a/test/unit/storage/services/svc-file-upload.tests.js
+++ b/test/unit/storage/services/svc-file-upload.tests.js
@@ -11,7 +11,7 @@ describe("Services: uploader", function() {
     });
   }));
   
-  var uploader, lastAddedFileItem;
+  var uploader, lastAddedFileItem, $timeout;
 
   beforeEach(function() {
   	inject(function($injector) {
@@ -27,6 +27,8 @@ describe("Services: uploader", function() {
       uploader.onBeforeUploadItem = function() {};
       uploader.onCancelItem = function() {};
       uploader.onCompleteItem = function() {};
+
+      $timeout = $injector.get("$timeout");
   	});
   });
 
@@ -49,7 +51,7 @@ describe("Services: uploader", function() {
       expect(uploader.queue[0].file.name).to.equal("folder/test1.txt");
     });
 
-    it("multiple files should be enqueued asynchronously after the first batch", function (done) {
+    it("multiple files should be enqueued asynchronously after the first batch", function () {
       var files = [];
 
       for(var i = 1; i <= uploader.queueLimit + 5; i++) {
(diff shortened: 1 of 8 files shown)


web/scripts/storage/services/svc-file-upload.js
```diff
@@ -19,45 +19,43 @@ angular.module('risevision.storage.services')
...
+          }
+        };
+
+        var loadBatch = function() {
+          if (currItem < files.length) {
+            while (svc.queue.length < svc.queueLimit && currItem < files.length) {
+              enqueue(files[currItem++]);
+            }
 
+            setTimeout(loadBatch, 500);
```
alex-deaconu: @fjvallarino I would much rather see the angular `$timeout` service used here. Thanks.

test/unit/storage/services/svc-file-upload.tests.js
```diff
@@ -48,21 +48,29 @@ describe("Services: uploader", function() {
...
+
+      expect(uploader.queue.length).to.equal(uploader.queueLimit);
       expect(uploader.queue[0].file.name).to.equal("folder/test1.txt");
-      
+
+      for(var j = uploader.queueLimit - 1; j >= 0; j--) {
+        uploader.removeFromQueue(j);
+      }
+
       setTimeout(function() {
```
alex-deaconu: @fjvallarino rather than using a 500ms timeout here, I would suggest using the `$timeout` service and `$timeout.flush(500);` call instead (for reference: https://github.com/Rise-Vision/rise-vision-apps/blob/master/test/unit/editor/controllers/ctr-widget-modal.tests.js#L85). Thanks.

fjvallarino:
@alex-deaconu updated

web/scripts/storage/services/svc-file-upload.js
```diff
@@ -19,45 +19,43 @@ angular.module('risevision.storage.services')
...
         var enqueue = function (file) {
-          var deferred = $q.defer();
-
-          // Checks it's a file and queue size is not exceeded
-          if ((file.size || file.type) && svc.queue.length < svc.queueLimit) {
+          // Checks it's a file
+          if (file.size || file.type) {
             var fileItem = new FileItem(svc, file, options);
             svc.queue.push(fileItem);
-            svc.onAfterAddingFile(fileItem).then(deferred.resolve);
```
alex-deaconu: @fjvallarino curious, why did you opt to no longer wait for the `UploadURIService.getURI` promise to resolve before queueing the next file? To confirm, you're just batching 10 files every 500ms regardless of whether the previous requests are complete. Are we relying on the browser's max request count mechanism? Thanks.

fjvallarino: @alex-deaconu because I wanted to have parallel uploads (which still don't work, unfortunately). If the browser allows it, we should upload more than one file at once to reduce total upload time. The number 10 was totally arbitrary; I wanted a small number in order to avoid freezing the UI (what was happening before) and to avoid having a really long list of active uploads (which made the UI unpredictable in terms of vertical size).

alex-deaconu: @fjvallarino got it. Actually, looking at the old code I think the issue was recursion in `enqueue` was adding a new function call for every file to the stack. So, I think just removing the promise would have fixed the issue; I guess timeout and batches would ensure there's enough time to process everything.

To confirm, you should remove the promise response from `onAfterAddingFile` since it's no longer used (https://github.com/Rise-Vision/rise-vision-apps/blob/master/web/scripts/storage/directives/dtv-upload.js#L56). 

Thanks.

fjvallarino: @alex-deaconu yes, the idea of the timeouts was to avoid queuing all the files at once. The promise was actually used only once, for the first file; the rest was queued in a for loop, which caused the lag (I assume because of all the XHR objects created at once).

I'll remove the promise you mention.

fjvallarino: @alex-deaconu updated

alex-deaconu:
@fjvallarino thanks for all the fixes. Just tested the stage version with apps/web/bower_components and, well the browser didn't crash (stage-4 upload crashed with the same folder). 

I did however encounter a few file errors: a few GCS 400 errors and a few "No User" errors from Storage server. However, the UI did not show an error, so we may have an issue there. Not related to this PR I think, but just wanted to bring it up.

Looks good 👍 . Thanks!

fjvallarino:
@alex-deaconu did the Retry button show up? The same happened to me (no explicit error was displayed), but retrying worked fine.

alex-deaconu:
@fjvallarino yes, retry did show up. Pressed retry and it started uploading a ton of files again (I think more than a few files failed). Had to close the window because the "Cancel All" button did not work properly. To confirm, I think Cancel All cancels the current files in queue, not the whole operation; we may have an issue there as files keep being added to the queue even after Cancel All is pressed. Thoughts? Thanks.

alex-deaconu:
This is probably not relevant to this PR as well, but trying to delete the folder resulted in a "Task size too large" error.
screen shot 2017-05-12 at 3 23 31 pm

fjvallarino:
@alex-deaconu I'll address the cancel button issue in a new commit (it's caused by this change set). The Task size issue is not related to this PR. Should we create a separate card?

alex-deaconu:
@fjvallarino added a separate card for fixing the Delete job. Thanks.

fjvallarino:
@alex-deaconu updated to remove all items when clicking Cancel All

alex-deaconu:
@fjvallarino fix looks good. I just added an additional unit test, @ezequielc wondering if you can review the last commit. Thanks.

ezequielc:
@alex-deaconu @fjvallarino 👍

PR #436, opened 2017-06-21
Remove beta label from Add Display email [stage-0]

alex-deaconu:
Looks like a PR was needed for this after all. common-i18n strings are already updated via new tag.


fjvallarino:
:+1:

PR #442, opened 2017-07-17
CH Update: fix script that automatically hides extra zendesk widget fields

settinghead:
- Rename common-header to rise-vision-common-header for easier bower link ops

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -16,7 +16,7 @@
     "angular-ui-codemirror": "~0.3.0",
     "angular-vertilize": "~1.0.1",
     "angular-mocks": "1.5.0",
-    "common-header": "https://github.com/Rise-Vision/common-header.git#v1.13.14",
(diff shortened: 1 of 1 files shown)


alex-deaconu:
@settinghead as the repo name is `common-header`, I don't think we should change the folder name here. If you want to fix bower link you can do so in the common-header repo.

Also, the build will most likely not pass at this point. There's another PR being reviewed which should fix the failing e2e tests.

Thanks.

settinghead:
@alex-deaconu ok makes sense re. changing CH name. I'll update the name in CH when an opportunity arises later. Thanks for the heads up

alex-deaconu:
@settinghead FYI Apps build should be fixed now. Thanks!

settinghead:
@alex-deaconu great! Thanks for the update.

PR #443, opened 2017-07-17
Test improved e2e clicks, city challenge

alex-deaconu:
Use recommended env variable
Run unit and e2e tests separately

Refactor loginPage & remove loader checks from other pages

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -16,7 +16,8 @@ dependencies:
 test:
   override:
     - echo $CHROME_INSTANCES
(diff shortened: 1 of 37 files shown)


alex-deaconu:
To confirm, this branch is used to test the common-e2e branch (see other PR). Tests pass pretty well in CCI. Also tested with the onboarding branch and that seems to pass as well now (with an additional improvement).

I've split the unit & e2e tests for better visibility. 

Also, added an env parameter that is supposed to help with e2e tests hanging (https://qualityshepherd.com/2016/12/chrome-hangs-during-protractor-test-runs-on-ubuntu/)

Finally, I've refactored the tests to check loader in the common login function.

fjvallarino:
:+1:

PR #445, opened 2017-07-18
Presentation selector merge into Apps

alex-deaconu:
[stage-0]

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -18,9 +18,9 @@
     "angular-mocks": "1.5.0",
     "common-header": "https://github.com/Rise-Vision/common-header.git#v1.13.15",
     "component-storage-selector": "https://github.com/Rise-Vision/component-storage-selector.git",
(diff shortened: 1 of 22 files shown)


alex-deaconu:
I did this work as part of the Onboarding implementation. However, the final PR took a different route for retrieving the Presentations list and this change was not needed anymore.

However, it is still a piece of refactoring that we may consider keeping. Basically, the Presentation Selector was used in the Schedules App, so it was made a Component. However, as the Apps are merged now, it doesn't need to be separated from the main code.

To confirm, the code is mostly copy/paste with a few tweaks. Main one is that the PresentationListController is re-used in the partial, rather than having its own custom controller. Also updated the i18n strings. 

Associated PRs:
https://github.com/Rise-Vision/common-app-components/pull/74
https://github.com/Rise-Vision/common-i18n/pull/228

fjvallarino:
@alex-deaconu looks good! :+1:

PR #449, opened 2017-07-27
Add Templates Library trial

fjvallarino:


ezequielc:
@fjvallarino Code looks good. WE only need to make sure build works before deploying to master. Thanks!

ezequielc:
@fjvallarino 👍

PR #457, opened 2017-08-04
Check for "cros" string but avoid confusing with "Microsoft"

fjvallarino:


ezequielc:
@fjvallarino 👍

PR #458, opened 2017-08-04
Use previous Common Header version (without Zendesk updates)

fjvallarino:


ezequielc:
@fjvallarino 👍

PR #463, opened 2017-08-15
Forward users to Store from Pricing Page 

ezequielc:
Forward users to Store from Pricing Page using show_product query parameter.

https://apps-stage-1.risevision.com/signup?show_product=123

It should forward to Store if you are signed in.
It should remove the param if you are not signed in, so you are not redirected after signing in/up.

### DIFF
--- a/test/unit/app.tests.js
+++ b/test/unit/app.tests.js
@@ -101,4 +101,73 @@ describe('app:', function() {
 
     spy.should.have.been.called;
   });
+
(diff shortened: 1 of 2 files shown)


ezequielc:
@Rise-Vision/apps You can see a better diff from here: https://github.com/Rise-Vision/rise-vision-apps/pull/463/files?w=1 . Thanks

alex-deaconu:
@ezequielc looks good 👍

PR #473, opened 2017-08-30
Show old Rise players as unsupported

fjvallarino:
.circle/config.yml will be reverted before merging

@alex-deaconu please review

### DIFF
--- a/web/partials/displays/display-fields.html
+++ b/web/partials/displays/display-fields.html
@@ -121,15 +121,20 @@
           </tr>
           <tr class="table-body__row" ng-class="{'alert-warning':factory.isOutdatedPlayer(display)}">
             <td class="table-body__cell"><span translate>displays-app.fields.player.playerVersion</span></td>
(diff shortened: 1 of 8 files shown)


web/partials/displays/display-fields.html
```diff
@@ -121,13 +121,21 @@
           </tr>
           <tr class="table-body__row" ng-class="{'alert-warning':factory.isOutdatedPlayer(display)}">
```
alex-deaconu: @fjvallarino @RV-Adrian this is the line I was concerned about. Should we use an 'alert-warning' or 'alert-danger' class for Unsupported players? Thanks.

fjvallarino: @alex-deaconu @RV-Adrian I have ready that change in my machine (restoring the 'alert-danger' for Unsupported players). Let me know and I'll push it!

alex-deaconu:
@fjvallarino looks good to me 👍 . Don't forget to update the common-i18n tag and revert circle.yml before merging. Thanks

PR #476, opened 2017-09-07
Hide onboarding message after logout

fjvallarino:
@alex-deaconu @ezequielc this is deployed on stage-1. Not sure exactly if it is worth testing or not (it should be an e2e test)

### DIFF
--- a/test/unit/launcher/directives/dtv-onboarding-steps.tests.js
+++ b/test/unit/launcher/directives/dtv-onboarding-steps.tests.js
@@ -413,4 +413,14 @@ describe('directive: onboarding steps', function() {
     });
   });
 
+  describe('on logout: ', function() {
+    it('should reset showOnboarding', function() {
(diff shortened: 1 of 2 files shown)


alex-deaconu:
@fjvallarino I think this change needs to go in the CH. The Onboarding bar is used in Store and other Apps so the issue would appear there as well. Thanks.

fjvallarino:
@alex-deaconu the problem is the component itself is created on Apps' index.html

alex-deaconu:
@fjvallarino oops, you're totally right. Sorry about that. Re: test, I think a unit test can be added for this as discussed in Standup. Thanks.

fjvallarino:
@Rise-Vision/apps updated with unit test

alex-deaconu:
@fjvallarino looks good 👍 , thanks for fixing!

PR #483, opened 2017-10-13
Handle sign up

fjvallarino:
This PR has the logic to avoid calling endpoints/loading data if the user is not already registered. It's complementary with the route handling in common-app-components. The majority of the commits were used to deploy new versions while updating dependencies (not squash because of rebase issues)

@ezequielc please review


ezequielc:
@fjvallarino 👍

PR #484, opened 2017-10-13
Show confirmation messages

fjvallarino:
All this PR does is restoring the referenced branches to the correct values

@ezequielc please review


ezequielc:
@fjvallarino 👍

PR #486, opened 2017-10-18
Rework registration route

alex-deaconu:
Updated CH and components versions
Remove workarounds for registration

[stage-3]


alex-deaconu:

alex-deaconu:
@ezequielc @fjvallarino upon testing, there's an existing issue with the Storage selector and Unregistered users. To confirm, when an Unregistered user is logged in, the Storage selector fails to load properly. This is because the controller required for registration (and all the related registration code) is located in the CH which storage selector doesn't load.

However, this is an existing issue present in production. To confirm, the chance that this behaviour would happen for a user is almost non-existent, so my suggestion is we leave as is.

Thanks.

ezequielc:
@alex-deaconu Agreed, we can leave it as is for now. Thanks

fjvallarino:
@alex-deaconu looks good (after dependency branches updates) :+1:

PR #490, opened 2017-10-25
Reorder tests to prevent company creation conflicts

alex-deaconu:
Rename confusing test

[stage-3]


alex-deaconu:
Each time a test creates a sub-company, it also deletes it once it completes. If two test suites running in parallel create sub-companies, one may finish earlier and delete the others' sub company. This would cause failures.

Seeing a pattern where most failures are for tests that use sub-companies. This PR is trying to fix that.

ezequielc:
@alex-deaconu 👍 Good catch!

PR #491, opened 2017-10-25
Rename main test cases as per file namea

alex-deaconu:
- Replace double quotes
- Build/beautify fixes
- Lint fixes

Just a cleanup branch that I worked on while waiting for e2e tests to pass.

Main issue is some screenshots fail to be created due to the filename being too long. Also, easier to identify what file the test that is failing belongs to. And consistency.


alex-deaconu:
@fjvallarino the module name is added one level above (https://github.com/Rise-Vision/rise-vision-apps/blob/master/test/e2e/storage/storage.cases.js#L15). Ideally I wish I had added something like `Storage: Download:`, but this is ok as well. Thanks.

PR #504, opened 2017-11-13
Fix/product grid

ghislaineguerin:


ghislaineguerin:
@ezequielc tests need to be fixed for this.

alex-deaconu:
@ghislaineguerin FYI updated branch to use the CH Pr I just opened. Also, branch is now up to date with Master. Thanks.

alex-deaconu:
Removed optimizely js file.

Couple of notes: 
- the thumbnail images are cut off. They were cut off before but now a little differently: screen shot 2017-11-15 at 4 43 23 pm
- in the initial prototype only the Image and the Add buttons are clickable. I had to make the whole area clickable because it's required by e2e tests
- because the CH tag is already created (and a new one is live for the Ubuntu 16.04 changes), I cannot update the i18n string for the "Add" button without creating a whole mess. We can update after that other PR is merged

ezequielc:
👍

PR #519, opened 2017-12-07
Add Display Control creation/update

fjvallarino:
This PR adds loading and editing of the configuration file for Display Control

### DIFF
--- a/test/unit/displays/controllers/ctr-display-control-modal.tests.js
+++ b/test/unit/displays/controllers/ctr-display-control-modal.tests.js
@@ -9,7 +9,7 @@ describe('controller: display control modal', function() {
     $provide.service('displayControlFactory',function() {
       return {
         getConfiguration: function() {
-          return Q.resolve('contents');
+          return Q.resolve('loaded contents');
         },
         updateConfiguration: function() {
           return Q.resolve();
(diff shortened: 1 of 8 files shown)


test/unit/displays/controllers/ctr-display-control-modal.tests.js
```diff
@@ -0,0 +1,127 @@
...
+
+  it('should exist', function() {
+    expect($scope).to.be.truely;
+    expect($scope.saveConfiguration).to.be.a('function');
+    expect($scope.resetForm).to.be.a('function');
+    expect($scope.dismiss).to.be.a('function');
+    expect($scope.formData).to.be.ok;
+  });
+  
+  it('should initialize', function() {
```
fjvallarino: @alex-deaconu this is the location where I have the issue. This is one of the cases I want to test, but I need the opposite one (failed to load remote file, show default config file). Since the controller has already been created and the initialization code ran, it's too late to mock a method in the tests (the initialization already ran on the beforeEach block). I guess I can just keep a reference to $controller and create a new instance for this specific test, but probably there are better solutions.

alex-deaconu: @fjvallarino interesting. Actually in this case you can probably just return an unresolved promise for getConfiguration. Then, once the test is running you can resolve/reject the promise as needed to test the different scenarios. Would that work? Thanks.

web/scripts/displays/services/svc-display-control-factory.js
```diff
@@ -42,8 +38,24 @@ angular.module('risevision.displays.services')
...
+               '----------------------------------------------------------------\n' +
+               'interface=\n' +
+               'serial-port=\n' +
+               'serial-baud-rate=\n' +
+               'serial-data-bits=\n' +
+               'serial-parity=\n' +
+               'serial-stop-bits=\n' +
+               'serial-flow-control=\n' +
+               'serial-screen-on-cmd=\n' +
+               'serial-screen-off-cmd=';
```
alex-deaconu: @fjvallarino why not just make this a constant? Thanks.

alex-deaconu:
@fjvallarino code looks good. See my 2 comments in line. Thanks.

web/scripts/displays/services/svc-display-control-factory.js
fjvallarino: @alex-deaconu mainly a "just in case" situation, if we needed to input display id in the sample file, etc. No strong opinion about it.

fjvallarino:
@alex-deaconu updated unit test

web/scripts/displays/services/svc-player-pro-factory.js
```diff
@@ -81,9 +81,9 @@ angular.module('risevision.displays.services')
       factory.openConfigureDisplayControl = function (display) {
         var deferred = $q.resolve();
 
-        if (!display.offlineSubscription) {
-          deferred = factory.openPlayerProInfoModal();
-        }
+        //if (!display.offlineSubscription) {
```
fjvallarino: @alex-deaconu should I uncomment this lines now or should we do it before merging to master?

alex-deaconu: @fjvallarino I'm good with leaving as is, so we can test easier. Thanks.

web/scripts/displays/services/svc-display-control-factory.js
alex-deaconu: @fjvallarino ok, thanks.

alex-deaconu:
@fjvallarino looks good, thanks!

PR #522, opened 2017-12-08
Pass Display when opening modal from list

alex-deaconu:
[stage-4]

This is to ensure the Modal shows correct information for the specific Display in the List. In manage mode we just use the Factory value.


fjvallarino:
👍

PR #535, opened 2017-12-19
Fix subscription status issue

fjvallarino:
The intermediate commits have a lot of code that was later removed, so they are not really useful for review.

This is on stage-3


alex-deaconu:
@fjvallarino looks good 👍 , thanks!

PR #545, opened 2018-01-15
Show Plans modal on related Store links

fjvallarino:
This is ready for review on stage-1.


web/storage-selector.html
```diff
@@ -39,6 +39,9 @@
   <script src="bower_components/angular-local-storage/dist/angular-local-storage.js"></script>
   <script src="bower_components/ng-biscuit/dist/ng-biscuit.js"></script>
 
+  <script src="bower_components/common-header/dist/js/common-header.min.js"></script>
+  <script src="bower_components/common-header/dist/js/dependencies.min.js"></script>
```
alex-deaconu: @fjvallarino I thought the CH build files were no longer needed in Storage Selector after the refactor. Thanks.

web/scripts/editor/directives/dtv-subscription-status.js
```diff
@@ -8,6 +8,8 @@
           restrict: 'E',
           scope: {
             item: '=',
+            useCustomOnClick: '@',
```
alex-deaconu: @fjvallarino is this variable needed. In the CH PR you only have `customOnClick` and verify the presence of that function. Do we need the `useCustomOnClick` flag as well? Thanks.

web/scripts/editor/controllers/ctr-playlist-item-modal.js
```diff
@@ -2,23 +2,28 @@
...
 
       if (item.type === 'presentation') {
         $scope.widgetName = 'editor-app.playlistItem.presentation.name';
       } else if (!item.objectReference && item.settingsUrl) {
         $scope.widgetName = item.name;
       } else {
         if (item.objectReference && item.type === 'widget') {
+          if (item.gadget && plansProductCodes.indexOf(item.gadget.productCode) >= 0) {
+            $scope.isPlansProductCode = true;
+          }
```
alex-deaconu: @fjvallarino wondering what the purpose of this check is? ie. in what instance would the productCode be a plan product code? Thanks.

web/index.html
```diff
@@ -69,6 +69,7 @@
   <script src="bower_components/common-header/dist/js/components/timeline.js"></script>
   <script src="bower_components/common-header/dist/js/components/distribution-selector.js"></script>
   <script src="bower_components/common-header/dist/js/components/background-image-setting.js"></script>
+  <script src="bower_components/common-header/dist/js/components/plans.js"></script>
```
alex-deaconu: @fjvallarino isn't this component part of the common-header.js file? ie. I don't think we need the additional file reference. Thanks.

alex-deaconu:
@fjvallarino reviewed, few questions/comments. Thanks.

web/scripts/editor/directives/dtv-subscription-status.js
fjvallarino: @alex-deaconu I added it because the modal is not shown in all the cases (it's only shown for Embedded Presentations). I didn't need to do this for the subscription-component in Common Header, since all the places it's used can declare the custom-on-click handler unconditionally. I could not find a way of optionally setting the custom-on-click property, so I could check for a null value. If you have any idea of how it could be done, I'd really like to remove the double attribute.

web/scripts/editor/controllers/ctr-playlist-item-modal.js
fjvallarino: @alex-deaconu it's needed because the Plans modal should only be displayed for Embedded Presentations. Any other widget needs to show the standard Store link.

alex-deaconu: @fjvallarino sounds good, thanks!

alex-deaconu: @fjvallarino maybe you can just set the $scope.showPlansModal here. Leave it null otherwise. This will resolve the double variable issue in dtv-subscription-status. Thanks.

fjvallarino: @alex-deaconu the problem is that the custom-on-click attribute is already set to "showPlansModal()" in the partial

alex-deaconu: @fjvallarino I'm not sure which partial you're referring to. Is it playlist-item-modal.html? In that case, setting the custom-on-click to null would indicate that a customOnClick link is not available, you can check for the functions' existence. Maybe I'm missing something. Thanks.

fjvallarino: @alex-deaconu this partial: https://github.com/Rise-Vision/rise-vision-apps/pull/545/files#diff-122bc1305cebcde07e44c0355950cf06R19

How can I set that attribute conditionally? It should only be set when the product code is the one of Embedded Presentations.

PR #561, opened 2018-02-13
Prototype for Image Widget settings [stage-1]

alex-deaconu:


web/scripts/widgets/services/svc-widget-settings-factory.js
```diff
@@ -0,0 +1,94 @@
...
+          }
+        }
+      };
+
+      factory.showSettingsModal = function (item, softUpdate) {
+        if (!item || !item.objectReference && !item.settingsUrl) {
+          return;
+        }
+
+        var modalInstance = $modal.open({
```
ezequielc: @alex-deaconu How do we show the templates for other Widgets? I assume they all use the same controller widgetSettingsModal, but the templateUrl needs to be changed, right? 
Should we have another if statement here? Thanks

alex-deaconu: @ezequielc the templateUrl String will be part of the configuration for the Widget. Thanks.

ezequielc: @alex-deaconu Sounds good. Thanks!

alex-deaconu:
This PR does should actually change anything in Apps. I've disabled the In App UI for the Image Widget, so now it just creates the infrastructure required for the Twitter Widget (separate branch/PR).

The risevision.widgets.* files are imported from https://github.com/Rise-Vision/widget-settings-ui-core & https://github.com/Rise-Vision/widget-image. I did a bit of refactoring, but mostly added unit test coverage.

@ezequielc @fjvallarino @rudibravo please review. @rodrigopavezi FYI. Thanks.

fjvallarino:
@alex-deaconu the code looks good, but I wonder if there's any way I can test it in order to better understand what changed/was added.

alex-deaconu:
@fjvallarino unfortunately the only Widget that was working with this prototype is the Image Widget, but that was disabled. 

I think the most important thing to test at this point is that the existing (iframe) widget and embedded presentations still work correctly (getting and saving the settings). I'm going to try to re-stage the previous commit before disabling the Image Widget in App settings. Will link the URL. Thanks.

alex-deaconu:
@fjvallarino I've restaged the previous commit at https://apps-stage-1.risevision.com. Now the Image Widget settings will open inApp. Thanks.

fjvallarino:
@alex-deaconu looks good

PR #563, opened 2018-02-19
Refactoring store authorization to common-header

rudibravo:
Removed the store authorization to be able to access it's functionality from common header.
It's deployed to stage-3.

@Rise-Vision/apps please review.

Thanks.


alex-deaconu:
@rudibravo looks good. Don't forget to update the branch name back to feature/display-licensing before merge. Thanks.

web/scripts/editor/services/svc-store-authorization.js
```diff
@@ -1,66 +0,0 @@
...
-  ])
-  .factory('checkTemplateAccess', ['storeAuthorization', 'TEMPLATE_LIBRARY_PRODUCT_CODE',
-    function (storeAuthorization, TEMPLATE_LIBRARY_PRODUCT_CODE) {
-      return function (templateCode) {
-        return storeAuthorization.check(TEMPLATE_LIBRARY_PRODUCT_CODE)
-          .catch(function () {
-            return storeAuthorization.check(templateCode);
-          });
-      };
-    }
```
alex-deaconu: @rudibravo actually, I would rather you leave this `checkTemplateAccess` service here. Thanks.

rudibravo: @alex-deaconu I think it makes sense moving that method to svc-template instead of recreating svc-store-authorization on rise-vision-apps. Do you agree? Thanks.

alex-deaconu: @rudibravo I'm not entirely comfortable with that. `svc-template.js` is a different type of service (for direct server access). This is more like a util function. Just move to subscriptionStatusFactory, add a checkTemplateAccess function to it. And I would suggest moving the product code constant there as well (from the controller). Thanks.

rudibravo:
@alex-deaconu addressed your comment. Thanks.

alex-deaconu:
@rudibravo looks good, thanks for fixing 👍

PR #564, opened 2018-02-19
Implement Twitter settings with authentication. Missing tests

rodrigopavezi:


rodrigopavezi:
[Implement Twitter Settings [5]](https://trello.com/c/7ZCdwGVw/4377-implement-twitter-settings-5)

rodrigopavezi:
@ezequielc @fjvallarino @alex-deaconu I would like to get it to beta today if possible as we need to start testing it and unblock the design and test of the Twitter Template. Could you please review it? Cheers

rodrigopavezi:
[Release Twitter to Apps Beta (2)](https://trello.com/c/oIO0hew3/4827-release-twitter-to-apps-beta-2)

alex-deaconu:
@rodrigopavezi just a quick heads up, the `feature/widget-settings` branch is now merged to master. Wondering if you can update this PR to be rebased off master. Will be reviewing shortly. Thanks.

rodrigopavezi:
@alex-deaconu will do. cheers

web/scripts/widgets/common/services/svc-settings-utils.js
```diff
@@ -1,10 +1,12 @@
+/* globals $ */
```
alex-deaconu: @rodrigopavezi I would rather not add a jslint exception. Just leave as is and we can fix this warning in the future by using lodash or something. Thanks.

web/scripts/widgets/common/controllers/ctr-widget-settings-modal.js
```diff
@@ -1,3 +1,4 @@
+/* globals $ */
```
alex-deaconu: @rodrigopavezi same point as below. Let's not add this exception, I'd rather fix (at a later date). Thanks.

web/scripts/config/beta.js
```diff
@@ -29,6 +29,8 @@
     .value('MESSAGING_URL',
       'https://display-messaging.risevision.com')
     .value('APPS_URL', '')
-    .value('ENV_NAME', 'BETA');
+    .value('ENV_NAME', 'BETA')
+    .value('OAUTH_TOKEN_PROVIDER_URL', 'https://services-stage.risevision.com/oauthtokenprovider/')
+    .value('OAUTH_PUBLIC_KEY', 'EJMI-lB9hB55OYEsYmjXDNfRGoY');
```
alex-deaconu: @rodrigopavezi are these values going to be different for various environments? If they're just constants, I'd rather have them in twitter/app.js. Thanks.

web/partials/widgets/twitter-settings-modal.html
```diff
@@ -0,0 +1,90 @@
...
+
+		<form role="form" name="settingsForm">
+		  <div class="form-group">
+		    <!--label class="control-label">Enter a #hashtag or @handle to display Tweets from any public account</label>
+		    <input type="text" class="form-control" placeholder="#hashtag/@handle"-->
+	      <label class="control-label">Enter a @handle to display Tweets from any public account</label>
+		    <input type="text" class="form-control" placeholder="@handle" ng-model="settings.additionalParams.screenName">
+		  </div><!--form-inline-->
+
+		  <!--h4>Options</h4>
```
alex-deaconu: @rodrigopavezi why is this whole block commented? If not being used, please remove. Thanks.

rodrigopavezi: @alex-deaconu we will eventually use it on the next releases. So I don't want to have to go back and check UX source code. It is already there for next releases. cheers

web/scripts/config/beta.js
rodrigopavezi: @alex-deaconu they will change eventually. We still need to create the Oauth Token provider on prod kubernetes. cheers

web/scripts/widgets/common/services/svc-oauth.js
```diff
@@ -0,0 +1,142 @@
+/* globals OAuth */
+'use strict';
+angular.module('risevision.widgets.services')
+  .factory('OAuthService', ['$http', 'OAUTH_TOKEN_PROVIDER_URL', 'OAuthio', '$q', '$log', 'userState',
+    function ($http, OAUTH_TOKEN_PROVIDER_URL, OAuthio, $q, $log, userState) {
+      var svc = {};
+      var provider = '';
+      var authorization = (userState.getAccessToken().token_type === 'Bearer') ? userState.getAccessToken().token_type +
+        ' ' + userState.getAccessToken().access_token : userState.getAccessToken().access_token;
```
alex-deaconu: @rodrigopavezi I don't think this takes into account the Access Token updating. I don't think this should be a constant in this factory. Thanks.

web/index.html
```diff
@@ -262,10 +265,14 @@
   <script src="bower_components/widget-settings-ui-components/dist/js/angular/tooltip.js"></script>
   <!-- <script src="components/widget-settings-ui-components/dist/js/angular/url-field.js"></script> -->
   <script src="bower_components/widget-settings-ui-components/dist/js/angular/file-selector.js"></script>
+  <!-- Twitter Settings -->
```
alex-deaconu: @rodrigopavezi I think this line is out of place. Can you move this to where the twitter settings files are imported? Thanks.

alex-deaconu:
screen shot 2018-02-22 at 10 22 30 am

@rodrigopavezi coveralls.io coverage seems pretty weak for this PR. I don't really look at unit test specifics, but I'm wondering if it could be improved. Thanks.

web/scripts/widgets/twitter/services/svc-twitter-oauth.js
```diff
@@ -0,0 +1,14 @@
+'use strict';
+angular.module('risevision.widgets.twitter')
+  .factory('TwitterOAuthService', ['$q', '$log', 'OAuthService',
+    function ($q, $log, OAuthService) {
+      var svc = {};
+      OAuthService.initialize('twitter');
```
alex-deaconu: @rodrigopavezi wondering how this particular code would work for more than 1 type of provider. Thoughts? Thanks.

web/partials/widgets/twitter-settings-modal.html
```diff
@@ -0,0 +1,90 @@
+<div class="image-settings-fs" ng-controller="TwitterSettingsController" ng-cloak>
+	<div class="modal-header">
+		<button id="dismissButton" type="button" class="close" ng-click="closeSettings()">
+			<i class="fa fa-times"></i>
+		</button>
+		<h3 class="modal-title">Twitter Widget Settings</h3>
```
alex-deaconu: @rodrigopavezi are there plans to have this widget internationalized? Thanks.

web/partials/widgets/twitter-settings-modal.html
```diff
@@ -0,0 +1,90 @@
+<div class="image-settings-fs" ng-controller="TwitterSettingsController" ng-cloak>
```
alex-deaconu: @rodrigopavezi I don't think this class is needed. Thanks.

web/scripts/config/beta.js
alex-deaconu: Sounds good. Thanks.

web/scripts/widgets/common/services/svc-oauth.js
```diff
@@ -0,0 +1,142 @@
...
+              deferred.reject();
+            }
+          }, function (response) {
+            deferred.reject();
+            $log.debug('Could not authenticate with OAuth Token Provider! ' + response);
+          });
+        return deferred.promise;
+      };
+
+      svc.authenticate = function () {
```
alex-deaconu: @rodrigopavezi it seems this function's main purpose is to Generate a Token, correct? Maybe it should be named accordingly. Thanks.

web/scripts/widgets/common/services/svc-oauth.js
```diff
@@ -0,0 +1,142 @@
...
+          .then(function (response) {
+            deferred.resolve(response);
+          }, function (response) {
+            deferred.reject(response);
+            $log.debug('Could not get Status! ' + response);
+          });
+        return deferred.promise;
+      };
+
+      svc.getConnectionStatus = function () {
```
alex-deaconu: @rodrigopavezi it seems this functions' main purpose is to validate a token. Maybe it should be named accordingly. Thanks.

web/scripts/widgets/common/services/svc-oauth.js
rodrigopavezi: @alex-deaconu good point. I will change it to a function. cheers

web/index.html
rodrigopavezi: @alex-deaconu good catch. will adjust that. cheers

web/scripts/widgets/common/services/svc-oauth.js
```diff
@@ -0,0 +1,142 @@
...
+                    key = newKey;
+                    deferred.resolve(newKey);
+                  }, function () {
+                    deferred.reject();
+                  });
+              }, function () {
+                deferred.reject();
+              });
+          }, function () {
+            deferred.reject();
```
alex-deaconu: @rodrigopavezi wondering if you could chain these promises rather than this. You can do a .then(null, function(){}) or .catch() block at the end to catch errors. Thanks.

web/scripts/widgets/twitter/services/svc-twitter-oauth.js
rodrigopavezi: @alex-deaconu I envision which provider creating its own OAuthService that calls the initialize with its own provider name. cheers

web/partials/widgets/twitter-settings-modal.html
rodrigopavezi: @alex-deaconu yes, missed that. cheers

web/partials/widgets/twitter-settings-modal.html
rodrigopavezi: @alex-deaconu good catch, I will take this out. cheers

web/scripts/widgets/common/services/svc-oauth.js
rodrigopavezi: @alex-deaconu it is more than just generate a token. It will ask a user to authenticate with Twitter and then save the token. I think the name makes sense based on the Oauth.io documentation. cheers

web/scripts/widgets/common/services/svc-oauth.js
rodrigopavezi: @alex-deaconu we call in the settings a provider connection so we need to check connection status so here it gets the connection status and the controller will check that and set a scope variable with the right status. I think the name is correctly. cheers

web/scripts/widgets/common/services/svc-oauthio.js
```diff
@@ -0,0 +1,28 @@
+'use strict';
+angular.module('risevision.widgets.services')
+  .factory('OAuthio', ['OAUTH_PUBLIC_KEY', '$window', '$q', '$log',
+    function (OAUTH_PUBLIC_KEY, $window, $q, $log) {
+      var svc = {};
+
+      svc.initialize = function () {
+        $window.OAuth.initialize(OAUTH_PUBLIC_KEY);
```
alex-deaconu: @rodrigopavezi couple of things.
This init method should probably just be called when this service is loaded (rather than a public function). Is that possible?
You should probably check for $window.OAuth existence before calling the function.
Thanks.

web/scripts/widgets/common/services/svc-oauth.js
```diff
@@ -0,0 +1,142 @@
...
+          'authorization': authorization
+        },
+        'withCredentials': true,
+        'responseType': 'json'
+      };
+      var key = '';
+
+      svc.initialize = function (newProvider) {
+        OAuthio.initialize();
+        provider = newProvider;
```
alex-deaconu: @rodrigopavezi as per my suggestion below, if we no longer need to initialize OAuthio then we can probably just call this setProvider. Thanks.

web/scripts/widgets/twitter/services/svc-twitter-oauth.js
alex-deaconu: @rodrigopavezi okay, but you're calling a static object here. A second provider will just overwrite the provider and this factory would use the updated provider name. Anyways, I guess we'll need to fix with the next provider implementation.

rodrigopavezi:
@alex-deaconu back to you for reviewing latest commits. I will rerun the build for getting E2E passing. Cheers

alex-deaconu:
@rodrigopavezi awesome improvement in coverage, looks good to merge 👍 . 

Don't forget to add the i18n strings when ready.

Thanks!

PR #569, opened 2018-02-23
Update Display Details page to show RPP toggle

fjvallarino:
@Rise-Vision/apps @rudibravo please review


rudibravo:
@fjvallarino looks good 👍

fjvallarino:
@Rise-Vision/apps @rudibravo I added the Display List updates in the same PR as Display Details, since some updates were required on both sides.

I will address the coverage reduction in a separate PR.

web/scripts/displays/controllers/ctr-display-details.js
```diff
@@ -159,57 +156,6 @@ angular.module('risevision.displays.controllers')
         }
       };
 
-      var refreshSubscriptionStatusListener = $rootScope.$on('refreshSubscriptionStatus', function () {
-        $loading.start('loading-trial');
-      });
-
-      var subscriptionStatusListener = $rootScope.$on('subscription-status:changed',
```
rudibravo: @fjvallarino you removed this listener for subscription status but I don't see a listener for the plans changed. right now if I start a plan trial I still can't toggle the RPP.

fjvallarino: @rudibravo this subscription status listener was for individual RPP, not for Plans. Are you emitting an event when starting a Plan trial? I could listen for that one. There are some complications, though, since you could have switched the current company, and the display can only be marked as Pro on the company that created it.

rudibravo: @fjvallarino when the user starts a trial we updating the companystate, which emits a "risevision.company.updated". But maybe we want to listen to "risevision.plan.loaded" since it also called when the user changes company.

fjvallarino: @rudibravo updated and ready on stage-1. I went the "risevision.company.updated" path; the other path leads to using information of a (potentially) different company, which is something we decided not to do (the discussion starts here: https://trello.com/c/eV5HN3vS/3060-apps-show-rpp-toggle-in-the-display-page-2#comment-5a8ee8b54aa2a6130857d514)

rudibravo: @fjvallarino 👍

rudibravo:
@fjvallarino looks good 👍

PR #571, opened 2018-03-01
Remove product trials

fjvallarino:
@rudibravo please review


rudibravo:
@fjvallarino looks good 👍

PR #582, opened 2018-03-12
Improve unit tests

fjvallarino:


web/scripts/displays/controllers/ctr-display-details.js
```diff
@@ -26,7 +26,7 @@ angular.module('risevision.displays.controllers')
       displayFactory.getDisplay(displayId).then(function () {
         $scope.display = displayFactory.display;
         $scope.monitoringEmailsString = ($scope.display.monitoringEmails || []).join(EMAIL_DELIMITER);
-        $scope.monitoringSchedule = _parseTimeline($scope.display.monitoringSchedule);
+        $scope.monitoringSchedule = $scope.parseTimeline($scope.display.monitoringSchedule);
```
alex-deaconu: @fjvallarino curious, why not just pass the monitoringSchedule String to the directive, and let it handle the parsing and updating of the variable? ie. move all this logic there. Thanks.

fjvallarino: @alex-deaconu mainly because I wanted the timeline-basic component to be as close as possible to the existing timeline one, and also because I didn't want to tie it to the Datastore representation (which was not set in stone at that point). No preference here, to be honest.

alex-deaconu: @fjvallarino okay, in that case why not create a `monitoring` directive similar to the `screenshot`? I think there's plenty of functionality & isolation to warrant that. Thanks.

fjvallarino: @alex-deaconu updated. I tried creating a directive wrapping timeline-basic-textbox (you can see the commits below), but it was a dead end. Since I had to watch for changes on a couple fields, some scenario always ended up not being reflected properly. To fix that, I added an independent attribute to use for synchronization... at which point, besides being an ugly solution, both parent and embedded directives became too coupled to make sense as separate components. For this reason, I ended up moving all the parsing/formatting logic into the timeline-basic-textbox directive (the one in Common Header).

fjvallarino:
@Rise-Vision/apps please review. This is deployed on stage-1

alex-deaconu:
@fjvallarino 👍

PR #583, opened 2018-03-13
Add support to edit emails as tag objects

fjvallarino:

### DIFF
--- a/web/partials/displays/display-fields.html
+++ b/web/partials/displays/display-fields.html
@@ -191,7 +191,7 @@
             <td class="table-body__cell" style="white-space: nowrap; vertical-align: top; border-top: 0;">
               <span class="">Email Recipients</span>
             </td>
-            <td class="table-body__cell" ng-class="{ 'has-error': !areEmailsValid() }" style="padding: 0">
+            <td class="table-body__cell" style="border-top: 0;" ng-class="{ 'has-error': !areEmailsValid() }">
               <tags-input ng-model="monitoringEmailsList"
                           ng-disabled="!display.monitoringEnabled"
                           type="email"
--- a/web/partials/displays/display-fields.html
+++ b/web/partials/displays/display-fields.html
@@ -191,11 +191,12 @@
             <td class="table-body__cell" style="white-space: nowrap; vertical-align: top; border-top: 0;">
               <span class="">Email Recipients</span>
             </td>
-            <td class="table-body__cell" style="border-top: 0;" ng-class="{ 'has-error': !areEmailsValid() }">
-              <tags-input ng-model="monitoringEmailsList"
-                          ng-disabled="!display.monitoringEnabled"
+            <td class="table-body__cell" style="border-top: 0;" ng-class="{ 'has-error': !displayDetails.monitoringEmails.$valid }">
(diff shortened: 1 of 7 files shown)


web/partials/displays/display-fields.html
```diff
@@ -190,10 +190,23 @@
...
+                          ng-model="monitoringEmailsList"
+                          ng-disabled="!display.monitoringEnabled"
+                          add-on-enter="true"
+                          add-on-space="true"
+                          add-on-comma="true"
+                          add-on-blur="true"
+                          add-on-paste="true"
+                          on-tag-adding="isValidEmail($tag)">
+              </tags-input>
+              <p class="help-block validation-error-message-email" ng-show="!displayDetails.monitoringEmails.$valid">
```
alex-deaconu: @fjvallarino I don't think it's too important, but when I type in an invalid email, the text turns red, however it doesn't show this error. If too hard to fix with the tags-input component I good with leaving out the feature. Thanks.

web/scripts/displays/controllers/ctr-display-details.js
```diff
@@ -98,15 +98,8 @@ angular.module('risevision.displays.controllers')
                !playerProFactory.isUnsupportedPlayer($scope.display);
       };
 
-      $scope.areEmailsValid = function () {
-        var emails = $scope.monitoringEmailsString.split(EMAIL_DELIMITER);
```
alex-deaconu: @fjvallarino I believe the 'EMAIL_DELIMITER' variable is no longer used and can probably be removed. Thanks.

web/scripts/displays/controllers/ctr-display-details.js
```diff
@@ -98,15 +98,8 @@ angular.module('risevision.displays.controllers')
...
-        var allValid = true;
-
-        for(var i = 0; i < emails.length; i++) {
-          allValid = allValid && (!emails[i] || EMAIL_REGEX.test(emails[i]));
-        }
-
-        return allValid;
+      $scope.isValidEmail = function (email) {
+        return email && email.text && EMAIL_REGEX.test(email.text);
```
alex-deaconu: @fjvallarino wondering if you could add some simple unit tests validating this function. Thanks.

web/partials/displays/display-fields.html
fjvallarino: @alex-deaconu yeah, that's the issue @RV-Adrian mentioned. I spent a couple of hours on it, but could not find a solution that worked properly (for instance, after removing the invalid text, the message was still shown because the tags-input does not update its $valid status until it loses focus or a tag related action is executed.

RV-Adrian: @fjvallarino The hover issue should be fixed now. I'm good with the validation appearing when off-focus in this case (as the user has to move to 'Save' in any scenario)
Also, the SCSS component addition looks good 👍 
Thanks!

alex-deaconu:
@fjvallarino looks good 👍 . Thanks!

web/partials/displays/display-fields.html
fjvallarino: @RV-Adrian thanks for the help!

PR #589, opened 2018-03-15
Fix onboarding issue on /unsubscribe path

fjvallarino:


alex-deaconu:
@fjvallarino shouldn't this go directly to master? Thanks.

fjvallarino:
@alex-deaconu the /unsubscribe route is deployed to master, but nobody uses it (users will only have access to it when the Monitoring emails are deployed)

alex-deaconu:
@fjvallarino sounds good. PR looks good 👍 . Thanks!

PR #590, opened 2018-03-19
Twitter as a Professional Widget

alex-deaconu:


web/partials/editor/store-products-modal.html
```diff
@@ -17,7 +17,67 @@ <h3 id="storeModalTitle" class="modal-title">{{ 'editor-app.storeProduct.' + (fa
 
     <search-filter filter-config="filterConfig" search="search" do-search="factory.doSearch"></search-filter>
 
-    <div class="template-filters u_margin-sm-bottom">
+    <div ng-hide="factory.search.category === 'Templates'">
+      <h4 class="u_margin-md-top">{{ 'editor-app.storeProduct.professional.title' | translate }} - <a href="#" translate="editor-app.storeProduct.professional.learnMore"></a></h4>
```
alex-deaconu: @ezequielc @ghislaineguerin what URL is the help link supposed to point to? Thanks.

alex-deaconu:
Still working to fix 1 Twitter Settings related test.

@Rise-Vision/apps please review. Thanks.

web/scripts/editor/services/svc-widget-utils.js
```diff
@@ -94,8 +94,28 @@ angular.module('risevision.editor.services')
...
+      url: 'https://widgets.risevision.com/stable/components/rise-twitter/rise-twitter-widget.html'
+    },
+    {
+      env: 'PROD',
+      name: 'Twitter Widget',
+      imageUrl: 'https://s3.amazonaws.com/Rise-Images/UI/twitter-widget%402x-100.jpg',
+      imageAlt: 'add twitter widget',
+      gadgetType: 'Widget',
+      id: '67e511ae-62b5-4a44-9551-077f63596079',
+      url: 'https://widgets.risevision.com/stable/components/rise-twitter/rise-twitter-widget.html'
```
alex-deaconu: @sheadarlison @rodrigopavezi FYI I've updated the URL for the Twitter Widget.

Also, to confirm, I've named the widgets slightly differently (test widget is called 'Twitter Widget Test') to differentiate between which version is being used.

Finally, to confirm, while this does no longer require a product code, it will require an `id` field. This `id` does not necessarily have to match with an entity in the datastore, but will act as an unique identifier for the Widget.

Thanks.

sheadarlison: Thanks @alex-deaconu 

@rodrigopavezi should the test point to beta instead of stable? 

Thanks

test/e2e/editor/cases/twitter-settings.js
```diff
@@ -0,0 +1,224 @@
...
+    before('Add Presentation & Placeholder: ', function () {
+      presentationsListPage.openNewPresentation();
+
+      helper.clickWhenClickable(workspacePage.getAddPlaceholderButton(), 'Add Placeholder button');
+
+      browser.sleep(500);
+
+    });
+
+    describe('Should only Twitter Widget when on a Plan Trial: ', function() {
```
fjvallarino: @alex-deaconu is a word missing here?

test/e2e/editor/cases/twitter-settings.js
```diff
@@ -0,0 +1,224 @@
...
+      it('should display the current screen name', function() {
+        placeholderPlaylistPage.getItemNameCells().get(0).click();
+        
+        helper.wait(twitterSettingsPage.getTwitterSettingsModal(), 'Twitter Settings Modal');
+
+        expect(twitterSettingsPage.getTwitterScreenName().getAttribute('value')).to.eventually.equal('risevision');
+      });
+    });
+
+    xdescribe('should authenticate user: ', function() {
```
fjvallarino: @alex-deaconu is this one intentionally excluded?

web/scripts/editor/services/svc-widget-utils.js
rodrigopavezi: @sheadarlison good point. @alex-deaconu can you change Twitter Test to point to the beta folder. 

It should be https://widgets.risevision.com/beta/components/rise-twitter/rise-twitter-widget.html 

@alex-deaconu Thanks for thanking the work. 

Cheers

test/e2e/editor/cases/twitter-settings.js
alex-deaconu: @fjvallarino yes, sorry forgot to mention that. This test was failing for me the whole day yesterday, and as per my conversation with Rodrigo, it seems Twitter has a similar mechanism to Google, they require further confirmation of the account after multiple retries: https://s3.amazonaws.com/uploads.hipchat.com/106443/793075/cDFiXNOvDgYPgTg/editor_twitter_settings_should_authenticate_user___should_log_in_user_to_twitter.png. I'm disabling this test and Creative is adding a card to have a more permanent solution for it. Thanks.

test/e2e/editor/cases/twitter-settings.js
alex-deaconu: @fjvallarino you're right, updated. Thanks.

ezequielc:
@alex-deaconu It is a minor issue, but the Twitter Widget image is not clickable. You should be able to click on it to add it to the presentation like the other widgets. Thanks

alex-deaconu:
@ezequielc I've updated and made the whole area clickable, for the Twitter Widget, in both states. This is consistent with the other Store Products' behaviour. 

However, I do agree that it would be better if only the Image was clickable instead of the whole area. I can make it behave in that manner, including for other products to make it consistent. I haven't done it already because I'm worried it may break e2e tests. Let me know if you think I should.

To confirm, looking at the partial, the behaviour is also inconsistent for the Blank template.

Thanks.

ezequielc:
@alex-deaconu I am fine with what you currently have. We can improve later if needed, considering the effort. Thanks

PR #592, opened 2018-03-20
Fix/trial expired modal

RV-Adrian:


web/partials/displays/display-fields.html
```diff
@@ -167,31 +167,31 @@
     <div class="panel panel-default">
       <table class="table table--pro-features" ng-class="{ 'table--disabled': !display.playerProAuthorized }">
         <tbody class="table-body">
-          <tr class="table-body__row" ng-show="!isPlanActive() && !display.playerProAuthorized" style="text-align: center; padding: 22px;">
+          <tr class="table-body__row text-center u_padding-md" ng-show="!isPlanActive() && !display.playerProAuthorized">
```
RV-Adrian: @fjvallarino 
Inline styles removed and a few other cleanups in HTML partials. Thanks!

fjvallarino: @RV-Adrian I need to restore the ones with "translate" (I added those as part of translation changes). Were they causing you issues?

RV-Adrian: @fjvallarino 
Looks like I made the updates before pulling in your translate changes. Should be good to go now!

PR #601, opened 2018-04-06
Fix logic for missing product

alex-deaconu:
Instead of removing product,
Assume product is missing,
And add it when subscribed

Ensure search query is applied

Fix replace issue in Add User email (https://github.com/Rise-Vision/common-header/pull/793)

[stage-3]

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -15,7 +15,7 @@
     "Sortable": "^1.4.2",
     "angular-truncate": "*",
     "angular-slugify": "^1.0.1",
-    "common-header": "https://github.com/Rise-Vision/common-header.git#fix/replace",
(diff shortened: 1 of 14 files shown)


alex-deaconu:

alex-deaconu:
@ezequielc the object currently contains the full API response for that Product. To be honest, I didn't know which fields are needed and being used, so it was more or less a trial and error process. Will remove the extra lines as it doesn't seem those are being used. Thanks.

ezequielc:
@alex-deaconu Sounds good. Thanks!

alex-deaconu:
@ezequielc updated and also pushed some apparently overdue Formatting and JSHint fixes. Thanks.

ezequielc:
@alex-deaconu 👍

PR #605, opened 2018-04-13
Show Pro Toggle for all Player types, to be able to turn it off (disabled in other cases)

fjvallarino:
This is deployed on stage-1. You can test the failing scenario we had using the following displays (please, don't turn Pro off on the first one):

- Third Party with Pro license: https://apps-stage-1.risevision.com/displays/details/9KGARZAJ7JJQ?cid=44ac909d-ade3-436f-80b7-a122d7b23ead
- Third Party without Pro license: https://apps-stage-1.risevision.com/displays/details/6VZJ3RDUGUJK?cid=44ac909d-ade3-436f-80b7-a122d7b23ead
 
I did not add a text message mainly because I was not sure where to add it. If you have a suggestion about it, let me know and I'll add it.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -15,7 +15,7 @@
     "Sortable": "^1.4.2",
     "angular-truncate": "*",
     "angular-slugify": "^1.0.1",
-    "common-header": "https://github.com/Rise-Vision/common-header.git#v2.3.2",
+    "common-header": "https://github.com/Rise-Vision/common-header.git#fix/show-toggle-3rd-party",
     "oauthio-web": "0.6.2",
     "ng-tags-input": "^3.2.0"
   },
--- a/bower.json
+++ b/bower.json
@@ -15,7 +15,7 @@
     "Sortable": "^1.4.2",
     "angular-truncate": "*",
     "angular-slugify": "^1.0.1",
-    "common-header": "https://github.com/Rise-Vision/common-header.git#fix/show-toggle-3rd-party",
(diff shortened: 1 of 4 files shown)


alex-deaconu:
@fjvallarino PR looks good 👍 .

The place where the warning could be added is here: https://github.com/Rise-Vision/rise-vision-apps/pull/605/files#diff-15e79bbd8e228a6fddc2bfa898ac9919L43, next to the "All licenses used" warning.

Thanks.

fjvallarino:
@alex-deaconu I updated the PR. Currently, it's only showing the message if the player is 3rd Party/Unsupported and RPP is Off (it's not shown when it's On). Thinking about it a bit more, I guess it makes more sense to always show the message.

alex-deaconu:
@fjvallarino yes, I think that warning should be shown in both cases.

@ezequielc wondering if you could review the behaviour and wording.

Thanks.

ezequielc:
@fjvallarino @alex-deaconu Agree on showing the message on both cases.
Looks good other than that. Thanks!

fjvallarino:
@alex-deaconu @ezequielc updated and ready on stage-1 (the same links on the PR's note can be used)

ezequielc:
@alex-deaconu That is a good point. The problem is that all other RPP features will not work, like Professional widgets and that might be confusing for users. But anyway, I think it is a good thought to share with Delivery. Thanks

Issue #603, opened 2018-04-12
Downloading a file from Storage with commas in the name makes it have no file extension so it cannot be played on your computer


robb-price:
<b>How it works now</b>
Try to download both files below. Notice that one can be played with a media player on your machine, the other cannot. The one without the file extension cannot be played. The file extension is not included since the file name has a comma.

<a href="https://storage.googleapis.com/risemedialibrary-556323b8-984c-4127-aa08-4d73d019bcb8/Commas/Ask%20Doreen-%20March%2028%2C%202018.mp4"> File with comma</a>

<a href="https://storage.googleapis.com/risemedialibrary-556323b8-984c-4127-aa08-4d73d019bcb8/Commas/LTA%20March%2021%202018.mp4">File without a comma</a>

<b>How it should work:</b>
Regardless of commas in the file name, the file extension should be included when you download the file so you can play it.

robb-price:
![](https://github.trello.services/images/mini-trello-icon.png) [(3) Issue 603: Commas cause files in Storage to not have file extensions](https://trello.com/c/1MfGnkJz/7807-3-issue-603-commas-cause-files-in-storage-to-not-have-file-extensions)

ByronDarlison:
@ezequielc should this defect be here or recorded somewhere else? Thanks

ezequielc:
@ByronDarlison @robb-price I assume by 'download' it means the download button in Apps/Storage. If so, it makes sense to be here. 

But if the issue also affects other places, like Player/Viewer, then it should be logged and fixed in https://github.com/Rise-Vision/storage-server. 

Thanks!

ezequielc:
@ByronDarlison I talked with @robb-price and confirmed the issue doesn't affect Displays. In this case it is not urgent, but we will have it fixed by around Tuesday. I will keep track of it in SoS. Thanks!

alex-deaconu:
Verified fix with the file provided above. Also, tested various other scenarios in this Storage account, including downloading from a sub-folder and downloading sub-folder contents (https://apps-stage-3.risevision.com/storage?cid=6bc18847-6537-4473-9a52-defaefbeb1c0).

@robb-price this issue is now resolved. Closing. Thanks!

PR #606, opened 2018-04-17
Using character encoding for the filename

alex-deaconu:
(https://tools.ietf.org/html/rfc8187#section-3.2.3)

[stage-3]


alex-deaconu:
Fix for #603 

olegrise:
@alex-deaconu interesting solution. Looks good 👍

PR #608, opened 2018-04-17
Remove blank language parameter

alex-deaconu:
It's not functioning correctly

[stage-3]


alex-deaconu:
It seems the previous fix did not fix all instances of the issue. The extra quotes (which are supposed to be defining a language code did not work correctly. Removed and it seems the use cases work better.

You can try downloading the files at this staging link to ensure the filenames are correct:
https://apps-stage-3.risevision.com/storage?cid=6bc18847-6537-4473-9a52-defaefbeb1c0

To confirm, for some reason the `"` character is being transformed into a `_`, not sure why.

PR #609, opened 2018-04-18
Correcting header according to google documentation

alex-deaconu:
response-content-disposition header has its own syntax
https://cloud.google.com/storage/docs/xml-api/reference-headers#responsecontentdisposition

[stage-3]

### DIFF
--- a/web/scripts/storage/services/svc-download-factory.js
+++ b/web/scripts/storage/services/svc-download-factory.js
@@ -21,9 +21,8 @@ angular.module('risevision.storage.services')
         iframe.id = hiddenIFrameID;
         iframe.style.display = 'none';
         iframeContainer.appendChild(iframe);
-        // Using character encoding for the filename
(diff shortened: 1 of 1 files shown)


alex-deaconu:
Seems the previous fix messed up downloads from a folder. The folder name is being appended to the file name. Reviewed some more and it seems the parameter does not follow the specification, but has its own way of defining the filename parameter: https://cloud.google.com/storage/docs/xml-api/reference-headers#responsecontentdisposition.

You can test downloading various files here, also try downloading from the folder:
https://apps-stage-3.risevision.com/storage?cid=6bc18847-6537-4473-9a52-defaefbeb1c0

As before, `"` is being replaced with a `_`.

alex-deaconu:
@fjvallarino good point. Just double checked and it seems the file names are correct when being downloaded via the folder option. Thanks.
screen shot 2018-04-18 at 11 32 53 am

PR #614, opened 2018-04-23
Feature/ux chrome os

RV-Adrian:


fjvallarino:

web/partials/displays/display-add-modal.html
```diff
@@ -30,41 +28,199 @@ <h3 class="modal-title" translate="displays-app.download.player" ng-show="downlo
...
+      </ul>
+
+      <select id="select-rise-player-os-mobile" ng-model="currentTab" class="form-control visible-xs">
+        <option translate="displays-app.add.instructions.os.windows10" value="windows"></option>
+        <option translate="displays-app.add.instructions.os.chromeos" value="chromeos"></option>
+        <option translate="displays-app.add.instructions.os.linux" value="linux"></option>
+        <option translate="displays-app.add.instructions.os.pi" value="pi"></option>
+      </select>
+
+      <div ng-show="currentTab == 'windows'" class="prototype-download-content row u_margin-md">
```
alex-deaconu: @fjvallarino since these tabs are like pages for this modal, I'm wondering if it would be easier if we separated them into partials (one for each tab). Thoughts? Thanks.

web/partials/displays/add-display-email.html
```diff
@@ -0,0 +1,149 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd">
```
alex-deaconu: @fjvallarino I don't believe this template is needed anymore. Thanks.

web/partials/displays/display-fields.html
```diff
@@ -245,20 +234,22 @@
...
             </td>
           </tr>
+          
         </tbody>
       </table>
     </div>
 
     <!-- SCREENSHOT  -->
-    <screenshot></screenshot>
+    <div ng-show="!playerProFactory.isCROS(display)">
```
alex-deaconu: @fjvallarino wondering why use `ng-show` here and not `ng-if`. I think `ng-if` doesn't render the elements so it may save some resources. Just a thought. Applies to the other 2 instances for this condition above. Thanks.

web/partials/displays/display-add-modal.html
```diff
@@ -30,41 +28,199 @@ <h3 class="modal-title" translate="displays-app.download.player" ng-show="downlo
...
+        <div class="col-xs-6 text-center">
+          <a href="http://install-versions.risevision.com/installer-win-32.exe#_ga=2.127839630.200841446.1523456923-448654264.1459728419" class="u_link-no-underline">
```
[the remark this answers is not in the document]
fjvallarino: @alex-deaconu I left that because I assumed it's some sort of tracking feature. @RV-Adrian is it ok if I remove that part of the link?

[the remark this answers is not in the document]
RV-Adrian: 👍

web/images/chrome-b.svg
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="utf-8"?>
```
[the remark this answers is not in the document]
RV-Adrian: @alex-deaconu @fjvallarino 
That's a good point, I didn't think of this. I'm not sure if the benefit is only if the icon is re-used (we're only using it in one place right now)
@ghislaineguerin I believe you set this up? Any thoughts? Thanks!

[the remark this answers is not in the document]
fjvallarino: @alex-deaconu I'll move them into that library

web/partials/displays/display-add-modal.html
fjvallarino: @alex-deaconu I'm not sure it's worth it, considering they are relatively small and they are only being used as part of this modal.

alex-deaconu: @fjvallarino that's fine. Thanks.

alex-deaconu:
@fjvallarino PR looks good 👍 . Thanks!

Issue #612, opened 2018-04-20
Inconsistent function with item schedule


tylermwight:
If you choose to select displays instead of "show on all displays" but you do not select any displays it will continue to show that item. If you select 1 display (that isn't being used) it will go away as expected. But if you then remove that display to get back to "no displays selected" that item will still not show up. So you can save the schedule in the "same state" from what we see on screen and have different functionalities.

This came from here, I do agree that setting a schedule that way to hide the item but not delete it would be nice as the default behavior.

Steps to replicate

1. Have two items in a place holder
2. On the second item click on it's schedule and uncheck "select all displays". Not not select any displays and save
3. It still shows the second item even though no displays are chosen
4. Go in and select a display that this presentation isn't playing on and it will hide it
5. Edit the item's schedule again and now de-select all displays so we have no displays selected, just like we did in step 2
6. Now it is hidden even though we saved in the 'same state' as step 2

Video here: https://goo.gl/UxNwVg

alex-deaconu:
@tylermwight this issue should now be resolved in Apps and RVA. To confirm, existing Presentations` Distribution will not be affected, however the Items should consistently hide if the Presentation is Saved/Published. And of course, newly created Presentations should behave correctly as well.

Please verify and re-open if you still see this happening.

Thanks.

PR #617, opened 2018-04-25
Set empty distribution array when no Display is selected

alex-deaconu:

[stage-3]


alex-deaconu:
Fix for https://github.com/Rise-Vision/rise-vision-apps/issues/612

Issue #538, opened 2017-12-22
Firefox Quantum - Can't submit support request or move schedule playlist items


kaden-rise:
**Reproduction Steps:**

1. Support Request

Select "Get Support" from the "Need Help?" menu at the top of the Apps or Store page. With a non-priority support account, you'll then select "Contact Support" from the next window.

2. Playlist items

In the schedules menu try to change the order of the playlist items
![image](https://user-images.githubusercontent.com/33877348/34306946-0b1970a6-e703-11e7-85fb-a1659d0311e7.png)

**Expected Results:**

1. Support Request

A little window should pop-up in the bottom right that lets you ask a question, but it doesn't

2. Moving playlist items

You should be able to drag the playlist item to change the order, but nothing happens

alex-deaconu:
The issue with moving Playlist Items should now be resolved. Leaving this Issue open however for the Zendesk widget not opening correctly. Thanks.

ezequielc:
@kaden-rise @alex-deaconu  I opened a separate issue for the Support Widget fix (https://github.com/Rise-Vision/rise-vision-apps/issues/629). I am closing this as Playlist Item issue is fixed. Thanks

PR #624, opened 2018-04-27
Add fallback flag to fix Firefox

alex-deaconu:
Fix for #538 

Seems there is a Firefox bug with the drag handler not being called correctly. Setting this flag should fix the issue however I believe it's no longer using HTML5 native functionality anymore if the flag is set. However, it seems to work as before either way.

https://github.com/RubaXa/Sortable/issues/370

[stage-3]


alex-deaconu:
@fjvallarino I don't believe we use the groups function. It's used:
'To drag elements from one list into another, both lists must have the same group value...'
Thanks.

PR #633, opened 2018-05-04
Fix file upload logic

alex-deaconu:
0 byte file
Fail upload if Range header parsing fails

[stage-3]


alex-deaconu:
Fix for https://github.com/Rise-Vision/storage-server/issues/274

alex-deaconu:
@Rise-Vision/apps please review. 

@fjvallarino had to copy the e2e test fix from your branch due to failures, nice work!

Thanks.

PR #636, opened 2018-05-07
Restart upload if Range header is missing

alex-deaconu:
As per documentation, if the Range header is missing
No bytes were sent with the upload

Restarting upload in this case


alex-deaconu:
Fix suggested by @tejohnso as per documentation:

> If you received a 308 Resume Incomplete response, process the response's Range header, which specifies which bytes the server has received so far. The response will not have a Range header if no bytes have been received yet.

Staging on Beta as of now and informing users of this option. If all goes well we will release tomorrow.

@Rise-Vision/apps please review. Thanks.

alex-deaconu:
As discussed at Standup, I've improved the code to reset the progress bar if restarting the download.

Also, I've integrated @RV-Adrian's improvement the `activeUploadCount` as well as the "Retry" button for individual upload failures and error styling for upload failures.

See last 4 commits.

Interesting thing is, when trying to test this on Blake's machine where we could replicate the issue reliably yesterday, it seems the issue no longer occurs. We're assuming something was fixed on Google's side.

ezequielc:
@alex-deaconu I tried to simulate an upload failure by blocking the URL in Dev Console. I can see the "Retry failed uploads" button, but not the individual Retry and styling changes.
Is there any other way I can replicate this? And, I think the error state should also cover this network failure.
Thanks!

ezequielc:
@alex-deaconu Actually, I tested on stage-0 but I am wondering if it got replaced by the last CH tag deployment... Thanks

alex-deaconu:
@ezequielc actually this is staged on the "beta" branch (https://apps-beta.risevision.com). Thanks.

ezequielc:
@alex-deaconu 👍

RV-Adrian:
@alex-deaconu 👍 
There's no note of 'Your files are uploaded one at a time' but that's kinda obvious right?

I'll remove my pull requests and reference branch around 'activeUploadCount'

alex-deaconu:
@ezequielc @RV-Adrian after some discussion and testing it turns out the `activeUploadCount` does not show the total number of files remaining, but just those that are added to the queue (current queue size is 10). I've updated the PR to no longer show a file count, it will just say "Uploading Files".

I've also prototyped a fix for this via a `remainingFileCount` variable, however I will add a card for correcting this properly.

Thanks.

PR #644, opened 2018-05-14
Remove Need Support button

fjvallarino:
Should only require a Common Header update, but Zendesk load on startup breaking unit tests requires replicating the fix applied on Common Header. Still thinking about a better solution.


fjvallarino:
@alex-deaconu routes removed

olegrise:
@fjvallarino looks good. I will approver after CH version is updated.

fjvallarino:
@olegrise updated

PR #645, opened 2018-05-17
Improve error messages for Presentations

fjvallarino:
This is the main idea for improving the user facing messages, after looking at the different exceptions Core can generate while working with Presentations. Originally I planned on modifying Core, but I think we can avoid doing that, mainly because we can't only rely on error codes and need to analyze error messages (for instance, Conflict -409- is used for a bunch of situations). The same logic used here can be applied to svc-display-factory.js and svc-schedule-factory.js. 

Let me know if you agree with this, or you'd rather do it differently. One obvious improvement would be using translation strings, of course.

### DIFF
--- a/web/scripts/editor/services/svc-editor-factory.js
+++ b/web/scripts/editor/services/svc-editor-factory.js
@@ -11,12 +11,12 @@ angular.module('risevision.editor.services')
     'presentationTracker', 'store', 'VIEWER_URL', 'REVISION_STATUS_REVISED',
     'REVISION_STATUS_PUBLISHED', 'DEFAULT_LAYOUT', 'TEMPLATES_TYPE',
     '$modal', '$rootScope', '$window', 'scheduleFactory', 'messageBox',
-    '$templateCache',
+    '$templateCache', '$log',
     function ($q, $state, userState, userAuthFactory, presentation,
       presentationParser, distributionParser, presentationTracker, store,
       VIEWER_URL, REVISION_STATUS_REVISED, REVISION_STATUS_PUBLISHED,
       DEFAULT_LAYOUT, TEMPLATES_TYPE, $modal, $rootScope, $window,
-      scheduleFactory, messageBox, $templateCache) {
+      scheduleFactory, messageBox, $templateCache, $log) {
       var factory = {};
       var JSON_PARSE_ERROR = 'JSON parse error';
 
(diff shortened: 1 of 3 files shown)


fjvallarino:

web/scripts/editor/services/svc-editor-factory.js
```diff
@@ -507,12 +507,63 @@ angular.module('risevision.editor.services')
...
       var _showErrorMessage = function (action, e) {
         factory.errorMessage = 'Failed to ' + action + ' Presentation.';
-        factory.apiError = e.result && e.result.error.message ?
-          e.result.error.message : e.toString();
+        factory.apiError = _processErrorCode('Presentation', action, e);
 
         messageBox(factory.errorMessage, factory.apiError);
       };
 
+      var _processErrorCode = function (itemName, action, e) {
```
alex-deaconu: @fjvallarino hmm, I would have expected a separate factory for this, with more comprehensive unit tests. I will review this as is, however we should consider refactoring next time we touch this. Thanks.

fjvallarino: @alex-deaconu that's the idea I mentioned in the PR's header. I'd rather do it in a cleaner way too. The objetive was checking if the mechanism was ok, and the extend it for the other two entities. Maybe I can extract as a separate service/factory as part of this PR, with better unit tests, and then add it to the other entities in a new one.

alex-deaconu: @fjvallarino yes, I recall the discussion. We can address as part of the next card. Thanks.

web/scripts/editor/services/svc-editor-factory.js
```diff
@@ -507,12 +507,63 @@ angular.module('risevision.editor.services')
...
+          }
+          else if (errorString.indexOf('is required') >= 0) {
+            return messagePrefix + ' ' + errorString;
+          }
+          else {
+            return messagePrefix + ' ' + tryAgainMessage;
+          }
+        }
+        else if(e.status === 401) {
+          return 'You do not have permissions to ' + actionName + ' this ' + itemName + ', because you are not authenticated. Please, sign in.';
```
alex-deaconu: @fjvallarino should this not be `actionName.action`? Thanks.

web/scripts/editor/services/svc-editor-factory.js
```diff
@@ -507,12 +507,63 @@ angular.module('risevision.editor.services')
...
+        factory.apiError = _processErrorCode('Presentation', action, e);
 
         messageBox(factory.errorMessage, factory.apiError);
       };
 
+      var _processErrorCode = function (itemName, action, e) {
+        var error = e.result && e.result.error;
+        var errorString = error.message ? error.message : error.message.toString();
+        var actionName = { Get: 'loaded', Add: 'added', Update: 'updated', Delete: 'deleted', Publish: 'published', Restore: 'restored' }[action];
+        var messagePrefix = 'The ' + itemName + ' could not be ' + actionName + '.';
```
alex-deaconu: @fjvallarino same as below, should this not be `actionName.action`? Thanks.

web/scripts/editor/services/svc-editor-factory.js
fjvallarino: @alex-deaconu no, ```actionName``` is just a string. I'm using ```action``` to index into the map (I did not want to use a bunch of if statements, but probably it would have been clearer, or at least have a separate variable for the map and index into it in a separate line)

alex-deaconu: Yes, I see that now: screen shot 2018-05-18 at 4 52 01 pm

I think also gulp build would re-format that object.

In this case I think our extended line length is also a problem as the Github UI does not resize in width past the size of that screenshot.

web/scripts/editor/services/svc-editor-factory.js
```diff
@@ -507,12 +507,63 @@ angular.module('risevision.editor.services')
...
-        factory.apiError = e.result && e.result.error.message ?
-          e.result.error.message : e.toString();
+        factory.apiError = _processErrorCode('Presentation', action, e);
 
         messageBox(factory.errorMessage, factory.apiError);
       };
 
+      var _processErrorCode = function (itemName, action, e) {
+        var error = e.result && e.result.error;
+        var errorString = error.message ? error.message : error.message.toString();
```
alex-deaconu: @fjvallarino as discussed, I think `error.message.toString();` gives a very cryptic `[Object Object]` message when trying to parse an object.

Maybe `JSON.stringify(error.message)` would work however that would be even more cryptic.
alternatively we can just show a generic message:
`var errorString = error.message ? error.message : "An Error Has Occurred";`.

Thoughts? Thanks.

alex-deaconu:
@fjvallarino one suggestion in line for an unrelated fix. Otherwise this PR looks good, with the noted suggestions for improvement in subsequent PRs. Thanks!

ezequielc:
@fjvallarino 👍

PR #650, opened 2018-05-22
Allow saving presentations created from Templates without requiring changes

fjvallarino:
I tried to find a cleaner way of checking this, but it would have been cumbersome and I'm not sure it would have been a real improvement.

Redeployed on Beta because Templates can't be tested on staging.

### DIFF
--- a/web/scripts/editor/controllers/ctr-workspace.js
+++ b/web/scripts/editor/controllers/ctr-workspace.js
@@ -20,7 +20,7 @@ angular.module('risevision.editor.controllers')
       $scope.placeholderFactory = placeholderFactory;
       $scope.isSubcompanySelected = userState.isSubcompanySelected;
       $scope.isTestCompanySelected = userState.isTestCompanySelected;
-      $scope.hasUnsavedChanges = false;
+      $scope.hasUnsavedChanges = editorFactory.presentation.name.indexOf('Copy of ') === 0;
 
       var _isEqualIgnoringFields = function (o1, o2) {
         if (typeof o1 === 'object') {
--- a/web/scripts/editor/controllers/ctr-workspace.js
+++ b/web/scripts/editor/controllers/ctr-workspace.js
@@ -20,7 +20,7 @@ angular.module('risevision.editor.controllers')
       $scope.placeholderFactory = placeholderFactory;
       $scope.isSubcompanySelected = userState.isSubcompanySelected;
       $scope.isTestCompanySelected = userState.isTestCompanySelected;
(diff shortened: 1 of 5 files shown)


web/scripts/editor/controllers/ctr-workspace.js
```diff
@@ -20,7 +20,7 @@ angular.module('risevision.editor.controllers')
       $scope.placeholderFactory = placeholderFactory;
       $scope.isSubcompanySelected = userState.isSubcompanySelected;
       $scope.isTestCompanySelected = userState.isTestCompanySelected;
-      $scope.hasUnsavedChanges = false;
+      $scope.hasUnsavedChanges = editorFactory.presentation.name.indexOf('Copy of ') === 0;
```
alex-deaconu: @fjvallarino to confirm, this fix also works for copying a presentation, correct? Thanks.

fjvallarino: @alex-deaconu it does. It was not required, but it definitely works. I think it makes sense, actually.

alex-deaconu:
@fjvallarino this does fix the issue at hand, however, if I open any existing Presentation who's name matches that string, it would always let me save it regardless of changes. Just want to make sure that we're ok with that. @ezequielc thoughts? Thanks.

ezequielc:
@alex-deaconu I think that is a good point. There is no strong reason to stop users from Saving at any point. I agree with the suggestion.
But to confirm, that code is also used to show the unsaved changes popup, so that remains, right? Thanks

alex-deaconu:
@ezequielc agreed, I would not change the unsaved changes part of the functionality. Thanks.

fjvallarino:
@alex-deaconu @ezequielc the thing is that everything is tied to $scope variable called ```hasUnsavedChanges```. It's the thing deciding if Save button is enabled or not, and whether to show the confirmation dialog when navigating away from the editor. If we allow saving any presentation (without making further changes to our code), users will always receive the popup when navigating away.

alex-deaconu:
@fjvallarino I think all we need to do is remove the `ng-disabled="!hasUnsavedChanges"` here: https://github.com/Rise-Vision/rise-vision-apps/blob/master/web/partials/editor/footerbar.html#L19,
and remove the extra condition you added in this PR.

Thoughts? Thanks.

fjvallarino:
@alex-deaconu @ezequielc this is ready again (after a few failed builds). I'm wondering if the same logic should also be applied to the Preview button.

alex-deaconu:
@fjvallarino looks good 👍 . I would leave the Preview button as is (Save & Preview if Save is needed, and only Preview if no changes are made). Thanks.

PR #653, opened 2018-05-29
Only use fallback when browser is Firefox

fjvallarino:
This version only uses fallback when the browser is Firefox, to avoid ghosting issues. Since this causes Firefox to present the ghosting problem, a style was added here as a workaround: https://github.com/Rise-Vision/common-header/pull/817

Regarding the browser detection part, I kept it as simple as possible (maybe too simple, actually); I'm not sure if we have a common module we can use for this kind of functions. The obvious alternative is using something like https://www.npmjs.com/package/detect-browser, but I'm not sure we actually need it.

This is deployed on stage-1.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -15,7 +15,7 @@
     "Sortable": "^1.4.2",
     "angular-truncate": "*",
     "angular-slugify": "^1.0.1",
-    "common-header": "https://github.com/Rise-Vision/common-header.git#fix/ghost-dragging",
+    "common-header": "https://github.com/Rise-Vision/common-header.git#v2.4.7",
     "oauthio-web": "0.6.2",
     "ng-tags-input": "^3.2.0"
   },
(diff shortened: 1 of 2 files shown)


alex-deaconu:
@fjvallarino the simple Firefox check works for me. If we need it elsewhere we can add a utility function. Looks good 👍 . Thanks!

fjvallarino:
@alex-deaconu @RV-Adrian fix for Firefox should work fine on both Presentations and Schedules now

RV-Adrian:
@fjvallarino Looks good! The only thing I noted is on Firefox in mobile view, I was not able to move any playlist items (not a ghosting issue, just didn't move at all?) If we decided not to support that then 👍

fjvallarino:
@RV-Adrian mmm, I don't think we support Firefox mobile. I'm merging for the moment

PR #654, opened 2018-05-30
Process errors for Schedules, Displays and Lists

alex-deaconu:
Show Storage List loading/error status

[stage-3]


alex-deaconu:
Initial error handling PR. Main feature is it shows more user friendly messages to users when API calls fail. Adding more `$log.error` logging for these cases to allow easier debugging.

Storage Files List currently has no error message for failures. Added a basic failure message which I will expand in a future PR.

PR #660, opened 2018-06-06
Detect network errors and Storage refactor

alex-deaconu:
Detect and inform user of gapi.client load errors

Use Storage v2 api
API throws Core-like error objects
Refactor error messages and error catching for Storage

Remove startTrial for Storage

[stage-3]

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -15,7 +15,7 @@
     "Sortable": "^1.4.2",
     "angular-truncate": "*",
     "angular-slugify": "^1.0.1",
-    "common-header": "https://github.com/Rise-Vision/common-header.git#v2.4.9",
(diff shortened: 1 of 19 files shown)


alex-deaconu:

alex-deaconu:
Easiest way to test is to use Chrome's "Request Blocking" function available in the console's menu (
screen shot 2018-06-06 at 5 31 27 pm). Personally I tested with both Storage URLs (https://storage-dot-rvacore-test.appspot.com*) and Core (I used Schedules https://rvacore-test.appspot.com/_ah/api/core/v1/schedule*) to ensure errors are shown correctly across the board. You can enable and disable blocking before or after loading the page to test the GAPI load errors.

However, to note we don't show errors if any other gapi functions fail to load (Core or Store) as it was not in the scope of this fix.

Thanks!

Issue #668, opened 2018-06-13
Trial subscriptions show strange status.


btfrise:
**How to recreate**
1. Create a Presentation in a Company, add a trial of a 3rd party Widget to that Presentation.

**What happens**
The Widget shows _editor-app.subscription.status.premium - 7 Days Trial_.

**What should happen**
The Widget trial status should show _7 Days Trial
![trialstatus](https://user-images.githubusercontent.com/8458926/41362406-e34689ac-6ef6-11e8-9d36-0abf8f1ab616.png)
_

alex-deaconu:
@btfrise this issue is now fixed. It should say "Not Subscribed" in this case. Thanks.
<img width="819" alt="screen shot 2018-06-13 at 3 35 18 pm" src="https://user-images.githubusercontent.com/6218188/41373833-7770d5e2-6f1f-11e8-8268-6ccacb32d17d.png">

PR #670, opened 2018-06-13
Fix Premium gadget wording [stage-0]

alex-deaconu:
Correction for #668 . Solution discussed on the card.


ezequielc:
@alex-deaconu I couldn't find a Widget to test as we don't have them test env, but code looks good 👍

PR #671, opened 2018-06-13
Fix embedded Presentations product appearing in Templates list

alex-deaconu:
Remove Subscription Status from Templates Library
Updated wording for Windows 7

[stage-0]


alex-deaconu:

Issue #669, opened 2018-06-13
If only one item in Schedule, the duration of that item is blank, and PUD is not selected, Display shows that nothing is assigned to Schedule.


btfrise:
**How to recreate**
1. Add a Presentation to a Schedule that is assigned to a Display.
2. Remove any entry in the Duration field.
3. Make sure PUD is unchecked.

**What happens**
Display shows that nothing is scheduled.

**What should happen**
An error message should appear when trying to save a playlist item with no duration and PUD is not checked.

alex-deaconu:
@btfrise this issue should now be corrected. We've added validation for the Duration field in the Placeholder Item and Schedule Playlist Item settings. It should not allow blank duration or less than 1. Thanks.

PR #675, opened 2018-06-18
Add validation for Duration fields

alex-deaconu:
Fix for #669 


alex-deaconu:
@olegrise interesting. Being in seconds, it would be really hard to define what the upper limit would be, however. Someone may intentionally want to set something to play for a few hours at a time. We can always add that in later if we see an issue. Thanks.

PR #683, opened 2018-06-27
Fix issue with header height in editor

alex-deaconu:
Update via new CH tag

[stage-4]


alex-deaconu:

PR #687, opened 2018-07-04
Show 3rd party widgets as licensed

fjvallarino:
I was able to add a couple of widgets on testing, but I left the rest only for production. One of the widgets does not have an assigned image (StockTrack Rankings), but that's something that's already happening in the current version.

@alex-deaconu please review


fjvallarino:
@alex-deaconu updated

web/scripts/editor/services/svc-products-factory.js
```diff
@@ -15,6 +15,191 @@ angular.module('risevision.editor.services')
     'paymentTerms': 'Subscription',
     'trialPeriod': 14,
     'productCode': 'd3a418f1a3acaed42cf452fefb1eaed198a1c620'
+  }, {
```
alex-deaconu: @fjvallarino maybe we should split these out into their own file as well? Thanks.

alex-deaconu:
@fjvallarino thanks for updating. PR looks good 👍

alex-deaconu:
@fjvallarino looks good 👍 thanks for updating.

Issue #698, opened 2018-07-25
Widgets search is not working


ezequielc:
Regardless of what I search for, all 'For Licensed Displays' appear in the widgets list.

To replicate:
Search for 'youtube' in widgets list.
All 'For Licensed Displays' get listed.

What should happen instead:
Search for 'youtube' in widgets list.
Only widgets with 'youtube' keyword (in title or description) should be listed.

PR #700, opened 2018-07-26
Filter professional widgets based on query

alex-deaconu:
Add no result messages

[stage-0]


alex-deaconu:
Fix for #698 

@olegrise please review. Thanks!

alex-deaconu:
@olegrise as per discussions on the card, I'm now hiding the wording when no results are found to reduce confusion. Please review last commit. Thanks.

Issue #701, opened 2018-08-03
Background Image causes the background to not take up the entire Editor background


robb-price:
### What is the expected behavior?
Using a background image in a Presentation will make it take up the entire background of the Presentation.

### What is the actual behavior? 
When you set a Presentation to use a background image, the bars on the top and left of the Presentation Editor cause it to not extend the entire width and height of the Presentation.

What steps can be taken to replicate the problem?
1. Set a Presentation to have an image background, and select an image from Storage.
2.  Notice that the background does not extend across the entire Presentation.

The current workaround is to just use a Background Placeholder, but this should still be corrected.

This video shows the issue in detail: https://goo.gl/S4btKb

alex-deaconu:
@robb-price wondering if you could provide a link for that Presentation so we can use for testing. Thanks.

robb-price:
@alex-deaconu I sure can: https://apps.risevision.com/editor/workspace/f4b24494-05e5-46ee-a2b4-17ff24a67064/?cid=556323b8-984c-4127-aa08-4d73d019bcb8

alex-deaconu:
@robb-price thanks!!

alex-deaconu:
@robb-price the issue should now be resolved. Thanks.

PR #702, opened 2018-08-03
Fix width/height ratio for the artboard

alex-deaconu:
[stage-0]


alex-deaconu:
Fix for #701 

Turns out the issue was not necessarily the padding, but that we were applying it unevenly. Because the Height was only applying 1*padding, the Width/Height ratios were not matching, causing the background to be distorted.

alex-deaconu:
@fjvallarino thanks!

PR #705, opened 2018-08-07
Check Chargebee account exists and act accordingly

fjvallarino:
Subscriptions information for /billing will be added in a further PR. The same will happen with translation strings for app-billing.html.

@alex-deaconu @olegrise please review


fjvallarino:
@alex-deaconu @olegrise @ezequielc I added a new (small) commit to handle related functionality in the /billing page. I will not be pushing further commits to this branch unless it's required by the review process.

Issue #713, opened 2018-08-16
Player controls in Display Settings throw obscure errors for non-Administrators


PatrickA-RiseVision:
### What is the expected behavior?
Users who are not Display Administrators do not have access to Player controls in Display Settings.

### What is the actual behavior?
Player controls appear in Display Settings for non-Administrators and are clickable, but the app objects with an obscure error when clicked.

![image](https://user-images.githubusercontent.com/40774237/44238182-f2267c00-a189-11e8-9371-ca861ea0d2ac.png)

"Error:[object Object]"

### What steps can be taken to replicate the problem?
1. In a test company with at least 1 display, create a user who does not have the "Display Administrator" permissions.
2. While logged in as the user, go to https://apps.risevision.com/displays.
3. Click on "Restart Rise Player", then click "Yes" when prompted about whether you are sure.

alex-deaconu:
@PatrickA-RiseVision this issue has been corrected. To confirm, 2 fixes were implemented.

First off, errors from the API will no longer show the `[object Object]` style error message you saw before. It should correctly parse and show the error message:
<img width="761" alt="screen shot 2018-08-17 at 2 10 01 pm" src="https://user-images.githubusercontent.com/6218188/44281953-70852b80-a227-11e8-9927-a21dc0191a4f.png">

Second, we are no longer going to show the Restart/Reboot buttons for users without the specific "Display Administrator" role.

Thanks.

PR #714, opened 2018-08-17
Correct error parsing for restart/reboot commands

alex-deaconu:
[stage-0]


alex-deaconu:
Fix for #713 

You can use a request blocker to test out the error response. If you take off the Display Administrator the Restart/Reboot buttons should no longer show up.

PR #726, opened 2018-08-30
Use extend instead of copy for Updating Company Address

alex-deaconu:
[stage-0]


alex-deaconu:

Issue #751, opened 2018-09-27
Display License Status Not Updating Correctly


colby-maughan:
### What is the expected behavior?

Clicking the toggle switch for display should update the display's licensed status instantly. 

### What is the actual behavior?

The page must be reloaded to update the display's license status. If a user clicks more than once, they will eventually receive an error claiming they have used all of their display licenses. Refreshing the page resolves the issue. 

### What steps can be taken to replicate the problem?

1. Toggle the license status on a display. It does not matter if it is a free display or licensed. 
2. The status will appear to reset to back to whatever it was toggled off of. Refreshing shows correct license status.
3. Continue clicking the toggle switch. Eventually, you will receive an error stating you are out of licenses to use. 
4. Refresh the page.
5. Attempt to toggle license, which will result in step 2 repeating itself.

ezequielc:
@colby-maughan We released a fix earlier today. I will go ahead an close the issue. You might need to clear your cache to get the latest version. Thanks!

PR #752, opened 2018-09-27
Update Display Assignment and Authorization

alex-deaconu:
If the call succeeds update variables
Fix listener function to not reverse the values

[stage-0]


alex-deaconu:
Fix for #751 

PR #763, opened 2018-10-17
text update [stage-4]

ghislaineguerin:


ghislaineguerin:
@alex-deaconu Please review this change. Also, I think there's technical debt on this page since the text should be on common header locales folder, right?

alex-deaconu:
@ghislaineguerin looks good. I'm not sure re internationalization strings, we previously decided that it would not be applied to onboarding pages.

Just trying to get the tests to pass for this branch so we can merge.

Thanks!

PR #768, opened 2018-10-18
Fix old Player versions being shown as Supported

fjvallarino:

### DIFF
--- a/web/scripts/displays/services/svc-player-pro-factory.js
+++ b/web/scripts/displays/services/svc-player-pro-factory.js
@@ -7,15 +7,15 @@ angular.module('risevision.displays.services')
   .value('DISPLAY_CONTROL_PLAYER_VERSION', '2018.01.15.16.31')
   .value('CHROMEOS_PLAYER_VERSION', '2018.07.20.10229')
   .value('CHROMEOS_SCREENSHOT_PLAYER_VERSION', '2018.08.17.8388')
-  .factory('playerProFactory', ['$rootScope', '$q', '$modal', 'userState',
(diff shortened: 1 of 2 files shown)


alex-deaconu:
@fjvallarino looks good 👍 

Do you think it's worth adding the regex validation to the Electron Player check as well?

Thanks.

fjvallarino:
@alex-deaconu I don't think we check for dates in ```isElectronPlayer```, so it wouldn't be necessary

alex-deaconu:
@fjvallarino okay, I was referring in general to other date checks. Just a suggesiton. Thanks.

web/scripts/displays/services/svc-player-pro-factory.js
```diff
@@ -89,17 +93,17 @@ angular.module('risevision.displays.services')
       factory.isScreenshotCompatiblePlayer = function (display) {
         return !!(display && factory.isElectronPlayer(display) &&
             display.playerVersion >= SCREENSHOT_PLAYER_VERSION) ||
```
alex-deaconu: @fjvallarino I think you missed this comparison and the one in `isOutdatedPlayer`. Thanks.

fjvallarino: @alex-deaconu no, I actually added the comparison and then realized those objects are already dates. Also thought about moving that logic to simple text comparison, but we need to add an expiration threshold (3 months), so I kept it the way it currently is implemented.

fjvallarino:
@alex-deaconu updated!

PR #769, opened 2018-10-19
fix status styles [stage-4]

ghislaineguerin:


ezequielc:
Cleaning old PRs. Thanks

Issue #734, opened 2018-09-11
Users are not warned when uploading file types we do not support. 


sheadarlison:
### What is the expected behavior?
When a user uploads a file type that is not supported in our players they should be warned as such and provided with instructions on what files are supported. 

- https://help.risevision.com/hc/en-us/articles/201788630-Video-Widget-supported-formats
- https://help.risevision.com/hc/en-us/articles/115001957183-Supported-file-types-for-the-Image-Widget

### What is the actual behavior?
Users are not warned when they upload unsupported file types. 

### Files types that should trigger a warning
Videos

- .mov  
- .wmv
- .m4v
- .flv
- .avi
- .ogg

Images

- .tiff

FYI @alanclayton

andrecardoso:
Fix deployed to production.

PR #774, opened 2018-11-01
Show warning when uploading non supported files

andrecardoso:
Deployed to https://apps-stage-7.risevision.com


andrecardoso:
[(3) [Issue #734] Users are not warned when uploading file types we do not support.](https://trello.com/c/0T2xAtOZ/4699-3-issue-734-users-are-not-warned-when-uploading-file-types-we-do-not-support)

PR #786, opened 2018-12-04
Fix last connection time

tejohnso:
Displays that never connected will return Number(null) which is 0 which
results in AngularJS returning the EPOCH start date.

### DIFF
--- a/web/scripts/displays/services/svc-display-status-factory.js
+++ b/web/scripts/displays/services/svc-display-status-factory.js
@@ -30,7 +30,7 @@ angular.module('risevision.displays.services')
               if (idStatus[id]) {
                 idStatus.lastConnectionTime = Date.now();
               } else {
-                idStatus.lastConnectionTime = lastConnection(id);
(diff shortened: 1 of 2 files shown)


web/scripts/displays/services/svc-display-status-factory.js
```diff
@@ -30,7 +30,7 @@ angular.module('risevision.displays.services')
               if (idStatus[id]) {
                 idStatus.lastConnectionTime = Date.now();
               } else {
-                idStatus.lastConnectionTime = lastConnection(id);
+                idStatus.lastConnectionTime = lastConnection(id) || "-";
```
alex-deaconu: @tejohnso the Apps repo uses `'` (single quotes) for strings; wondering if you could update for consistency and prevent linting errors. Also, wondering if you could provide a unit test for this fix? Thanks.

tejohnso: @alex-deaconu see latest

alex-deaconu:
@tejohnso looks good, thanks for the fixes 👍

PR #788, opened 2018-12-10
Remove preview header and image from Twitter settings

stulees:
- Staged version of twitter settings here - https://apps-stage-6.risevision.com/editor/workspace/98ad407d-21d8-453d-909d-45cee0b955fb/?cid=b428b4e8-c8b9-41d5-8a10-b4193c789443
- @sheadarlison Copy changes also applied via latest common-header

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -15,7 +15,7 @@
     "Sortable": "^1.4.2",
     "angular-truncate": "*",
     "angular-slugify": "^1.0.1",
-    "common-header": "https://github.com/Rise-Vision/common-header.git#v3.2.9",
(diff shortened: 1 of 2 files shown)


stulees:
@santiagonoguez Not sure why that would happen for you. I believe @sheadarlison was able to access okay, as can I. Are you signed into your risevision Google account when you try it?

santiagonoguez:
staged_error

I am, but no worries. If @sheadarlison could check the behavior then that's fine.

alex-deaconu:
@santiagonoguez I would assume your Company/Username does not have access to that Presentation. Thanks.

PR #797, opened 2019-01-11
 Allow user to Subscribe even if Trial is available

alex-deaconu:
[stage-3]


alex-deaconu:

PR #801, opened 2019-01-15
Improve e2e tests. Run smaller, independent, CircleCI Workflow steps

fjvallarino:
This PR improves a few situations:

- Better handle of clicking
- Some explicit delays in a test that was constantly failing and for which it was not possible to wait for some explicit element
- Use different Sub Company name for some conflicting test suites
- Break e2e_tests workflow step into smaller steps, which allow for running them in parallel and retrying them individually

### DIFF
--- a/test/e2e/circle.launcher-schedules.js
+++ b/test/e2e/circle.launcher-schedules.js
@@ -1,0 +1,2 @@
+require('./launcher/launcher.cases.js');
+require('./schedules/schedules.cases.js');
--- a/test/e2e/common/cases/first-signin.js
+++ b/test/e2e/common/cases/first-signin.js
@@ -133,6 +133,10 @@ var FirstSigninScenarios = function() {
         expect(getStartedPage.getGetStartedContainer().isDisplayed()).to.eventually.be.false;
       });
 
+      it('removes current SubCompany',function(){
+        commonHeaderPage.deleteCurrentCompany();
+      });
+
       xdescribe('Onboarding Bar: ', function() {
         it('should show Add Presentation CTA home page', function () {
           helper.wait(homepage.getPresentationCTA(), 'Presentation Call to Action');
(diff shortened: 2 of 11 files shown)


test/e2e/common/cases/first-signin.js
```diff
@@ -117,22 +121,26 @@ var FirstSigninScenarios = function() {
...
 
         helper.waitDisappear(commonHeaderPage.getLoader(), 'CH spinner loader');
 
         expect(getStartedPage.getGetStartedContainer().isDisplayed()).to.eventually.be.false;
       });
 
+      it('removes current SubCompany',function(){
+        commonHeaderPage.deleteCurrentCompany();
+      });
+
```
alex-deaconu: @fjvallarino not an important point, but why was this moved above the disabled test? I think it's confusing to have it there. Thanks.

test/e2e/circle.launcher-schedules.js
```diff
@@ -0,0 +1,2 @@
+require('./launcher/launcher.cases.js');
+require('./schedules/schedules.cases.js');
```
alex-deaconu: @fjvallarino curious, why not have two different suites for Launcher and Schedules? Do these tests depend on each other? Thanks.

fjvallarino: @alex-deaconu yeah, apparently there is. I did not delve too much into what the problem was, but you can see the errors here:

https://circleci.com/workflow-run/fecc87ed-b9b8-42aa-9a08-47d0df6ffa18
https://circleci.com/workflow-run/cebb7cc3-1043-4185-97a0-347d9642b041

I retried a few times and the same errors kept popping up, and they disappeared when ran as a single step. I did not want to spend more time into it.

test/e2e/common/cases/first-signin.js
fjvallarino: @alex-deaconu I moved it up mainly because I had not realized it was there, and assumed someone else could have the same issue. I actually thought about removing the whole disabled suite, but I was not sure why we still have it (didn't we remove the Onboarding Bar?)

Anyway, I'm fine with moving it back down in case it's clearer.

alex-deaconu: @fjvallarino yes, the Onboarding Bar is actually hidden but not removed. I'm good with leaving as is. Thanks.

alex-deaconu:
@fjvallarino PR looks good to me 👍. However, as discussed, you need to fix the Pass/Fail status of the build in Github. 

Thanks!

fjvallarino:
@alex-deaconu done with the updates

Issue #802, opened 2019-01-16
Clicking on a Widget in the Presentation Placeholder Playlist in Editor incorrectly redirects to Home


PatrickA-RiseVision:
### What is the expected behavior?
Clicking on a Widget in the Presentation Placeholder Playlist in Editor opens the Widget settings.
### What is the actual behavior?
Clicking on a Widget in the Presentation Placeholder Playlist in Editor redirects to Home, as though the user has just logged in.  If I try this with a sub-company using my Rise Vision administrator account, I end up in the root of Rise.
### What steps can be taken to replicate the problem?
1. Open an affected presentation (?  It may not be per presentation, but I've not been able to replicate it with other presentations yet)
2. Click on a Placeholder
3. Click on an item in the Playlist
4. You are redirected to Home page

fjvallarino:
@PatrickA-RiseVision @pcsandford I'm closing this issue, since we have not been able to reproduce it in the Apps team. I did see it happen on Patrick's machine, but I could never replicate it outside that environment. This makes me think it's cache related (I know you mentioned clearing the cache fixed the problem temporarily for some users, but then it came back) but, most importantly, not being able to reproduce it makes it really hard to figure out what the cause could be. In case this is reported again, please re open the issue. Thanks!

fjvallarino:
@ezequielc @alex-deaconu FYI

PatrickA-RiseVision:
This has been reported again a few times.  I will link specific examples on the card.

alex-deaconu:
@PatrickA-RiseVision we have released a fix for this issue. Please let us know if you encounter any more tickets related to this. Thanks!

PR #814, opened 2019-02-04
Fix issue with iframe redirects

alex-deaconu:
Redirect happens randomly when the directive compiles late
src is set to '{{ widgetUrl }}' causing a console error
and a redirect to the root domain

Replacing src with about:blank
and setting value via directive seems to fix the issue

[stage-0]


alex-deaconu:
Fix for #802 

I was able to replicate locally. This warning gave me a hint as to what might be going on:
screen shot 2019-02-04 at 1 14 25 pm

It seems the src property is not being updated in time before the iframe renders. It is incorrectly parsed as `{{ widgetUrl }}`, causing an error and a redirect to root. Then, when the src is reset to the correct Widget URL it has no impact.

This fix sets the URL to a valid `about:blank` before setting the `src` property via Javascript.

alex-deaconu:
@andrecardoso that's actually a very good suggestion! Updated and it works as expected. Thanks!

PR #816, opened 2019-02-06
Add support for creating presentation on 'HTML Templates' template type

fjvallarino:
This PR adds the placeholder for what will represent the new Template Editor. This is the rationale for some of the decisions:

- A new service and controller pair were created for this editor. The idea was to avoid having the existing `editorFactory` becoming larger. There is only one situation where logic related to the new editor was kept in that service: when selecting an `HTML Template`, the `editorFactory` calls the corresponding method on the new service if the `productTag` matches.
- In the Presentation List, the `ui-sref`s were replaced by `ng-click`, since depending on the Presentation Type the editor we need to open is different. I chose this method because it avoided duplication (there are four links in that list), but it has the disadvantage of disabling `Ctrl-Click` to open the Presentation in a new tab. Let me know if you'd rather handle this differently.
- A new route (`/templates`) was added, although there is no specific menu entry for the Editor.
- We were not showing a spinner when opening a Presentation. The current solution uses the existing Presentation List spinner to indicate loading, although using the Global Spinner also makes sense.
- For testing, you can use https://apps-stage-1.risevision.com
  - Adding a new Presentation can be tested with the product/template named _HTML Template Product 1_. This Product only exists on Staging. This only shows the Presentation placeholder, but currently we don't have a Save action, so nothing can be saved.
  - To test opening a Presentation, you can use _Empty Presentation_ in https://apps-stage-1.risevision.com/editor/list?cid=a847f5bb-e61a-4925-817a-552075edf203

Tests failed in the last push because of a CircleCI issue. I'll rebuild as soon as it's fixed.

@Rise-Vision/apps @santiagonoguez @stulees please review


web/scripts/editor/services/svc-template-editor-factory.js
```diff
@@ -0,0 +1,81 @@
+'use strict';
+
+angular.module('risevision.editor.services')
```
alex-deaconu: @fjvallarino should we not just create a new entity `risevision.template-editor.services`? Thanks.

web/scripts/editor/controllers/ctr-presentation-list.js
```diff
@@ -24,12 +25,23 @@ angular.module('risevision.editor.controllers')
...
+        'templateEditorFactory.loadingPresentation'], function (newValues) {
+        if (!newValues[0] && !newValues[1] && !newValues[2]) {
           $loading.stop('presentation-list-loader');
+        } else {
+          $loading.start('presentation-list-loader');
         }
       });
+
+      $scope.openPresentation = function(presentation) {
+        if (presentation.presentationType !== HTML_TEMPLATE_TYPE) {
```
santiagonoguez: @fjvallarino why is HTML_TEMPLATE_TYPE ( 'HTMLTemplates' ) used and not HTML_PRESENTATION_TYPE ( 'HTML Template' )  ?

fjvallarino: @santiagonoguez for `Presentation Type` I used what you defined in Core's `PresentationType` (HTML Template). The problem comes with what we use in `Product`'s `productTag`. For the existing templates, we use _Templates_. If we used _HTML Templates_ (following the naming convention from the existing type of using plural) for the new ones, because of how Document Search works, any query requesting _Templates_ would also include _HTML Templates_. It's not currently an issue (we only query from the Products Modal), but I'd rather avoid future problems.

web/scripts/editor/services/svc-template-editor-factory.js
fjvallarino: @alex-deaconu sounds good. I guess the same applies for controllers. Should we also create a separate folder for html partials?

web/scripts/editor/controllers/ctr-presentation-list.js
santiagonoguez: @fjvallarino so if I'm understanding correctly this presentation.presentationType comes from Product.productTag and not from Presentation.presentationType.

web/scripts/editor/services/svc-template-editor-factory.js
```diff
@@ -0,0 +1,81 @@
...
+          })
+          .finally(function () {
+            factory.loadingPresentation = false;
+          });
+
+        return deferred.promise;
+      };
+
+      var _showErrorMessage = function (action, e) {
+        factory.errorMessage = 'Failed to ' + action + ' Presentation.';
```
santiagonoguez: @fjvallarino if this is presented to the user using a messageBox, shouldn't it be localized ?

web/scripts/editor/controllers/ctr-presentation-list.js
fjvallarino: @santiagonoguez yes. It's fixed now (I forgot to update Presentation List when switching to an extra constant)

web/scripts/editor/services/svc-template-editor-factory.js
fjvallarino: @santiagonoguez the problem is we rely on another service (`processErrorCode`) which builds the message based on what the backend sends, which is not in a format ready for localization. We decided at that point it was not worth the effort, but if needed we should re-evaluate and add an extra card. Just for reference, that logic is basically the same we do on Presentations/Displays/Schedules (i.e., it's not something new).

santiagonoguez: I agree it's maybe not worth the effort now

web/scripts/editor/services/svc-template-editor-factory.js
```diff
@@ -0,0 +1,81 @@
...
+        _clearMessages();
+
+        factory.presentation = {
+          id: undefined,
+          productId: productDetails.productId,
+          name: 'Copy of ' + productDetails.name,
+          presentationType: HTML_PRESENTATION_TYPE,
+          revisionStatusName: undefined,
+          isTemplate: false,
+          isStoreProduct: false
```
santiagonoguez: @fjvallarino I think we should add here
templateAttributeData: '{}' 
so an empty attribute data gets created.

web/scripts/editor/services/svc-template-editor-factory.js
alex-deaconu: @fjvallarino yes, I think new folders for both JS and partials would be useful. Thanks.

web/scripts/editor/services/svc-template-editor-factory.js
stulees: @fjvallarino @santiagonoguez If it's the same logic we already apply for Presentations/Displays/Schedules, I agree lets not pursue this now and can be an improvement in the future.

web/scripts/template-editor/services/svc-template-editor-factory.js
```diff
@@ -14,6 +14,7 @@ angular.module('risevision.editor.services')
           productId: productDetails.productId,
           name: 'Copy of ' + productDetails.name,
           presentationType: HTML_PRESENTATION_TYPE,
+          templateAttributeData: {},
```
santiagonoguez: @fjvallarino it should be "{}", as the templateAttributeData won't be traveling as a JSON object, but as a string, thanks !

fjvallarino: @santiagonoguez I think this is the correct state when creating a Presentation. This is what the new Editor receives and what it works with. We will take care of serializing to a String when creating/updating the Presentation. We should also take care of deserializing when retrieving it (I'll add that now, while I'm at it).

santiagonoguez: @fjvallarino oh, nice ! thanks !

fjvallarino:
@alex-deaconu @santiagonoguez updates are in place. Thanks!

fjvallarino:
@alex-deaconu @santiagonoguez sorry for the delay. I had forgotten to move the uni tests and partials to their own folder (separate from the existing `editor` one). Thanks!

stulees:
@fjvallarino 👍 Looks good

web/scripts/template-editor/services/svc-template-editor-factory.js
santiagonoguez: @fjvallarino is the invocation to Core that adds the new presentation implemented in this PR, or that comes later ?

fjvallarino: @santiagonoguez I assumed that would be handled in the task _Execute saving template_ (from Execution Strategy). There is nothing to be saved at this point, and no element to interact with that performs that task

alex-deaconu:
@fjvallarino reviewed and PR looks good. I did not however spend too much time looking at specifics so I will defer approval to @Rise-Vision/content. Thanks!

web/scripts/template-editor/services/svc-template-editor-factory.js
santiagonoguez: @fjvallarino that means the presentation won't exist in the DataStore until it's saved by the user using the presentation editor Save button. Makes sense if we want the user to personalize it before he can schedule it.

@stulees @sheadarlison

fjvallarino: @santiagonoguez not necessarily. If we decide it should be created at this point, we can call the endpoint when the user selects the template. I don't know how that should work and, given we have a task specifically for _Execute saving template_, I preferred not to do anything related to it

santiagonoguez: My only concern with this is that this description seems to imply it's created when the user selects it to be opened by the editor:

https://docs.google.com/document/d/1frVqdkFY66l-y_hOaWvmF-x8ARKKA8BptRnCdLTWRBs/edit#heading=h.21azjcnyk1yi

Thoughts ?

fjvallarino: @santiagonoguez in my opinion we should not save until the user actually clicks on a Save button, for consistency with the existing Editor, but I don't want to decide that (you guys have been working on the UX and Design). The objective of this PR was to:

- Should open new editor if Template type is HTML
- On selection, should open new editor if template type is HTML
- Load HTML Templates in Templates Modal

I think we can decide how to handle Saving in a separate task (maybe the one I mentioned)

PR #820, opened 2019-02-12
Auto save HTML template upon user selection

stulees:
This implements saving of a HTML Template, with a temporary execution of it happening upon user selecting the template from the template list. This is temporary because we have not built Template editor yet and we need to move forward with other tasks in this milestone that require an HTML template to already be saved. We will eventually revise when a HTML template is saved when we get into building the Template Editor. 

I have validated the changes work with a new sub company (trial started) and the test product _HTML Template Product Test_. - https://apps-stage-9.risevision.com/editor/list?cid=a489ce69-e0fb-496c-b3d3-49643923adf6

### DIFF
--- a/web/scripts/template-editor/services/svc-template-editor-factory.js
+++ b/web/scripts/template-editor/services/svc-template-editor-factory.js
@@ -23,7 +23,8 @@ angular.module('risevision.template-editor.services')
       };
 
       factory.addPresentation = function () {
-        var deferred = $q.defer();
+        var deferred = $q.defer(),
+          presentationVal = JSON.parse(JSON.stringify(factory.presentation));
 
         _clearMessages();
 
@@ -31,9 +32,9 @@ angular.module('risevision.template-editor.services')
         factory.loadingPresentation = true;
         factory.savingPresentation = true;
 
-        JSON.stringify(factory.presentation.templateAttributeData);
+        presentationVal.templateAttributeData = JSON.stringify(presentationVal.templateAttributeData);
 
-        presentation.add(factory.presentation)
+        presentation.add(presentationVal)
           .then(function (resp) {
             if (resp && resp.item && resp.item.id) {
               $rootScope.$broadcast('presentationCreated');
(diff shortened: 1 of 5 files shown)


web/scripts/template-editor/services/svc-template-editor-factory.js
```diff
@@ -22,6 +22,42 @@ angular.module('risevision.template-editor.services')
...
+      factory.addPresentation = function () {
+        var deferred = $q.defer();
+
+        _clearMessages();
+
+        //show loading spinner
+        factory.loadingPresentation = true;
+        factory.savingPresentation = true;
+
+        JSON.stringify(factory.presentation.templateAttributeData);
```
santiagonoguez: @stulees is this ok ? it stringifies the attribute data, but is not assigning it anywhere, or am I missing something ?

stulees: @santiagonoguez It's not ok, I overlooked it. It has no impact in test coverage. I've revised in latest commit, thanks.

fjvallarino: @stulees wondering if it's worth it to validate `templateAttributeData` is turned to a JSON object when calling `presentation.add(presentationVal)`?

fjvallarino: @stulees sorry, to a String

fjvallarino: @stulees sorry for the multi message: I mean, I think it would be worth adding a line in the unit tests to check `templateAttributeData` is actually converted to a String when calling `presentation.add` (the reverse of what is tested for `getPresentation`)

stulees: @fjvallarino I've improved on the unit tests to ensure the `templateAttributeData` value is stringified when provided to the `presentation.add()` call, but the `factory.presentation.templateAttributeData` remains intact (not stringified). Please review last commit, thanks!

fjvallarino: @stulees thanks!

PR #823, opened 2019-02-14
Add missing writable fields when adding presentation [stage-8]

stulees:
`productId`, `presentationType`, and `templateAttributeData` need to be writable fields for adding presentation

### DIFF
--- a/web/scripts/editor/services/svc-presentation.js
+++ b/web/scripts/editor/services/svc-presentation.js
@@ -4,7 +4,8 @@
 
 angular.module('risevision.editor.services')
   .constant('PRESENTAION_WRITABLE_FIELDS', [
-    'name', 'layout', 'distribution', 'isTemplate', 'embeddedIds'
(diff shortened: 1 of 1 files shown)


web/scripts/editor/services/svc-presentation.js
```diff
@@ -4,7 +4,8 @@
 
 angular.module('risevision.editor.services')
   .constant('PRESENTAION_WRITABLE_FIELDS', [
-    'name', 'layout', 'distribution', 'isTemplate', 'embeddedIds'
+    'name', 'layout', 'distribution', 'isTemplate', 'embeddedIds', 'productId',
+    'presentationType', 'templateAttributeData'
```
santiagonoguez: @stulees that's why presentations are not opening in the template editor ?

Thanks !

stulees: @santiagonoguez That's correct.

PR #824, opened 2019-02-15
Feature/overall editor layout

santiagonoguez:
This separates the template editor in 4 sections: toolbar, attribute editor, preview, and footer. This is intended to achieve separation of those sections, and help reduce blocks other cards may have in this one.

Styles to match Mat's design are still incomplete.
Style code in the views are preliminary, will be separated later into common-header.

Preliminary view here:
https://apps-stage-9.risevision.com/templates/edit/d7e74f95-8c7c-495e-a3da-76fb514049c1?cid=cf85e5c4-9439-40ce-94d6-e5ea6ea2411c

Check also responsive behavior.

### DIFF
--- a/web/index.html
+++ b/web/index.html
@@ -237,6 +237,9 @@
   <script src="scripts/editor/directives/dtv-presentation-name.js"></script>
   <script src="scripts/editor/directives/dtv-rv-popover.js"></script>
 
+  <script src="scripts/template-editor/directives/dtv-attribute-editor.js"></script>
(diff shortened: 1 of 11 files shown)


santiagonoguez:
@fjvallarino yes, styles come after, the idea is to merge this on Monday first so work on attribute editor or other parts can start if needed without needing to wait for styles of the other sections. But with the new cards maybe that won't be an issue anyway.

PR #825, opened 2019-02-17
Use productCode instead of productId for HTML Templates

fjvallarino:
@Rise-Vision/apps @stulees @santiagonoguez please review


ezequielc:
@fjvallarino 👍

PR #832, opened 2019-02-21
Add Attribute Editor layout

fjvallarino:
This PR adds the layout of Attribute Editor, plus also loading Blueprint data from `blueprint.json`. I need to move styles into Common Header, but I wanted to validate this looks fine. The only concern I have is if we should care about environments for `blueprint.json` or just using a fixed one (currently `beta`) is ok.

You can test with this company: https://apps-stage-1.risevision.com/editor/list?cid=87977ab8-38b6-47fb-ad5e-256b8cc4b46d

- `Copy of Example Financial Template` will show the editor correctly
- The other presentations will fail to load the `blueprint.json` file and show an error message
- You can create new presentations with `Example Financial Template` product

@stulees @santiagonoguez please review


web/partials/template-editor/attribute-editor.html
```diff
@@ -1 +1,3 @@
-attribute editor
+<div class="col">
+    <template-attribute-list class="attribute-list"></template-attribute-list>
```
santiagonoguez: @fjvallarino did you separate this list into another fragment so you can show and hide this list and the upcoming detail view depending on some state ?

web/partials/template-editor/attribute-list.html
```diff
@@ -0,0 +1,41 @@
+<style>
+  .attribute-editor .attribute-list {
+    font-size: 1.1em;
+    font-weight: bold;
```
santiagonoguez: I think you should validate the fonts and colors and spacing look OK with @mmeiers

web/partials/template-editor/attribute-list.html
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  .attribute-row {
```
santiagonoguez: @fjvallarino seems to me that the top padding of rows need a little more space.

web/scripts/template-editor/services/svc-template-editor-factory.js
```diff
@@ -1,8 +1,11 @@
 'use strict';
 
 angular.module('risevision.template-editor.services')
-  .factory('templateEditorFactory', ['$q', '$log', '$state', '$rootScope', 'messageBox', 'presentation', 'processErrorCode', 'userState', 'HTML_PRESENTATION_TYPE',
-    function ($q, $log, $state, $rootScope, messageBox, presentation, processErrorCode, userState, HTML_PRESENTATION_TYPE) {
+  .constant('BLUEPRINT_URL', 'https://storage.googleapis.com/widgets.risevision.com/beta/templates/PRODUCT_CODE/blueprint.json')
```
santiagonoguez: @fjvallarino  As you say, beta works for now, but in the future 'stable' will be a better fit here ( and we should remember to change it before the epic finishes, maybe add a note to the execution strategy, thoughts @stulees ? ).

Also you could eliminate storage.googleapis.com and call directly:
https://widgets.risevision.com/beta/templates/PRODUCT_CODE/blueprint.json

web/partials/template-editor/attribute-editor.html
fjvallarino: @santiagonoguez yes, exactly, so we can apply the same `fade-in` effect we use in the other editor when navigating. Does that work for you, or did you guys have something else in mind?

santiagonoguez: @fjvallarino works for me, thanks !

PR #834, opened 2019-02-23
Add navigation in Attribute Editor and sample placeholders

fjvallarino:
@santiagonoguez @stulees please review


web/partials/template-editor/component-financial.html
```diff
@@ -0,0 +1,3 @@
+<div class="col">
```
santiagonoguez: @fjvallarino @stulees I'm wondering if using something like web/partials/template-editor/financial/component.html

as we'll potentially will have several components, and maybe we don't want every file under template-editor/ folder. Thoughts ?

fjvallarino: @santiagonoguez makes sense, yes. I think we can keep `component.html` (in charge of including the different components) in the root and the specific components in a subfolder like `web/partials/template-editor/components` with, potentially, a subfolder per component?

stulees: @fjvallarino @santiagonoguez I agree with specific components  in a subfolder `web/partials/template-editor/components/` and keep `component.html` in root of `web/partials/template-editor/`. I'm not sure the name of the file is accurate though as it's singular. Perhaps `component-list.html`?

fjvallarino: @stulees it's not exactly a list, since it will only hold a single component at once. Maybe `current-component.html`? I'm fine with either name

stulees: @fjvallarino Okay lets stick your original name `component.html`

fjvallarino:
@santiagonoguez @stulees I moved styles to Common Header and added a couple of translations. I will update CH branch when the corresponding PR is approved and merged

PR #835, opened 2019-02-24
Name UI and editing feature

stulees:
- I spent too much time trying different approaches to get the input field to autosize on user input while maintaining responsive styling (ellipsis, etc) but had troubles. I found [Stretchy](https://leaverou.github.io/stretchy/) which is a script precisely for form element auto sizing (i.e input element). It allows a filter (class name) to only apply it specific elements and it works exactly as needed. I forked the repo and am including as a bower component. 
- The presentation name is disabled by default until user clicks pencil button
- The pencil button is disabled while editing the presentation name.
- When pencil icon is clicked, the focus is set to presentation name input and cursor position is set to end of text. 
- When user focuses out of presentation name input, presentation name input is disabled and pencil button is enabled. 

You can validate at https://apps-stage-8.risevision.com/templates/edit/8dfa4f2f-c89d-4a61-ae8f-44e8cbc01c26?cid=a489ce69-e0fb-496c-b3d3-49643923adf6


alex-deaconu:
@stulees as a general rule we don't import new libraries into Apps unless strictly necessary. I believe bootstrap has plenty of classes that pertain to field width at various resolutions. I would suggest investigating that solution before hacking it in with a new library.

Thanks!

web/partials/template-editor/toolbar.html
```diff
@@ -1,9 +1,53 @@
+<style>
+  .edit-name {
```
alex-deaconu: @stulees please do not use inline CSS in partials. Thanks!

web/index.html
```diff
@@ -301,5 +302,8 @@
 
   <!-- endbuild -->
 
+  <!-- Apply Stretchy filter so it is only applied to '.input-stretchy' class -->
+  <script>Stretchy.selectors.filter = ".input-stretchy, .input-stretchy *";</script>
```
alex-deaconu: @stulees please do not use inline Scripts in the index.html file. Thanks!

web/partials/template-editor/toolbar.html
stulees: @alex-deaconu This is temporary, so it's easy to make any changes based on Creative's validation. Once approved I will move the styles to common-header and then update this PR. We have already been doing this process in previous PRs. Thanks.

stulees:
@alex-deaconu The use of the Sticky library is not for responsive width styling. It's for auto-resizing of the element based on user real-time input. This is not a feature provided by Bootstrap and was particularly tricky to try and implement while maintaining responsive styling with ellipsis (desktop vs mobile). This is why I chose to use this Sticky script. I would not consider this hacked, the author is well known and the script is widely used. 

If you feel it is not acceptable to use, then I spoke with the team and we will have to compromise as we've invested our maximum time into developing this feature. We will have to set a fixed width to the input field for desktop and abandon the feature of auto-resizing. 

Let me know, thanks.

stulees:
@fjvallarino I've made changes  in the last few commits based on Matt's feedback and also moved the Sticky filter setting from an inline script to being executed in the `run` of `apps.js`. Please review, thanks

stulees:
@fjvallarino FYI I've removed the inline styles and updated common-header dependency. Validated again on https://apps-stage-8.risevision.com/templates/edit/8dfa4f2f-c89d-4a61-ae8f-44e8cbc01c26?cid=a489ce69-e0fb-496c-b3d3-49643923adf6 and all is good.

PR #836, opened 2019-02-25
Handle Presentation deletion on Template Editor

fjvallarino:
@stulees @santiagonoguez please review


stulees:
@fjvallarino Looks good pending common-header dependency change once you deploy common-header 👍

PR #843, opened 2019-03-04
Add instrument search service - popular and keyword search [stage-8]

stulees:
- Adding `instrumentSearchService` to retrieve instruments
  -  _popularSearch_  returns most common instruments based on _category_ 
  - _keywordSearch_ returns any instruments that match _keyword_ and _category_

### DIFF
--- a/web/scripts/template-editor/services/svc-instrument-search.js
+++ b/web/scripts/template-editor/services/svc-instrument-search.js
@@ -1,0 +1,43 @@
+'use strict';
+
+angular.module('risevision.template-editor.services')
+  .constant('INSTRUMENT_SEARCH_BASE_URL', '"https://contentfinancial2.appspot.com/_ah/api/financial/v1.00/"')
+  .service('instrumentSearchService', ['$q', '$log', '$http', 'INSTRUMENT_SEARCH_BASE_URL',
+    function ($q, $log, $http, INSTRUMENT_SEARCH_BASE_URL) {
+      var factory = {},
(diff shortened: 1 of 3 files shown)


web/scripts/template-editor/services/svc-instrument-search.js
```diff
@@ -0,0 +1,43 @@
...
+        results = {
+          keyword: {},
+          popular: {}
+        };
+
+      factory.keywordSearch = function( category, keyword ) {
+        var keywordProp = category + "|" + keyword;
+
+        if ( results.keyword[ keywordProp ] ) {
+          return $q.when( results.keyword[ keywordProp ] );
```
santiagonoguez: @stulees this is a cache, right ?

stulees: @santiagonoguez That's correct

fjvallarino:
@stulees looks good 👍

web/scripts/template-editor/services/svc-instrument-search.js
fjvallarino: @stulees should we consider expiring this cache at some point, or do the instrument prices last long enough for this to be not needed?

stulees: @fjvallarino The instruments returned in these endpoint calls do not return actual financial data for the instruments, that is a separate (license based) financial request that is facilitated by our `rise-data-financial` web component in the HTML template page. The instruments returned in these endpoint calls here will not change frequently and therefore we can just rely on caching in memory to reduce the amount of endpoint calls to make.

PR #846, opened 2019-03-04
Redirect unauthenticated template requests to signup

alex-deaconu:
Do not check canAccessApps for editor substates
Otherwise it overwrites call from main editor state

Refactor and simplify editor state logic

Fix blank state encoding issue in Common Header

[stage-3]

### DIFF
--- a/web/scripts/app.js
+++ b/web/scripts/app.js
@@ -371,15 +371,14 @@ angular.module('risevision.apps', [
                 }
 
                 return canAccessApps(signup).then(function () {
-                  if ($stateParams.presentationId && $stateParams.presentationId !== 'new') {
-                    return editorFactory.getPresentation($stateParams.presentationId);
-                  } else if (!$stateParams.copyPresentation) {
-                    if (copyOf) {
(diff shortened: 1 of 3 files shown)


alex-deaconu:

web/scripts/app.js
```diff
@@ -387,31 +392,16 @@ angular.module('risevision.apps', [
             return $templateCache.get('partials/editor/artboard.html');
           }],
           reloadOnSearch: false,
-          controller: 'ArtboardController',
-          resolve: {
-            canAccess: ['canAccessApps',
```
olegrise: @alex-deaconu to confirm, there is no need to resolve `canAccess` here because it's already resolved in the parent `apps.editor.workspace` state. Is that right?

alex-deaconu: @olegrise that is correct. Actually, when calling `canAccessApps(true)` in `apps.editor.workspace`, this second call was overwriting the value and redirecting twice to the authentication page. That's how I found out about this in the first place, so I had to correct it for the `signUp` redirect mechanism to work correctly. Thanks.

PR #848, opened 2019-03-05
adds functions to set and retrieve template attribute data [stage-9]

santiagonoguez:
This defines two functions to easily get and set attribute data from a presentation.

These functions also initialize the data structure for a component if it's still not present in the attribute data object.

Unit tests are included also.


santiagonoguez:
@stulees please review this

PR #849, opened 2019-03-05
get blueprint data [stage-9]

santiagonoguez:
I added a function similar to the previous PR, but this ones lets extract attributes from the blueprint structure.

I added unit tests.

### DIFF
--- a/web/scripts/template-editor/controllers/ctr-template-editor.js
+++ b/web/scripts/template-editor/controllers/ctr-template-editor.js
@@ -7,7 +7,8 @@ angular.module('risevision.template-editor.controllers')
       $scope.presentation = presentation;
 
       $scope.getBlueprintData = function(componentId, attributeKey) {
-        var component = _.find($scope.components, {id: componentId});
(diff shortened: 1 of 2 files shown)


web/scripts/template-editor/controllers/ctr-template-editor.js
```diff
@@ -6,6 +6,24 @@ angular.module('risevision.template-editor.controllers')
       $scope.factory = templateEditorFactory;
       $scope.presentation = presentation;
 
+      $scope.getBlueprintData = function(componentId, attributeKey) {
+        var component = _.find($scope.components, {id: componentId});
```
stulees: @santiagonoguez Where is `$scope.components` instantiated? Isn't it within the scope of attribute list directive?

santiagonoguez: @stulees yes, I'll correct, thanks !

santiagonoguez: @stulees it's corrected, thanks again !

PR #863, opened 2019-03-14
Enforce literal value for component instance labels [stage-8]

stulees:
cc: @Rise-Vision/apps

### DIFF
--- a/web/partials/template-editor/attribute-list.html
+++ b/web/partials/template-editor/attribute-list.html
@@ -3,7 +3,7 @@
     <i class="mr-2 fa fa-lg {{ getComponentIcon(comp) }}"></i>
     <span>
       {{ ( "template." + comp.type ) | translate }} -
-      {{ comp.label | translate }}
+      {{ comp.label }}
     </span>
   </div>
   <div class="col-xs-2 pr-0 align-right">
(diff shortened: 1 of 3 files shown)


web/partials/template-editor/attribute-list.html
```diff
@@ -3,7 +3,7 @@
     <i class="mr-2 fa fa-lg {{ getComponentIcon(comp) }}"></i>
     <span>
       {{ ( "template." + comp.type ) | translate }} -
-      {{ comp.label | translate }}
+      comp.label
```
santiagonoguez: @stulees you still need the double braces, don't you ? otherwise you'll be writing a literal string

web/partials/template-editor/component.html
```diff
@@ -9,7 +9,7 @@
     <i class="ml-2 mr-2 fa fa-lg {{ getComponentIcon(factory.selected) }}"></i>
     <span>
       {{ ( "template." + factory.selected.type ) | translate }} -
-      {{ factory.selected.label | translate }}
+      factory.selected.label
```
santiagonoguez: @stulees same as above

web/partials/template-editor/attribute-list.html
stulees: @santiagonoguez Whoops, sloppy mistake, fixed, thanks

Issue #862, opened 2019-03-13
Clicking Save on a new Template added from the library creates a duplicate


pcsandford:
###  What is the expected behaviour?
After adding a Template the user can click Save if they have made changes or not and then add to their schedule and only that Template is added.

### What is the actual behaviour?
After adding a new Template click Save and it looks like nothing happens, even if a change has been made the Publish button is not active. Click Save a few more times and then look at the Presentations view and you will see duplicate Templates, all listed as published, for the amount of times you clicked Save.

If you go back to the Template from the Presentation list and click Save the Publish button will become active as normal. Clicking Save no longer creates a duplicate.

### What steps can be taken to replicate the problem?
- Add a Template from the Library.
- Click on Save a few times.
- Go to the Presentations view.

PatrickA-RiseVision:
This also happens if you make a copy of a presentation, and then click "Save".

olegrise:
@pcsandford @PatrickA-RiseVision the fix is deployed to production. Thanks.

PR #867, opened 2019-03-18
Saving a new presentation created from template creates a duplicate

olegrise:
Fix for  #862.

The issue was introduced by the [recent changes in the Apps router](https://github.com/Rise-Vision/rise-vision-apps/commit/ef3497940c12802e013e394b56c569c05a7628c1#diff-8ec8983b9f4c33db4ddae15c48d179b5L367). The order of checking `$stateParams.presentationId` and `$stateParams.copyPresentation` parameters was reversed and that caused the issue. @alex-deaconu FYI 😉

@fjvallarino please review.

### DIFF
--- a/web/scripts/editor/services/svc-editor-factory.js
+++ b/web/scripts/editor/services/svc-editor-factory.js
@@ -182,7 +182,8 @@ angular.module('risevision.editor.services')
               $rootScope.$broadcast('presentationCreated');
 
               $state.go('apps.editor.workspace.artboard', {
-                presentationId: resp.item.id
+                presentationId: resp.item.id,
+                copyPresentation: null
               }).then(function () {
                 scheduleFactory.createFirstSchedule(resp.item.id,
                     resp.item.name)
--- a/web/scripts/editor/services/svc-editor-factory.js
+++ b/web/scripts/editor/services/svc-editor-factory.js
@@ -183,7 +183,7 @@ angular.module('risevision.editor.services')
 
               $state.go('apps.editor.workspace.artboard', {
                 presentationId: resp.item.id,
-                copyPresentation: null
+                copyPresentation: undefined
               }).then(function () {
                 scheduleFactory.createFirstSchedule(resp.item.id,
                     resp.item.name)


web/scripts/editor/services/svc-editor-factory.js
```diff
@@ -182,7 +182,8 @@ angular.module('risevision.editor.services')
               $rootScope.$broadcast('presentationCreated');
 
               $state.go('apps.editor.workspace.artboard', {
-                presentationId: resp.item.id
+                presentationId: resp.item.id,
+                copyPresentation: null
```
fjvallarino: @olegrise wouldn't this be _undefined_ if not set explicitly? I'm not sure how the _null_ value affects the outcome

olegrise: @fjvallarino I tested both `null` and `undefined` and both produce the same result i.e. the parameter is removed from the URL. The `undefined` was my first choice, but then I saw other people use `null` and decided to go with `null`. I changed it back to `undefined`. Please review. I tested the change on my local machine. Thanks.

fjvallarino: @olegrise sorry, I was not really clear. I'm fine either way with `null` or `undefined`. My point is, I'm not sure how this is used. If we're checking for `copyPresentation` somewhere in our code as a boolean (i.e. `if (params.copyPresentation) { ... } `) then it does not really matter if it's `null` or `undefined`. This also means we don't really need to add `copyPresentation` to the map, since it would `undefined` by default. My question, then, was: do we really need to add `copyPresentation` to `$state.go('apps.editor.workspace.artboard', {})`?

olegrise: @fjvallarino I see. Let me recap what causes the issue. Here is how URL looks like when presentation is just created from the template and is **not saved yet** `http://localhost:8000/editor/workspace//true?cid=...`. The issue is that calling just `$state.go('apps.editor.workspace.artboard', {presentationId: resp.item.id})` adds the `presentationId` parameter, but **does not remove** the `copyPresentation`. 

I can think of 2 possible fixes: 
1) change the the logic in the router so it ignores `copyPresentation` if `presentationId` is present.
2) remove the `copyPresentation` parameter from the URL after presentation is saved because this parameter is no longer needed. This could be done in 2 ways - a) we can either set the parameter to undefined or b) use `inherit:false` option like this  `$state.go('apps.editor.workspace.artboard', {presentationId: resp.item.id} , {inherit:false});`.

fjvallarino: @olegrise I get it now. So, the option you implemented is *2*, right? If that's the case, I'm fine with it. Thanks for the explanation!

olegrise: @fjvallarino yes! Good questions and thanks for the reviews!

PR #869, opened 2019-03-21
Fix presentation name ellipsis mistakenly applied on initial render

stulees:
- Stop applying `max-width` to the presentation input
- Stretchy width calculation for input _text_ type needs 1 additional pixel to prevent `text-overflow` styling applying the _ellipsis_ unnecessarily.

Stretchy PR - https://github.com/Rise-Vision/stretchy/pull/1
Common-Header PR - https://github.com/Rise-Vision/common-header/pull/930

Validate with https://apps-stage-8.risevision.com/templates/edit/64200a83-7461-449c-9156-94f67cd24562?cid=36748004-5ec5-4432-bab6-8fdc7014edde

Feel free to re-save with different presentation name values. 

I will change the dependencies in bower.json to target real versions once you've validated/approved.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -15,9 +15,9 @@
     "Sortable": "^1.4.2",
     "angular-truncate": "*",
     "angular-slugify": "^1.0.1",
-    "common-header": "https://github.com/Rise-Vision/common-header.git#v3.5.11",
+    "common-header": "https://github.com/Rise-Vision/common-header.git#chore/presentation-name",
     "oauthio-web": "0.6.2",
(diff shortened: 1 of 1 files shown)


stulees:
@santiagonoguez That change does not affect the trash icon following the mockup. The title used in my validation presentation is very long, and thus the field is as wide as it can (on mobile) be to display the name until ellipsis is applied. As mentioned, feel free to re-save the name to whatever length so you can test different lengths.  
@sheadarlison @mmeiers

PR #876, opened 2019-03-26
Update Industry selector via CH version

alex-deaconu:
Use updated image for e2e tests

[stage-3]


alex-deaconu:
Forgot to set a Stage on the last commit so this will be on https://apps-stage-0.risevision.com

PR #877, opened 2019-03-26
Use latest custom Docker image

fjvallarino:
@Rise-Vision/apps @santiagonoguez @stulees please review


santiagonoguez:
@fjvallarino why don't you define it once and reference it thereafter like in other projects:

https://github.com/Rise-Vision/common-template/blob/master/.circleci/config.yml#L5-L19

fjvallarino:
@santiagonoguez updated. Waiting for Circle to feel like running the jobs

PR #881, opened 2019-03-27
Fix/mobile button position

santiagonoguez:
Conditionally change the height depending if a mobile browser is found or not.

Tested on both desktop and mobile browser:

common-header PR: https://github.com/Rise-Vision/common-header/pull/938
As usual, I will update the reference to it when it's approved.

### DIFF
--- a/web/scripts/editor/services/svc-presentation-utils.js
+++ b/web/scripts/editor/services/svc-presentation-utils.js
@@ -3,9 +3,15 @@
 angular.module('risevision.editor.services')
   .constant('HTML_PRESENTATION_TYPE', 'HTML Template')
   .factory('presentationUtils', ['HTML_TEMPLATE_TYPE', 'HTML_PRESENTATION_TYPE',
-    '$state', 'checkTemplateAccess', 'plansFactory',
-    function (HTML_TEMPLATE_TYPE, HTML_PRESENTATION_TYPE, $state, checkTemplateAccess, plansFactory) {
+    '$state', '$window', 'checkTemplateAccess', 'plansFactory',
+    function (HTML_TEMPLATE_TYPE, HTML_PRESENTATION_TYPE, $state, $window, checkTemplateAccess, plansFactory) {
       var factory = {};
+
+      factory.isMobileBrowser = function() {
+        return /Android|webOS|iPhone|iPad|iPod|BlackBerry|IEMobile|Opera Mini/i.test(
+          $window.navigator.userAgent
+        );
+      }
 
       factory.isHtmlTemplate = function (product) {
         return product.productTag && product.productTag.indexOf(HTML_TEMPLATE_TYPE) >= 0;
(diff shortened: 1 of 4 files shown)


web/scripts/template-editor/controllers/ctr-template-editor.js
```diff
@@ -9,6 +9,12 @@ angular.module('risevision.template-editor.controllers')
       $scope.isTestCompanySelected = userState.isTestCompanySelected;
       $scope.hasUnsavedChanges = false;
 
+      $scope.isMobileBrowser = function() {
+        return /Android|webOS|iPhone|iPad|iPod|BlackBerry|IEMobile|Opera Mini/i.test(
```
stulees: @santiagonoguez Do you know if we've provided this type of function elsewhere in Apps? If we haven't, this feels like it's more appropriate to be a reusable utility function. Thoughts?

santiagonoguez: @stulees I searched but couldn't find something in that project. 

I think it's a good idea, but where that utility function would be located ?

@fjvallarino

fjvallarino: @santiagonoguez @stulees we don't really have a global `utils` service. We do have `svc-presentation-utils.js`, were we put some logic common to both Editors

santiagonoguez: @fjvallarino thanks ! I'll use that !

web/scripts/editor/services/svc-presentation-utils.js
```diff
@@ -3,10 +3,16 @@
...
   .constant('HTML_PRESENTATION_TYPE', 'HTML Template')
   .factory('presentationUtils', ['HTML_TEMPLATE_TYPE', 'HTML_PRESENTATION_TYPE',
-    '$state', 'checkTemplateAccess', 'plansFactory',
-    function (HTML_TEMPLATE_TYPE, HTML_PRESENTATION_TYPE, $state, checkTemplateAccess, plansFactory) {
+    '$state', '$window', 'checkTemplateAccess', 'plansFactory',
+    function (HTML_TEMPLATE_TYPE, HTML_PRESENTATION_TYPE, $state, $window, checkTemplateAccess, plansFactory) {
       var factory = {};
 
+      factory.isMobileBrowser = function() {
+        return /Android|webOS|iPhone|iPad|iPod|BlackBerry|IEMobile|Opera Mini/i.test(
```
stulees: @santiagonoguez Curious, is this a list you sourced somewhere?

santiagonoguez: @stulees yes, it's the top answer here ( not depending on jquery ):

https://stackoverflow.com/questions/3514784/what-is-the-best-way-to-detect-a-mobile-device-in-jquery

PR #888, opened 2019-04-01
Populate industry from URL [stage-3]

olegrise:


alex-deaconu:
@olegrise tested on stage and works like a charm! Looks good 👍 . Thanks.

PR #895, opened 2019-04-09
Show correct status when Licensed but Cancelled

alex-deaconu:
If a Product was cancelled or suspended, the status shows as red
Update the gadget status when it is in the licensed state

Refactor code by splitting into directives and filters

### DIFF
--- a/web/scripts/editor/filters/ftr-subscription-status-message.js
+++ b/web/scripts/editor/filters/ftr-subscription-status-message.js
@@ -4,29 +4,14 @@ angular.module('risevision.editor.filters')
   .filter('subscriptionStatusMessage', ['$filter',
     function ($filter) {
       return function (gadget) {
-        var _getRemainingDays = function (date) {
-          var oneDay = 24 * 60 * 60 * 1000; // hours*minutes*seconds*milliseconds
-          var today = new Date();
-          return Math.round(Math.abs((date.getTime() - today.getTime()) / (
-            oneDay)));
+        var _getMessage = function () {
+          if (gadget.isSubscribed && !gadget.isLicensed) {
+            return gadget.subscriptionStatus;
+          } else {
+            return $filter('translate')('editor-app.subscription.status.professional');            
+          }
         };
 
-        var _getMessage = function () {
-          var statusMessage = gadget.subscriptionStatus;
-          if (gadget.isLicensed) {
-            statusMessage = $filter('translate')
-              ('editor-app.subscription.status.professional');
-          } else if (gadget.subscriptionStatus === 'Not Subscribed' && gadget.trialPeriod > 0) {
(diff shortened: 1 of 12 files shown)


web/scripts/editor/filters/ftr-subscription-status-message.js
```diff
@@ -0,0 +1,33 @@
...
+            oneDay)));
+        };
+
+        var _getMessage = function () {
+          var statusMessage = gadget.subscriptionStatus;
+          if (gadget.isLicensed) {
+            statusMessage = $filter('translate')
+              ('editor-app.subscription.status.professional');
+          } else if (gadget.subscriptionStatus === 'Not Subscribed' && gadget.trialPeriod > 0) {
+            statusMessage = statusMessage + ' - ' + gadget.trialPeriod +
```
olegrise: @alex-deaconu do you think this message is outdated considering that we don't sell widgets separately anymore? Should it be "For Licensed Displays" instead?

alex-deaconu: @olegrise I tend to agree. To confirm, we would show the "For Licensed Displays" message for all cases except for when they're Subscribed directly to the Widget. While the message is the same, the colour would be dependent on whether they have a Plan (Green) or not (Red).
@ezequielc FYI. Thoughts? Thanks.

ezequielc: @alex-deaconu @olegrise Good suggestion. Thanks

alex-deaconu:
@Rise-Vision/apps implemented as Oleg suggested inline, please review last 2 commits. Thanks!

olegrise:
@alex-deaconu There is an inconsistency in case when I don't have a plan - the status message "For Licensed Displays" is red in placeholder playlist and green on the modal. BTW, it seems it was like that before your change, so I don't know if it's expected behaviour or not.

inconsistency

alex-deaconu:
@olegrise I did not make any changes to the classes of either of those elements. In fact the only states that are `red` in the Playlist Item Modal are defined in the CH css (cancelled, expired, suspended):
https://github.com/Rise-Vision/common-header/blob/master/src/scss/ui-components/subscription.scss#L30

Whereas the class `.editor-playlist-item .playlist-item-status` is always present in the Placeholder Playlist Item, so it will always be red whenever that notice is shown (when not subscribed and not licensed).

I don't think we need to change this with this fix. @ezequielc thoughts? Thanks.

ezequielc:
@alex-deaconu I agree with you, I don't think we need to change it with this fix, even though it is a weird styling. Thanks

olegrise:
@ezequielc @alex-deaconu sounds good. Thanks for confirming!

PR #900, opened 2019-04-11
Add Page that allows user to configure own player

alex-deaconu:


alex-deaconu:
@Rise-Vision/apps please review. The UI is not finalized, will work on that in a subsequent PR. Thanks!

PR #907, opened 2019-04-16
Fix spacing for Add Display flow [stage-1]

alex-deaconu:


alex-deaconu:

web/scripts/config/stage.js
```diff
@@ -30,7 +30,6 @@
     .value('VIEWER_URL', 'http://stage-test-dot-rvaviewer-test.appspot.com')
     .value('ALERTS_WS_URL',
       'https://rvacore-test.appspot.com/alerts/cap')
-    .value('SEGMENT_API_KEY', null)
```
ezequielc: @alex-deaconu Should we revert this to `null` before merging? Thanks

alex-deaconu: @ezequielc good point, I will revert as part of the CH update tag commit. Thanks for the reminder!

alex-deaconu:
@ezequielc updated the Image tag to use the responsive class, thanks for the tip.

Last commit also includes reverting the config change, re-naming the variable as per typo and updating the CH tag.

Thanks.

ezequielc:
👍

PR #912, opened 2019-04-17
All Modals will not close on Esc or click outside

alex-deaconu:
Removed modal configuration overrides

[stage-1]

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -15,7 +15,7 @@
     "Sortable": "^1.4.2",
     "angular-truncate": "*",
     "angular-slugify": "^1.0.1",
-    "common-header": "https://github.com/Rise-Vision/common-header.git#v3.7.0",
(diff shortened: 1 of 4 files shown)


olegrise:
@alex-deaconu @ezequielc hmm, not too sure what to think about this. It was very convenient that you could close the modal by clicking outside of the modal or pressing Esc.

alex-deaconu:
@olegrise interesting point. This is kind of what this Issue & Fix is all about, disabling that functionality. @ezequielc any thoughts on Oleg's points? 

Also, thought I asked for reviews, please review. Thanks!

ezequielc:
@alex-deaconu @olegrise Yeah, good point. I use Esc a lot, it would be annoying to remove that, especially to fix something very minor. 
I think we should clarify that in the issue and close it without fix. Thoughts? Thanks

PR #915, opened 2019-04-19
Weekly templates to Presentations, Schedules and Displays list

ezequielc:
Fixed directive scope conflict
Fixed spacing
Fixed punctuation

@Rise-Vision/apps Please review

[stage-4]


alex-deaconu:
@ezequielc looks good to me. However, as discussed I think we should optimize to not load the templates list on every page change. Doesn't have to be done in this PR, maybe we can do that as an improvement. Thanks.

ezequielc:
@alex-deaconu Sounds good. Let's release it and work that as an improvement. Thanks!

PR #917, opened 2019-04-23
Feature/provide attribute data to html template

santiagonoguez:
Provides attribute data once there is a change ( including first state ) and only if the iframe code has already been loaded ( which was a problem for the first notification ).

Templates are now read from stable ( so they match blueprint file )
I also moved all preview related code to the preview directive.

I manually validated it here:

### DIFF
--- a/web/scripts/template-editor/services/svc-template-editor-factory.js
+++ b/web/scripts/template-editor/services/svc-template-editor-factory.js
@@ -3,10 +3,10 @@
 angular.module('risevision.template-editor.services')
   .constant('BLUEPRINT_URL', 'https://widgets.risevision.com/stable/templates/PRODUCT_CODE/blueprint.json')
   .constant('HTML_TEMPLATE_URL', 'https://widgets.risevision.com/stable/templates/PRODUCT_CODE/src/template.html')
-  .factory('templateEditorFactory', ['$q', '$log', '$sce', '$state', '$rootScope', '$http', 'messageBox', 'presentation', 'processErrorCode', 'userState',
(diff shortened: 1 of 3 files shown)


web/scripts/template-editor/services/svc-template-editor-factory.js
```diff
@@ -2,11 +2,12 @@
 
 angular.module('risevision.template-editor.services')
   .constant('BLUEPRINT_URL', 'https://widgets.risevision.com/stable/templates/PRODUCT_CODE/blueprint.json')
-  .constant('HTML_TEMPLATE_URL', 'https://widgets.risevision.com/beta/templates/PRODUCT_CODE/src/template.html')
-  .factory('templateEditorFactory', ['$q', '$log', '$sce', '$state', '$rootScope', '$http', 'messageBox', 'presentation', 'processErrorCode', 'userState',
-    'HTML_PRESENTATION_TYPE', 'BLUEPRINT_URL', 'HTML_TEMPLATE_URL', 'REVISION_STATUS_REVISED', 'REVISION_STATUS_PUBLISHED',
-    function ($q, $log, $sce, $state, $rootScope, $http, messageBox, presentation, processErrorCode, userState,
-      HTML_PRESENTATION_TYPE, BLUEPRINT_URL, HTML_TEMPLATE_URL, REVISION_STATUS_REVISED, REVISION_STATUS_PUBLISHED) {
+  .constant('HTML_TEMPLATE_URL', 'https://widgets.risevision.com/stable/templates/PRODUCT_CODE/src/template.html')
+  .constant('WIDGETS_DOMAIN', 'https://widgets.risevision.com')
```
andrecardoso: Maybe we should call this `HTML_TEMPLATE_DOMAIN`

santiagonoguez: @andrecardoso right, it's much better, thanks !

PR #919, opened 2019-04-24
calculate aspect ratio with unit tests [stage-9]

santiagonoguez:
This PR is just for the calculation of the aspect ratio value, along with some unit tests.

### DIFF
--- a/test/unit/template-editor/directives/dtv-editor-preview-holder.tests.js
+++ b/test/unit/template-editor/directives/dtv-editor-preview-holder.tests.js
@@ -69,6 +69,22 @@ describe('directive: TemplateEditorPreviewHolder', function() {
     expect(aspectRatio).to.equal("50.00");
   });
 
+  it('should calculate the 16:9 aspect ratio', function() {
+    factory.blueprintData = { width: "1920", height: "1080" };
+
+    var aspectRatio = $scope.getTemplateAspectRatio();
+
+    expect(aspectRatio).to.equal("56.25");
+  });
+
+  it('should calculate the 4:3 aspect ratio', function() {
+    factory.blueprintData = { width: "800", height: "600" };
+
+    var aspectRatio = $scope.getTemplateAspectRatio();
+
+    expect(aspectRatio).to.equal("75");
+  });
+
   it('should calculate a 333.33 aspect ratio', function() {
     factory.blueprintData = { width: "300", height: "1000" };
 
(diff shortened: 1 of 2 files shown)


test/unit/template-editor/directives/dtv-editor-preview-holder.tests.js
```diff
@@ -0,0 +1,80 @@
...
+
+  it('should calculate the 50 aspect ratio', function() {
+    factory.blueprintData = { width: "2000", height: "1000" };
+
+    var aspectRatio = $scope.getTemplateAspectRatio();
+
+    expect(aspectRatio).to.equal("50.00");
+  });
+
+  it('should calculate a 333.33 aspect ratio', function() {
```
stulees: @santiagonoguez I think it would be good to have some realistic width/height > aspect ratio calculation tests here that are based on actual template sizes. For example, all templates right now are 1920 x1080, which is a 16:9 aspect ratio, which should equal 56.25% as a percentage. 4:3 aspect ratio is apparently the 2nd most common that templates will be built for which 800x600 would be another test (should equal 75%).

web/scripts/template-editor/directives/dtv-editor-preview-holder.js
```diff
@@ -9,6 +9,9 @@ angular.module('risevision.template-editor.directives')
         link: function ($scope) {
           $scope.factory = templateEditorFactory;
 
+          var DEFAULT_TEMPLATE_WIDTH = 800;
```
stulees: @santiagonoguez So essentially we will default on a 4:3 aspect ratio, I'm assuming this is to coincide with the fixed size applied for mobile? Should this be the default also for desktop?

test/unit/template-editor/directives/dtv-editor-preview-holder.tests.js
santiagonoguez: @stulees good points, I'll add those tests. 

This particular one I added to ensure floating point doesn't get crazy with lots of decimals.

web/scripts/template-editor/directives/dtv-editor-preview-holder.js
santiagonoguez: @stulees I added this just to ensure a common aspect ratio is applied if some template ( or test template ) does not have the weight/height values assigned, but this value should not be used in most cases. Templates should always specify these values explicitly.

stulees: @santiagonoguez That's true, 👍

test/unit/template-editor/directives/dtv-editor-preview-holder.tests.js
santiagonoguez: @stulees I've added the tests you suggested !

PR #920, opened 2019-04-24
Replace deleteAllSubCompanies calls with deleteCurrentCompany

fjvallarino:
This improves the build situation a little bit, as it avoids having the sub company used by the current workflow step deleted by other step of the same build.

@santiagonoguez @stulees @Rise-Vision/apps please review


fjvallarino:
@santiagonoguez yes, that's correct. Anyway, I moved one of the `deleteCurrentSubCompany` into an `after` method, in order for it to be called even if some of the tests failed (if `before` fails, that's another matter, but it would be an issue with `deleteAllSubCompanies` too). The idea of using this method is that each test suite takes care only of its own sub company.

PR #921, opened 2019-04-25
Feature/use html template aspect ratio for preview

santiagonoguez:
This sets the style for aspect ratio for both desktop and mobile:

The blueprint file for testing here:

https://console.cloud.google.com/storage/browser/widgets.risevision.com/stable/templates/736a3d01551189406b9b302dccc6230267fe18ad/?project=avid-life-623&pli=1

Can be changed to test different aspect ratios.

Related common-header PR: https://github.com/Rise-Vision/common-header/pull/959

Some captures:

mobile_landscape

desktop_landscape

mobile_portrait

desktop_portrait

desktop_portrait_wide

Something to note is that in this last capture, as we are setting width to 100% and height is not controlled ( embedding classes ), a scroll bar appears on desktop mode. I tried to control the height to limit it to the page, but then it complicates setting the width ( as it's not fixed as in mobile ).

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -15,7 +15,7 @@
     "Sortable": "^1.4.2",
     "angular-truncate": "*",
     "angular-slugify": "^1.0.1",
-    "common-header": "https://github.com/Rise-Vision/common-header.git#v3.7.4",
+    "common-header": "https://github.com/Rise-Vision/common-header.git#feature/use-html-template-aspect-ratio-for-preview",
     "oauthio-web": "0.6.2",
     "stretchy": "https://github.com/Rise-Vision/stretchy.git#gh-pages"
   },
@@ -35,6 +35,7 @@
     "angular-ui-router": "^0.2.18",
     "angular-sanitize": "~1.5.0",
     "angular-mocks": "~1.5.0",
+    "common-header": "feature/use-html-template-aspect-ratio-for-preview",
     "jquery": "^2",
     "font-awesome": "~4.7.0"
   }
--- a/bower.json
+++ b/bower.json
@@ -15,7 +15,7 @@
     "Sortable": "^1.4.2",
     "angular-truncate": "*",
     "angular-slugify": "^1.0.1",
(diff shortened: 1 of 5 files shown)


fjvallarino:
@santiagonoguez have you tried using `calc` as we do for Attribute Editor? You can rely on `screen width - attribute editor width`. The problem I see, though, is maintaining aspect ratio will not be easy if only using CSS.

santiagonoguez:
@fjvallarino I'll experiment with that, but first I wanted both of you to take a look into how it's behaving now :)

santiagonoguez:
but maybe it will be very tricky, because if preview can be seen completely, we want to adjust bottom padding as we do now and let it be 100%; but only if preview height is greater than parent height we would want to keep height to 100% and adjust width ! thoughts ?

fjvallarino:
@santiagonoguez yeah, I agree. This cannot be solved with CSS alone; it requires some Javascript

stulees:
@santiagonoguez What is the width/height of the template supposed to be in the last screenshot capture you spoke to?

santiagonoguez:
@stulees that's a portrait width: 1080 height 1920
I've been doing tests, one option is to use javascript to set different styles depending on the aspect ratio being landscape or portrait; I'm still figuring out how to avoid those clashing with mobile mode, but I think it's doable.

santiagonoguez:
@stulees @fjvallarino it's now working using the approach mentioned above:

desktop_portrait_update

Validate: https://apps-stage-9.risevision.com/templates/edit/d4a5a9c7-2cc7-4f99-87cf-1e8457d0ea9a?cid=cf85e5c4-9439-40ce-94d6-e5ea6ea2411c

This is now fixed to:
  "width": "1080",
  "height": "1920",

but can be easily changed to validate landscape also.

https://console.cloud.google.com/storage/browser/widgets.risevision.com/stable/templates/736a3d01551189406b9b302dccc6230267fe18ad/?project=avid-life-623&pli=1

Please review again.

PR #923, opened 2019-04-29
Make Apps e2e test cases independent in order to be able to run multiple builds in parallel

fjvallarino:
The main changes here are:

- Updating to latest Widget Tester which supports an additional Stage Environment parameter
- Updating to latest Common Header which makes Sub Company handling be stage environment dependent
- Update build process to provide the necessary information to the e2e tests
- Make changes to Displays and Editor tests so they work with stage environment dependent ids
- Make several small changes to make tests more reliable
- Re enable Presentation Add tests (I honestly don't remember why I disabled them 7 months ago, based on git blame)

I pursued the Sub Company with stage environment suffix approach because I could not really figure out an easy way to implement the alternative of using a single Subscribed and a separate Free company for tests dependent on subscription status changes (it required a larger refactor and I didn't think I could make it in time).

I will update Common Header and Widget Tester tags once those PRs are approved. I will also reenable staging deployments before merging (I disabled that to be able to test builds without affecting other environments)

@santiagonoguez @stulees @Rise-Vision/apps please review

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -116,14 +116,15 @@ jobs:
           command: |
             # Set +e flag to fix issue where no stage environment is specified
             set +e
-            export STAGE_ENV="$(git log -1 --pretty=%B | grep '\[.*\]' |sed -e 's/.*\[\(.*\)\].*/\1/g')";
+            STAGE_ENV="$(git log -1 --pretty=%B | grep '\[.*\]' |sed -e 's/.*\[\(.*\)\].*/\1/g')";
             if [ "$STAGE_ENV" != '' ]; then
               echo "Deploying to $STAGE_ENV";
             fi;
             if [ "$STAGE_ENV" == '' ]; then
               echo "stage deployment environment missing. Deploying to [stage-0].";
-              export STAGE_ENV='stage-0';
+              STAGE_ENV='stage-0';
             fi;
+            echo 'export STAGE_ENV="$STAGE_ENV"' >> $BASH_ENV
       - run: echo "Testing e2e tests - Env $STAGE_ENV"
       - store_artifacts:
           path: ~/rise-vision-apps/dist.tar.gz
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -127,8 +127,8 @@ jobs:
             echo 'export STAGE_ENV="$STAGE_ENV"' >> $BASH_ENV
             echo "BASH_ENV is $BASH_ENV"
       - run:
-        - name: Do things
-        - command: |
+          name: Do things
+          command: |
             echo $BASH_ENV
             source $BASH_ENV
             echo "Testing e2e tests - Env $STAGE_ENV"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,7 +7,6 @@ commands:
       casesfile:
         type: string
         default: "invalid_path"
-    working_directory: ~/rise-vision-apps
     steps:
       - attach_workspace:
           at: ~/
@@ -143,6 +142,7 @@ jobs:
               STAGE_ENV='stage-0';
             fi;
             #rsync -rptz -e ssh --delete dist apps@webserver-stage.risevision.com:/rise-front-end/apps/$STAGE_ENV;
+      - run: tar czvf dist.tar.gz dist
       - store_artifacts:
           path: ~/rise-vision-apps/dist.tar.gz
   deploy_beta:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,8 +17,24 @@ commands:
       - run: sudo apt-get update -qq
       - run: sudo apt-get install -y google-chrome-stable
       - run:
+          name: Set environment variables
+          command: |
+            # Set +e flag to fix issue where no stage environment is specified
+            set +e
+            STAGE_ENV="$(git log -1 --pretty=%B | grep '\[.*\]' |sed -e 's/.*\[\(.*\)\].*/\1/g')";
+            if [ "$STAGE_ENV" != '' ]; then
+              echo "Deploying to $STAGE_ENV";
+            fi;
+            if [ "$STAGE_ENV" == '' ]; then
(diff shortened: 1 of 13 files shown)


.circleci/config.yml
```diff
@@ -129,7 +141,7 @@ jobs:
               echo "stage deployment environment missing. Deploying to [stage-0].";
               STAGE_ENV='stage-0';
             fi;
-            rsync -rptz -e ssh --delete dist apps@webserver-stage.risevision.com:/rise-front-end/apps/$STAGE_ENV;
+            #rsync -rptz -e ssh --delete dist apps@webserver-stage.risevision.com:/rise-front-end/apps/$STAGE_ENV;
```
santiagonoguez: @fjvallarino why was this line commented out ?

fjvallarino: @santiagonoguez that line does the actual deployment to each staging environment; it's one of the things I mentioned I will re enable before merging. I commented it out in order to test my builds without actually affecting anybody

santiagonoguez: ok, thanks !

stulees:
@alex-deaconu @ezequielc @olegrise Can you  please review this, we are looking to deploy soon as @fjvallarino is away tomorrow. Thanks

olegrise:
@fjvallarino looks good to me 👍

alex-deaconu:
@fjvallarino is there any reason you're creating a Sub Company with a name format like
`E2E TEST SUBCOMPANY - TEMPLATE ADD - stage2`
and then searching for 
`E2E TEST SUBCOMPANY  TEMPLATE ADD  stage2`?

Thanks.

test/e2e/editor/cases/presentation-add.js
```diff
@@ -12,7 +12,7 @@ var ArtboardPage = require('./../pages/artboardPage.js');
 var PresentationAddScenarios = function() {
 
   browser.driver.manage().window().setSize(1920, 1080);
-  xdescribe('Presentation Add', function () {
+  describe('Presentation Add', function () {
```
alex-deaconu: @fjvallarino these tests were disabled because of this failure:
https://circleci.com/gh/Rise-Vision/rise-vision-apps/21510

This will fail pretty reliably from our experience, that's why we disabled it. Thanks.

test/e2e/launcher/cases/weekly-templates.js
```diff
@@ -43,7 +46,7 @@ var WeeklyTemplatesScenarios = function() {
       browser.sleep(500);
       helper.wait(commonHeaderPage.getCommonHeaderMenuItems().get(1), 'Presentations Common Header Menu Item');
       helper.clickWhenClickable(commonHeaderPage.getCommonHeaderMenuItems().get(1), 'Presentations Common Header Menu Item');
-      helper.waitDisappear(commonHeaderPage.getLoader(), 'CH spinner loader');
```
alex-deaconu: @fjvallarino I would check for both spinners to disappear here. I believe they show one on top of the other. Thanks.

fjvallarino:
@alex-deaconu the removal of the dashes is required because Document Search does not return anything if they are present. I could also remove the dashes from the names but, since we were already using them, I preferred to leave them

test/e2e/editor/cases/presentation-add.js
fjvallarino: @alex-deaconu oh, that makes sense. They have not failed today in the ~10 runs I did. Do you think it's worth it leaving them enabled and disabling them later if they cause issues?

fjvallarino: @alex-deaconu the error in the one you linked is related to the Presentation Settings. I had a really bad time on those, but I think they're more stable now after all these changes

test/e2e/launcher/cases/weekly-templates.js
fjvallarino: @alex-deaconu makes sense, I'll add it back

test/e2e/editor/cases/presentation-add.js
alex-deaconu: @fjvallarino You're right! For some reason that particular click does not register sometimes. Let's leave at is. Thanks.

alex-deaconu:
@fjvallarino sounds good re: dashes. One more comment inline. Thanks.

.circleci/config.yml
```diff
@@ -129,7 +141,7 @@ jobs:
               echo "stage deployment environment missing. Deploying to [stage-0].";
               STAGE_ENV='stage-0';
```
alex-deaconu: @fjvallarino should this logic not be removed and the variable created above reused? Thanks.

test/e2e/launcher/cases/weekly-templates.js
fjvallarino: @alex-deaconu updated!

.circleci/config.yml
fjvallarino: @alex-deaconu I intended to do that, but environment variables are not saved between `run` steps. That's why I saved it to a file in one part and also had to duplicate logic in other. The alternative was creating a separate `.sh` file, but I somehow thought that was uglier (I'm not sure now)

alex-deaconu: @fjvallarino but you're saving it to a `stage.txt` file. Can that not be added in the `dependencies` step when the environment is saved, and just loaded by each step? Thoughts? thanks.

fjvallarino: @alex-deaconu I tried that, but `deploy-staging` sometimes runs before any of the e2e tests (since we removed the dependency to allow for faster stage deployments), and it fails since the file was not yet created

alex-deaconu: @fjvallarino see my note re: `dependencies` step. Thanks.

fjvallarino: @alex-deaconu that could work! Let me try it

alex-deaconu:
@fjvallarino I think the change worked. No further comments, looks great! Thanks!

PR #937, opened 2019-04-30
First draft of the attribute editor for Text component [stage-3]

olegrise:
@andrecardoso I cerated a directive that can load, edit, and save "value" attribute of the `rise-text` component. The value is saved when you press "Back" button. You can run Apps locally or on stage-3.


olegrise:
@andrecardoso I made a small change to the style and updated CH version. Please take another look and then I'll merge. Thanks.

andrecardoso:
@olegrise looks good. Thanks

PR #941, opened 2019-05-02
Fix/aspect ratio

santiagonoguez:
The calculations to avoid cropping on preview are implemented here. Most of the previous code structure is kept.

Related common-header PR: https://github.com/Rise-Vision/common-header/pull/966

### DIFF
--- a/test/unit/template-editor/directives/dtv-editor-preview-holder.tests.js
+++ b/test/unit/template-editor/directives/dtv-editor-preview-holder.tests.js
@@ -106,7 +106,7 @@ describe('directive: TemplateEditorPreviewHolder', function() {
 
       var aspectRatio = $scope.getMobileWidth();
 
-      expect(aspectRatio).to.equal("269");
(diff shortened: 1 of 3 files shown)


test/unit/template-editor/directives/dtv-editor-preview-holder.tests.js
```diff
@@ -106,15 +106,15 @@ describe('directive: TemplateEditorPreviewHolder', function() {
 
       var aspectRatio = $scope.getMobileWidth();
 
-      expect(aspectRatio).to.equal("284");
+      expect(aspectRatio).to.equal("340");
```
stulees: @santiagonoguez This is minor, but calling this variable `aspectRatio` is confusing since it's actually a `width`.

santiagonoguez: @stulees good point, I'll change it !

PR #944, opened 2019-05-02
Fix/preview bottom margin

santiagonoguez:
Solves:
- The preview bottom margin issue. Some browsers were initially reporting a slightly higher client width and height, which was corrected if the window was resized shortly after. I assume it's related to the browser's own layer size calculations. So as workaround I added a small delay for the calculation and it worked.
- I also considered the margin on the calculations for desktop width.
- The mobile problem with small heights, was solved by taking into account the short mobile rule here: https://github.com/Rise-Vision/common-header/blob/master/src/scss/sections/template-editor-layout.scss#L44

Demo: https://apps-stage-9.risevision.com/templates/edit/c88d9cfd-5d49-467f-8e3e-0615c7e5fdf6?cid=cf85e5c4-9439-40ce-94d6-e5ea6ea2411c

### DIFF
--- a/web/scripts/template-editor/directives/dtv-editor-preview-holder.js
+++ b/web/scripts/template-editor/directives/dtv-editor-preview-holder.js
@@ -15,7 +15,7 @@ angular.module('risevision.template-editor.directives')
           var MOBILE_PREVIEW_HEIGHT_SHORT = 140;
           var MOBILE_MARGIN = 10;
           var DESKTOP_MARGIN = 20;
-          var PREVIEW_INITIAL_DELAY_MILLIS = 2000;
+          var PREVIEW_INITIAL_DELAY_MILLIS = 1000;
 
           var iframeLoaded = false;
           var attributeDataText = null;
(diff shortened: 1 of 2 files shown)


web/scripts/template-editor/directives/dtv-editor-preview-holder.js
```diff
@@ -12,8 +12,10 @@ angular.module('risevision.template-editor.directives')
           var DEFAULT_TEMPLATE_WIDTH = 800;
           var DEFAULT_TEMPLATE_HEIGHT = 600;
           var MOBILE_PREVIEW_HEIGHT = 200;
+          var MOBILE_PREVIEW_HEIGHT_SHORT = 140;
           var MOBILE_MARGIN = 10;
           var DESKTOP_MARGIN = 20;
+          var PREVIEW_INITIAL_DELAY_MILLIS = 2000;
```
stulees: @santiagonoguez 2 seconds is noticeable as a user experience - preview previously showed much quicker. Did you base 2 seconds on a specific reason?

santiagonoguez: @stulees let me experiment with a lower interval

santiagonoguez: @stulees with 1 second it works well, but no less.

I also added an immediate aspect ratio calculation so preview appears immediately even if it updates it's height shortly after.

PR #945, opened 2019-05-02
Always save text changes in Text component [stage-3]

olegrise:
@andrecardoso please review. There is no callback for the "Save" button, so the component's directive can't capture the moment when this button is pressed. I opted out for saving on every text change in `<input>` element. I think this is a better option as it gives user immediate visual feedback in Preview.


andrecardoso:
looks good 👍 !

PR #962, opened 2019-05-09
Load default text data [stage-3]

andrecardoso:


andrecardoso:
[(2) Load initial default component data on editor](https://trello.com/c/dTpYBGeE/44-2-load-initial-default-component-data-on-editor)

PR #964, opened 2019-05-10
Fix text component editor styles

andrecardoso:


andrecardoso:
[(1) Adjust text component editor UI](https://trello.com/c/gnJeyF3L/25-1-adjust-text-component-editor-ui)

PR #977, opened 2019-05-16
Add Storage Selector to Image Component

fjvallarino:
This PR implements Storage selector. This includes:

- Navigating folders back and forth
- Selecting files and adding them to the Image List
- Uploading files into a Storage folder

This is only a functional version, which means it looks really ugly. Styling will be taken care of in the next Trello card.

You can test here (if possible, do not save the changes so your tests do not interfere with other people's): https://apps-stage-7.risevision.com/templates/edit/1919032c-a981-42ac-98fb-0d0d5731c978?cid=87977ab8-38b6-47fb-ad5e-256b8cc4b46d


fjvallarino:
@santiagonoguez @stulees please review

web/scripts/template-editor/components/directives/dtv-basic-storage-selector.js
```diff
@@ -1,24 +1,115 @@
...
+          $scope.fileNameOf = function (path) {
+            var parts = path.split('/');
+
+            if ($scope.isFolder(path)) {
+              return parts[parts.length - 2] + '/';
             }
+            else {
+              return parts.pop();
+            }
+          };
```
santiagonoguez: @fjvallarino we have something similar here ( just for files in the image list ) here:
https://github.com/Rise-Vision/rise-vision-apps/blob/master/web/scripts/template-editor/components/directives/dtv-component-image.js#L178

shouldn't we have a single function for this ?

web/scripts/template-editor/components/directives/dtv-component-image.js
```diff
@@ -10,31 +10,32 @@ angular.module('risevision.template-editor.directives')
...
-                selectedImages.splice(idx, 1, newFile);
-              } else {
-                selectedImages.push(newFile);
-              }
+              _setMetadata(selectedImages);
+            }
+          };
+          $scope.storageManager = {
+            addSelectedItems: function (newSelectedItems) {
+              var selectedImages = _.cloneDeep($scope.selectedImages);
```
santiagonoguez: @fjvallarino I think we'll be needing this conditional also here to account for default images ( or extract it to a function ):
https://github.com/Rise-Vision/rise-vision-apps/blob/master/web/scripts/template-editor/components/directives/dtv-component-image.js#L178

do you agree ?

web/scripts/template-editor/components/directives/dtv-basic-storage-selector.js
```diff
@@ -1,24 +1,115 @@
...
+
+          $scope.selectItem = function (item) {
+            var idx = _.findIndex($scope.selectedItems, { name: item.name });
+
+            if (idx >= 0) {
+              $scope.selectedItems.splice(idx, 1);
+            }
+            else {
+              $scope.selectedItems.push(item);
```
santiagonoguez: @fjvallarino I'm seeing this find / splice / push logic in lost of places, shouldn't we create some generic utility for that and reuse ? thanks !

web/scripts/template-editor/components/directives/dtv-basic-uploader.js
```diff
@@ -37,6 +35,10 @@ angular.module('risevision.template-editor.directives')
             }
           };
 
+          $scope.fileNameOf = function( path ) {
+            return path.split('/').pop();
```
santiagonoguez: again, aren't we repeating this from 
https://github.com/Rise-Vision/rise-vision-apps/blob/master/web/scripts/template-editor/components/directives/dtv-component-image.js#L178

fjvallarino:
@santiagonoguez I agree with the idea of having common code. The reason for having that duplicate logic is those directives use those functions internally only (it's not exposed to its users) and, mainly, that we don't have a place for common code. Looking a bit at the code structure we have, we could add a `svc-utils.js` in `template-editor/services` where we could put these (and other) common functions. Do you think that could work?

I will also be moving `dtv-basic-storage-selector.js` down one level (from `components` to `template-editor/directives`, since it does not really belong there.

santiagonoguez:
@fjvallarino I really like the idea of svc-utils,

and about the storage directive moved down, I think it's a good choice also, because eventually we'll have video ( and maybe some other components that need storage ) and we'll need to reuse a big part of all these.

Thanks !

stulees:
@fjvallarino @santiagonoguez Working great! svc-utils makes sense for _template-editor/services_, I think we'll find more functionality will make sense to go there as we continue in further epics.

santiagonoguez:
@fjvallarino thanks for the changes

PR #978, opened 2019-05-17
Fix lint issues

fjvallarino:
@santiagonoguez @olegrise please review

### DIFF
--- a/web/scripts/template-editor/directives/dtv-basic-storage-selector.js
+++ b/web/scripts/template-editor/directives/dtv-basic-storage-selector.js
@@ -22,7 +22,9 @@ angular.module('risevision.template-editor.directives')
               $scope.isUploading = isUploading;
             },
             addFile: function (file) {
-              templateEditorUtils.addOrReplace($scope.folderItems, { name: file.name }, file);
+              templateEditorUtils.addOrReplace($scope.folderItems, {
+                name: file.name
+              }, file);
             }
           };
           $scope.storageManager = angular.extend($scope.storageManager, {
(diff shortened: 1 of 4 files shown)


web/scripts/template-editor/directives/dtv-basic-storage-selector.js
```diff
@@ -60,37 +62,42 @@ angular.module('risevision.template-editor.directives')
...
-              console.log('Failed to load files', err);
-            })
-            .finally(function () {
-              $loading.stop(spinnerId);
-            });
           };
 
           $scope.selectItem = function (item) {
-            templateEditorUtils.addOrRemove($scope.selectedItems, { name: item.name }, item);
+            templateEditorUtils.addOrRemove($scope.selectedItems, {
```
andrecardoso: I think the lint friendly version is worse for reading. Can we tweak the rules to allow this?

olegrise:
@fjvallarino if I understand it correctly, the changes in this PR is a result of using [jsbeautifier](https://github.com/Rise-Vision/rise-vision-apps/blob/master/gulpfile.js#L142), not lint.
 
@andrecardoso I don't mind if we tweak the rules, but I'm also okay with the current rules.

web/scripts/template-editor/directives/dtv-basic-storage-selector.js
fjvallarino: @andrecardoso I totally agree! There are a bunch of things the beautifier does that makes the situation worse. Maybe we should all discuss what we want our code to look like?

fjvallarino:
@olegrise yeah, it's the beautifier, not the linter. I'm going to close this PR until we reach an agreement on how we want the code to look like

@andrecardoso @santiagonoguez FYI

olegrise:
@fjvallarino up to you, but I would merge this PR and then work on reaching an agreement. You've got 3 approvals 😉

PR #988, opened 2019-05-23
Fix scope problems when loading multiple component directives 

ezequielc:
[stage-2]

When multiple components were used, $scope.save() was being overwritten in every component directive as the $scope was shared between them and only the last directive would work. I set individual scopes to prevent that from happening.

I updated unit tests, manually tested and didn't find problems with this approach.

@alex-deaconu Please review.

@stulees @andrecardoso @olegrise @fjvallarino @santiagonoguez fyi.
Thanks!


alex-deaconu:
@ezequielc looks good to me. Thanks for fixing 👍

olegrise:
@ezequielc Thanks for taking care of this!

stulees:
@ezequielc 👍

PR #989, opened 2019-05-23
Show current folder and corresponding icon in navigation bar

fjvallarino:
This PR shows the current folder name in the navigation bar when using Template Editor's Storage. It follows the styling as defined in Zeplin, which shows "Rise Storage" when in root, using an svg icon, and the folder name with 'fa-folder' when inside a specific folder.

- Root: https://app.zeplin.io/project/5c58a8a2b74a443081a59e09/screen/5ca038ed50d7a876f7748612
- Folder: https://app.zeplin.io/project/5c58a8a2b74a443081a59e09/screen/5ca038ed99837405aba3d750

You can test it here, using "Image Component" and clicking on "Select from Storage": https://apps-stage-7.risevision.com/templates/edit/808cf50c-199a-4f2e-bb0a-e80937c21614?cid=87977ab8-38b6-47fb-ad5e-256b8cc4b46d

I modified the existing `component-icon` directive in order to allow using some predefined SVG icons we have in Common Header (in particular, the one for Storage). I switched to `html` instead of `replaceWith` since I was having issues with duplicate icons when navigating back and forth. Oleg, please let me know if you are fine with this change!

There is also the issue that you can't really change the color of an SVG image once it's created. Since this is the Component Icon directive and it should match the colors we use for fonts, I'm setting the fill color while creating the SVG image, although it's not really clean. If you have ideas on how to improve this, please let me know. I evaluated exposing a `color` attribute that would be provided from the HTML partials, but I'm not sure it would be much of an improvement.

@olegrise @santiagonoguez @stulees please review

### DIFF
--- a/web/scripts/template-editor/directives/dtv-component-icon.js
+++ b/web/scripts/template-editor/directives/dtv-component-icon.js
@@ -16,9 +16,9 @@ angular.module('risevision.template-editor.directives')
                 $scope.icon + '</svg>';
             } else if ($scope.type === 'riseSvg') {
               return '<svg class="mr-2 fa fa-lg" viewBox="0 0 32 32" width="24" height="18" xmlns="http://www.w3.org/2000/svg">' +
-                       '<path fill="#4d4d4d" d="' + iconsList.icons1[$scope.icon] + '"></path>' +
-                       '<path fill="#4d4d4d" d="' + iconsList.icons2[$scope.icon] + '"></path>' +
-                     '</svg>';
+                '<path fill="#4d4d4d" d="' + iconsList.icons1[$scope.icon] + '"></path>' +
+                '<path fill="#4d4d4d" d="' + iconsList.icons2[$scope.icon] + '"></path>' +
+                '</svg>';
             } else {
               return '<i class="mr-2 fa fa-lg ' + $scope.icon + '"></i>';
             }
--- a/web/scripts/template-editor/directives/dtv-component-icon.js
+++ b/web/scripts/template-editor/directives/dtv-component-icon.js
@@ -15,9 +15,9 @@ angular.module('risevision.template-editor.directives')
               return '<svg class="mr-2 fa fa-lg" width="24px" height="18px" viewBox="0 0 24 18" xmlns="http://www.w3.org/2000/svg">' +
                 $scope.icon + '</svg>';
             } else if ($scope.type === 'riseSvg') {
-              return '<svg class="mr-2 fa fa-lg" viewBox="0 0 32 32" width="24" height="18" xmlns="http://www.w3.org/2000/svg">' +
-                '<path fill="#4d4d4d" d="' + iconsList.icons1[$scope.icon] + '"></path>' +
-                '<path fill="#4d4d4d" d="' + iconsList.icons2[$scope.icon] + '"></path>' +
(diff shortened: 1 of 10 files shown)


web/scripts/template-editor/directives/dtv-component-icon.js
```diff
@@ -14,14 +14,19 @@ angular.module('risevision.template-editor.directives')
             if ($scope.type === 'svg') {
               return '<svg class="mr-2 fa fa-lg" width="24px" height="18px" viewBox="0 0 24 18" xmlns="http://www.w3.org/2000/svg">' +
                 $scope.icon + '</svg>';
+            } else if ($scope.type === 'riseSvg') {
+              return '<svg class="mr-2 fa fa-lg" viewBox="0 0 32 32" width="24" height="18" xmlns="http://www.w3.org/2000/svg">' +
+                '<path fill="#4d4d4d" d="' + iconsList.icons1[$scope.icon] + '"></path>' +
+                '<path fill="#4d4d4d" d="' + iconsList.icons2[$scope.icon] + '"></path>' +
+                '</svg>';
```
santiagonoguez: @fjvallarino  SVG elements are part of the DOM and follow CSS rules, so an option that may be much simpler is to set the color using CSS.

fjvallarino: @santiagonoguez actually, that's the best option. I didn't know about the `fill`/`stroke` combination. I'll add it now!

fjvallarino:
@santiagonoguez updated to use styling from Common Header

web/scripts/template-editor/directives/dtv-component-icon.js
```diff
@@ -14,14 +14,19 @@ angular.module('risevision.template-editor.directives')
...
             } else {
               return '<i class="mr-2 fa fa-lg ' + $scope.icon + '"></i>';
             }
           };
 
           $scope.$watch('icon', function (icon) {
             if (icon) {
-              element.replaceWith(_html());
+              element.html(_html());
```
olegrise: @fjvallarino what was the reason for this change?

fjvallarino: @olegrise that's the issue I mentioned about getting duplicate elements (icons in this case) when navigating back and forth in Storage

fjvallarino: @olegrise it's most likely caused because the `ng-show` of the `component-icon` tag goes away after being replaced by the `svg` tag (which does not have that attribute)

olegrise: @fjvallarino I see. Oh, well. I really like that with `replaceWith` we didn't have big chunks of duplicate date in HTML.

Screen Shot 2019-05-23 at 4 54 39 PM

fjvallarino: @olegrise I understand what you mean. I'll try to come up with a solution/improvement next week. I'll merge this PR since I'm off tomorrow and we also can't deploy on Friday. Thanks!

olegrise: @fjvallarino sounds good. Thanks.

PR #991, opened 2019-05-24
Populate initial data for components on Preview load

ezequielc:
Populate initial data for components on Preview load. For now, the only use is to set company Display Address to Weather components.

I created a factory to hold registered directives/components information. Long term idea is to move some of the methods from Attribute Editor directive to that factory, so that they can be easily accessible from other places.

@alex-deaconu Thanks for your help on the discussions. Please review. 
Thanks


alex-deaconu:
@ezequielc looks good, thanks for the refactoring!

PR #998, opened 2019-05-29
Send Display Data to non-editable Weather components

ezequielc:
Along with https://github.com/Rise-Vision/common-template/pull/61, these changes improve Preview so that is sends Start event and Display data to the Template, in order for all weather components to receive display data.

I also reverted some of the changes from the previous strategy, as they are no longer relevant.

[stage-1]
To validate the changes, please proxy common-template.min.js requests to widgets.risevision.com/staging/common/2019.05.29.18.33/common-template.min.js


web/scripts/template-editor/directives/dtv-editor-preview-holder.js
```diff
@@ -31,7 +29,9 @@ angular.module('risevision.template-editor.directives')
 
             _postAttributeData();
 
-            _setupComponents();
+            _postStartEvent();
```
alex-deaconu: @ezequielc just to confirm, I don't believe this call is necessary as `_postAttributeData()` already calls the start event within RisePlayerConfiguration: https://github.com/Rise-Vision/common-template/blob/master/src/rise-attribute-data.js#L92. However, we can keep this here for good measure. Thanks.

alex-deaconu:

alex-deaconu:

web/scripts/template-editor/directives/dtv-editor-preview-holder.js
ezequielc: @alex-deaconu Well noticed. I agree, let's keep it. Thanks

Issue #932, opened 2019-04-30
Template Editor: Template could not be deleted without being saved first.


sheadarlison:
Video attached to card: https://trello.com/c/7167eN4X/6799-template-editor-issues

**Steps**

1. Add Presentation
2. Delete presentation without saving
3. Notification - “Failed to delete presentation. The presentation could not be found. Please validate it still exists.”

alex-deaconu:
@sheadarlison this issue is now resolved. To confirm, the issue occurs because the HTML Template doesn't actually exist since it hasn't been saved. The error is expected behaviour.

To fix this, we are disabling the Delete button until save.

Thanks!

sheadarlison:
👍  Thanks @alex-deaconu

PR #1003, opened 2019-05-31
Disable Delete button if HTML Template is not saved

alex-deaconu:
[stage-1]


alex-deaconu:
Fix for #932 

Disabling the button until save. To confirm, the issue occurs because the HTML Template doesn't actually exist since it hasn't been saved.

Issue #930, opened 2019-04-30
Template Editor Issue: URL behaves strangely when pushing save 


sheadarlison:
Video: https://www.screencast.com/t/W9WjxsFWR2C

alex-deaconu:
@sheadarlison the duplicate `cid` parameter issues should now be resolved. Closing. Thanks!

sheadarlison:
👍 Thanks @alex-deaconu

Issue #929, opened 2019-04-30
Template Editor: Back button doesn't trigger modal


sheadarlison:
Video:  https://www.screencast.com/t/uKs0mncEhvD

How it should work: When I click back the first time I am presented with the warning modal.

alex-deaconu:
@sheadarlison this issue should now be resolved. The additional page in the browser history should no longer be generated, so the warning should come up right away when the back button is clicked. Closing. Thanks!

sheadarlison:
👍 Thanks @alex-deaconu

PR #1004, opened 2019-05-31
Fix duplicate cid parameter added to the route

alex-deaconu:
`apps.editor.templates` route inherits `apps.editor`
additional cid parameter is not needed

add template editor routes to editor navStates

[stage-2]

### DIFF
--- a/web/scripts/app.js
+++ b/web/scripts/app.js
@@ -406,7 +406,7 @@ angular.module('risevision.apps', [
         })
 
         .state('apps.editor.templates', {
-          url: '/templates?cid',
+          url: '/templates',
           abstract: true,
           template: '<div class="templates-app" ui-view></div>'
         })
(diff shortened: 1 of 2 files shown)


alex-deaconu:
@ezequielc you reviewed before I added my story 💤 so here it is:

The additional `cid` parameter is making the routing behave badly. Basically, on a page change (add/edit) two `cid` parameters are added, and in some cases one is removed automatically, but this causes another page change, and more specifically 2 history tokens to be generated. Because of this, the back button hits the 2 cid token instead of going back.

Fixing this addresses both #929 and #930 and maybe some other issues related to this that aren't logged.

Thanks for the review!

ezequielc:
@alex-deaconu Oh wow, even #929 gets fixed by this. Nice!

olegrise:
Nice!

Issue #969, opened 2019-05-13
Template Editor:  When I attempted to delete the template it asked would you like to delete followed by a window asking if I wanted to save my work.


RowanDarlison:
**Steps to recreate:**
(only happens if template is not used in a schedule otherwise it shows the msg that the template is used in a schedule and cannot be deleted)
1. Add Presentation.
2. Save presentation
3. Change Presentation
4. Delete presentation
5. Notification - “Are you sure you want to delete this presentation”
6. Clicked delete forever
7. Notification - “Do you want to save changes before leaving? Your changes will be lost if you don’t save them.” given options; save, don’t save, cancel.
8. Clicked save - nothing happened
9. Clicked don’t save - presentation deleted.
10. Clicked cancel - app breaks and template turns white, when buttons are clicked you see a white screen with words "AccessDeniedAccess denied.Anonymous caller does not have storage.objects.get access to widgets.risevision.com/stable/templates/undefined/src/template.html.". Can’t click back.

alex-deaconu:
@RowanDarlison this issue should now be resolved. Thanks!

RowanDarlison:
Hey @alex-deaconu I went through the same steps and all the errors are still happening.

alex-deaconu:
@RowanDarlison wondering if you cleared your cache before testing this? I just tried it again and I don't get any notifications. Thanks.

RowanDarlison:
Yes sorry looks good now!

PR #1014, opened 2019-06-05
Fix editor routing issues and save notification on delete

alex-deaconu:
Editor and Template Editor routes will no longer refresh on first save

Deleting a Template will no longer cause the unsaved changes
notification to pop up
This is achieved by having both a synchronous and asynchronous unsaved
changes setter and using it in different instances

[stage-2]


alex-deaconu:
Fix for #969 and some additional improvements.

PR #1016, opened 2019-06-05
Don't have unsaved changes when opening components

alex-deaconu:
[stage-2]


alex-deaconu:
To replicate, add a new Template, open any component without making any changes and try to navigate away from the page. It happens to an existing Saved Template if the Component settings were never opened before.

PR #1017, opened 2019-06-05
Slides validation

olegrise:


olegrise:
@andrecardoso I think this is it for this PR. For tests and UI work we can create a separate PR. I reviewed your changes - they look good. Please review mine. Thanks.

PR #1022, opened 2019-06-10
Update editor and template routing

alex-deaconu:
Add routes are now persistent
Add routes now include original productId or templateId

[stage-2]


alex-deaconu:

PR #1023, opened 2019-06-10
Re-validation Slides URL

olegrise:
@andrecardoso please review

### DIFF
--- a/web/scripts/template-editor/components/directives/dtv-component-slides.js
+++ b/web/scripts/template-editor/components/directives/dtv-component-slides.js
@@ -64,6 +64,7 @@ angular.module('risevision.template-editor.directives')
               element.show();
               $scope.componentId = $scope.factory.selected.id;
               _load();
+              $scope.saveSrc(); //validate Slides URL
             }
           });
 
--- a/web/scripts/template-editor/components/directives/dtv-component-slides.js
+++ b/web/scripts/template-editor/components/directives/dtv-component-slides.js
@@ -10,8 +10,8 @@ angular.module('risevision.template-editor.directives')
         link: function ($scope, element) {
           $scope.factory = templateEditorFactory;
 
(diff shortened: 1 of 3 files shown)


andrecardoso:
[(2) Validate on blur](https://trello.com/c/oqkRWIuC/109-2-validate-on-blur)

olegrise:
@andrecardoso please review last commit. The changes are:
- check if directive is visible. My original assumption was that `templateComponentSlides` is instantiated every time when <rise-slides> component is selected in the Editor. Turns out that it is instantiated only once when user opens a presentation editor. That's why it's necessary to check if it's visible or not.
- removed check for `$scope.validationError !== 'VALID'`. I realized that we don't save the `'VALID'` state in `$scope.validationError`. Saving it there would make logic a bit more complex, so I decided to remove this optimization (i.e. not revalidating if it's last validation was successful) for the sake of simplicity.

web/scripts/template-editor/components/directives/dtv-component-slides.js
```diff
@@ -58,9 +64,16 @@ angular.module('risevision.template-editor.directives')
...
               $scope.componentId = $scope.factory.selected.id;
               _load();
+              $scope.saveSrc(); //validate Slides URL
             }
           });
 
+          function _directiveIsVisible() {
+            // This directive is instantiated once by templateAttributeEditor
+            // It becomes visible when <rise-slides> is selected
+            return $scope.factory.selected && ($scope.factory.selected.type === 'ise-slides');
```
andrecardoso: Is this correct? Shouldn't it be 'rise-slides' ?

olegrise: @andrecardoso good catch! I wonder how that happen. Fixed. See the last commit.

Issue #1026, opened 2019-06-10
Template Editor: Financial attribute list shows empty list message while loading data


ezequielc:
### What happens
Behind the loading spinner, a message says that are no symbols in the list:

<img width="1128" alt="Screen Shot 2019-06-10 at 6 05 03 PM" src="https://user-images.githubusercontent.com/7738553/59226908-6e230200-8baa-11e9-8d4a-0694a16d4a57.png">

### What should happen
It should show a blank empty state, instead of a misleading message.

stulees:
This has been fixed from changes in #1031

PR #1031, opened 2019-06-12
Prevent empty list UI showing during list initialization

stulees:
- Both Financial and Image empty list UI now leverage `factory.loadingPresentation` as a conditional on showing or not, which is only applicable during an initial build of the list from a new presentation. 
- This fixes issue #1026 

Validate with https://apps-stage-8.risevision.com/editor/list?cid=dd474bee-b237-46e3-aa20-98e975679773 and add a new _Example Financial Template_ presentation and click into Financial and then back and click into Image. You should not see the empty list UI in either scenario. 

@ezequielc FYI

### DIFF
--- a/web/partials/template-editor/components/component-image/image-list.html
+++ b/web/partials/template-editor/components/component-image/image-list.html
@@ -36,7 +36,7 @@
     </div>
   </div>
 
-  <div class="image-component-list-empty" ng-show="selectedImages.length === 0 && !isUploading">
+  <div class="image-component-list-empty" ng-hide="selectedImages.length !== 0 || factory.loadingPresentation">
     <div class="row">
       <div class="col-xs-12">
         <h2>You have no images here.</h2>
--- a/web/partials/template-editor/components/component-image/image-list.html
+++ b/web/partials/template-editor/components/component-image/image-list.html
@@ -10,7 +10,10 @@
     </div>
   </div>
 </div>
-<div class="image-component-list te-scrollable-container" ng-class="{'active-duration' : selectedImages.length > 1 && !isUploading}">
+<div class="image-component-list te-scrollable-container"
+     ng-class="{'active-duration' : selectedImages.length > 1 && !isUploading}"
+     rv-spinner rv-spinner-key="template-editor-loader"
+     rv-spinner-start-active="1">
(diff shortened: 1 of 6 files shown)


stulees:
@fjvallarino Yes I think we should for consistency. I'll add to this PR.

stulees:
@fjvallarino @santiagonoguez Spinner added, same staged location.

fjvallarino:
@stulees nice!!!

web/partials/template-editor/components/component-image/image-list.html
```diff
@@ -10,7 +10,10 @@
     </div>
   </div>
 </div>
-<div class="image-component-list te-scrollable-container" ng-class="{'active-duration' : selectedImages.length > 1 && !isUploading}">
+<div class="image-component-list te-scrollable-container"
+     ng-class="{'active-duration' : selectedImages.length > 1 && !isUploading}"
+     rv-spinner rv-spinner-key="template-editor-loader"
```
olegrise: @stulees the `"template-editor-loader"` key is also used [here](https://github.com/Rise-Vision/rise-vision-apps/blob/66d2b69f14312abad10d93ddad56208b80e9a44e/web/partials/template-editor/components/component-financial.html#L5). You might get some undesired side effects.

olegrise: I mean both spinners with the same key are going to be part of the DOM at the same time.

stulees: @olegrise That key is shared between Image and Financial because they both need their list spinner to behave from the same $watch of `factory.loadingPresentation` value. See [here](https://github.com/Rise-Vision/rise-vision-apps/blob/master/web/scripts/template-editor/controllers/ctr-template-editor.js#L147-L152). They will be part of the DOM at same time but neither will be visible at the same time. I haven't seen any undesired effects from testing. Thanks

olegrise: @stulees I guess that's okay in this case. I was thinking that spinner directive itself might have a problem, but I reviewed its implementation and it's fine to have spinners with the same key. Thanks.

ezequielc:
@stulees Thanks for the fix!

PR #1032, opened 2019-06-12
Financial Data License Message

ezequielc:
When adding a new Template, checks if it uses `rise-data-financial` and shows a message telling users that a data license is required.

Also, styling refactoring related to https://github.com/Rise-Vision/common-header/pull/995.


web/scripts/template-editor/services/svc-template-editor-factory.js
```diff
@@ -62,6 +63,12 @@ angular.module('risevision.template-editor.services')
           });
       };
 
+      var _checkFinancialDataLicenseMessage = function (blueprintData) {
```
alex-deaconu: @ezequielc the `blueprintData` variable is not used here, since you use the factory value below. Thanks.

ezequielc: @alex-deaconu Good catch, I will fix that. Thanks

alex-deaconu:

PR #1041, opened 2019-06-13
Feature/auto save

santiagonoguez:
This is a big one, sorry ! But code changes aren't hard to follow, hopefully !

Implementation of auto-save:
- Auto-saves after changes to HTML template presentation
- Auto-saves when a new HTML template presentation is created
- Includes auto-save UI changes from @fjvallarino 
- Updates to unit and all template editor E2E tests

Two minor side effects remain that will need to be fixed in a following PR.

Validate - open one of the first presentations, and/or create a new one from example template or Indices:

https://apps-stage-9.risevision.com/editor/list?cid=cf85e5c4-9439-40ce-94d6-e5ea6ea2411c

### DIFF
--- a/test/e2e/template-editor/template-editor.cases.js
+++ b/test/e2e/template-editor/template-editor.cases.js
@@ -59,7 +59,7 @@
 
     //var templateEditorAddScenarios = new TemplateEditorAddScenarios();
     var financialComponentScenarios = new FinancialComponentScenarios();
-    //var textComponentScenarios = new TextComponentScenarios();
+    var textComponentScenarios = new TextComponentScenarios();
     //var weatherComponentScenarios = new WeatherComponentScenarios();
     //var imageComponentScenarios = new ImageComponentScenarios();
 
--- a/test/e2e/template-editor/template-editor.cases.js
+++ b/test/e2e/template-editor/template-editor.cases.js
@@ -59,9 +59,9 @@
 
     // Text component scenarios deal with the auto schedule modal, so they always should come first.
     var textComponentScenarios = new TextComponentScenarios();
-    //var templateEditorAddScenarios = new TemplateEditorAddScenarios();
-    //var financialComponentScenarios = new FinancialComponentScenarios();
-    //var weatherComponentScenarios = new WeatherComponentScenarios();
+    var templateEditorAddScenarios = new TemplateEditorAddScenarios();
+    var financialComponentScenarios = new FinancialComponentScenarios();
+    var weatherComponentScenarios = new WeatherComponentScenarios();
     var imageComponentScenarios = new ImageComponentScenarios();
 
     after(function() {
--- a/test/e2e/template-editor/template-editor.cases.js
+++ b/test/e2e/template-editor/template-editor.cases.js
@@ -57,9 +57,10 @@
       _selectSubCompany();
     });
 
+    // Text component scenarios deal with the auto schedule modal, so they always should come first.
+    var textComponentScenarios = new TextComponentScenarios();
     var templateEditorAddScenarios = new TemplateEditorAddScenarios();
     //var financialComponentScenarios = new FinancialComponentScenarios();
-    //var textComponentScenarios = new TextComponentScenarios();
     //var weatherComponentScenarios = new WeatherComponentScenarios();
     //var imageComponentScenarios = new ImageComponentScenarios();
 
--- a/test/e2e/template-editor/template-editor.cases.js
+++ b/test/e2e/template-editor/template-editor.cases.js
@@ -57,9 +57,9 @@
       _selectSubCompany();
     });
 
-    //var templateEditorAddScenarios = new TemplateEditorAddScenarios();
-    var financialComponentScenarios = new FinancialComponentScenarios();
-    var textComponentScenarios = new TextComponentScenarios();
+    var templateEditorAddScenarios = new TemplateEditorAddScenarios();
+    //var financialComponentScenarios = new FinancialComponentScenarios();
+    //var textComponentScenarios = new TextComponentScenarios();
     //var weatherComponentScenarios = new WeatherComponentScenarios();
     //var imageComponentScenarios = new ImageComponentScenarios();
 
--- a/test/e2e/template-editor/template-editor.cases.js
+++ b/test/e2e/template-editor/template-editor.cases.js
@@ -60,7 +60,7 @@
     // Text component scenarios deal with the auto schedule modal, so they always should come first.
     var textComponentScenarios = new TextComponentScenarios();
     var templateEditorAddScenarios = new TemplateEditorAddScenarios();
-    //var financialComponentScenarios = new FinancialComponentScenarios();
+    var financialComponentScenarios = new FinancialComponentScenarios();
     //var weatherComponentScenarios = new WeatherComponentScenarios();
     //var imageComponentScenarios = new ImageComponentScenarios();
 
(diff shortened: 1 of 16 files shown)


web/scripts/template-editor/controllers/ctr-template-editor.js
```diff
@@ -73,10 +79,53 @@ angular.module('risevision.template-editor.controllers')
         return component;
       }
 
+      function _getCurrentTimestamp() {
+        return new Date().getTime();
+      }
+
+      function _programSave() {
+        _saveTimeout = $timeout(function () {
+          _saveTimeout = null;
```
fjvallarino: @santiagonoguez could it happen that `_saveTimeout` is not null at this point (I'm aware of `_clearSaveTimeout`)? If it could be, should we cancel the timer it references?

santiagonoguez: @fjvallarino thanks, that's what [_reprogramSave()](https://github.com/Rise-Vision/rise-vision-apps/blob/203fbf4e2967709fb24337dc27f4541d59134a28/web/scripts/template-editor/controllers/ctr-template-editor.js#L108) is for:

The only other cases where _programSave() is being called are:
- [here](https://github.com/Rise-Vision/rise-vision-apps/blob/203fbf4e2967709fb24337dc27f4541d59134a28/web/scripts/template-editor/controllers/ctr-template-editor.js#L92): where the time has just expired and thus is not set.
- [and here](https://github.com/Rise-Vision/rise-vision-apps/blob/203fbf4e2967709fb24337dc27f4541d59134a28/web/scripts/template-editor/controllers/ctr-template-editor.js#L126): where we already checked that the timer is not set.

fjvallarino: @santiagonoguez curious, then: do we need to set it to null here?

santiagonoguez: @fjvallarino I think we need to clear the timer handle to indicate that it has expired and that it's no longer running. I don't think javascript auto-clears it, and even if it does, that line does not harm :P

web/scripts/template-editor/controllers/ctr-template-editor.js
```diff
@@ -73,10 +79,53 @@ angular.module('risevision.template-editor.controllers')
...
       var _bypassUnsaved = false,
         _initializing = false;
       var _setUnsavedChanges = function (state) {
         $scope.hasUnsavedChanges = state;
+
+        if ($scope.hasUnsavedChanges) {
+          if (_saveTimeout) {
+            var elapsed = _getCurrentTimestamp() - _lastSavedTimestamp;
+
+            if (elapsed < MAXIMUM_INTERVAL_BETWEEN_SAVES) {
```
stulees: @santiagonoguez I'm curious, when could `elapsed` be >= `MAXIMUM_INTERVAL_BETWEEN_SAVES`? And what should happen in that state?

web/scripts/template-editor/controllers/ctr-template-editor.js
fjvallarino: @santiagonoguez sorry, I had not seen you were relying on it being null (I saw you were setting it to null in `_reprogramSave` and missed `_setUnsavedChanges `). Cancelling the timer does not set it to null, that's correct 👍

web/scripts/template-editor/controllers/ctr-template-editor.js
santiagonoguez: @stulees this is related to a comment @fjvallarino make in the design document: https://docs.google.com/a/risevision.com/document/d/1KqmqzQWJ0nqVsYJu0Fysl5q27mW-VowMQ7AFHxFMfjo/edit?disco=AAAADKJKXxo

Basically the timer will reprogram each time a change is detected until we detect that the maximum is surpassed here. When that maximum is surpassed, the current timer is left to finish and save; we already know a timeout is in process because we have just checked the _saveTimeout variable a few lines above.

stulees: @santiagonoguez Okay makes sense, thanks for clarifying

stulees:
@santiagonoguez Can you clarify the behaviour I'm seeing - https://www.screencast.com/t/Y7AIC60X2eoP . Notice the auto save status is switching to unsaved and then saved once I enter the presentation, which subsequently enables Publish button. Then it does it another save status change/update upon navigating into Financial component. Neither situation has anything actually changed.

santiagonoguez:
@stulees 
1.- The first save you mention. We are saving when the presentation just opens to avoid the issue with the presentation being published the first time, as mentioned here: 
https://docs.google.com/document/d/1KqmqzQWJ0nqVsYJu0Fysl5q27mW-VowMQ7AFHxFMfjo/edit#heading=h.muv6fbam20lp

2.- When you enter a financial component or an image component, they get the financial labels and image thumbnails respectively; and then they update the attribute data with that information. So even if the user hasn't explicitly changed anything, the attribute data has been updated, and that triggers those saves. The E2E tests are expecting those saves if you examine them closely at the beginning of this PR.

Hope this clarifies. If you believe this may represent some confusion to the end user, we may want to revise the strategy for displaying the save status messages to the user.

santiagonoguez:
@stulees that being said, we may be triggering an unnecessary save when the presentation is being added for the first time, I'll check that, but even if that were the case, I'd leave that for a following PR, there are already a lot of changes in this one.

stulees:
@santiagonoguez Okay understood on both points. Regarding # 2 though, I do somewhat think it's confusing from a user perspective as it indicates they are doing something to cause changes when they actually aren't. I wonder if we should leverage flag for entering into a component and use it to avoid the change in UI. However, it might be best to wait and see on this one as this might introduce unnecessary complexity.

santiagonoguez:
@stulees the extra save when the presentation is being created is solved in the latest commit, and the E2E tests updated to reflect that. No big changes on the code were needed, mainly excluding some presentation fields from the comparison. Thanks !

stulees:
@santiagonoguez Nice one! Looks good 👍

test/e2e/template-editor/template-editor.cases.js
```diff
@@ -57,9 +57,10 @@
       _selectSubCompany();
     });
 
+    // Text component scenarios deal with the auto schedule modal, so they always should come first.
+    var textComponentScenarios = new TextComponentScenarios();
```
olegrise: @santiagonoguez sorry for the late question. Do this mean that all other scenarios below depend on the `textComponentScenarios`?

santiagonoguez: @olegrise yes, rise-text tests are now dealing with the auto-schedule dialog, that is opened for the first save on a new company:
https://github.com/Rise-Vision/rise-vision-apps/blob/master/test/e2e/template-editor/cases/components/text.js#L35

As a company is created every time the full suite of E2E tests are run we had to deal with this.

I moved this responsibility to text instead of the example template that had rise-data-financial to avoid the issue that I reported here:
https://github.com/Rise-Vision/rise-vision-apps/issues/1049

But maybe later we can put that responsibility back to where it was originally. 

And please, if you have more doubts about this or other changes done to E2E as a result of this PR, please contact me.

olegrise: @santiagonoguez thanks for the explanation. I'm good with this change. I just wanted to get a better understanding. Thanks!

PR #1051, opened 2019-06-17
Fix double validation in Slides component when user hits Enter [stage-3]

olegrise:
The `<form>` tag [commits view value](https://docs.angularjs.org/api/ng/type/ngModel.NgModelController#$commitViewValue) immediately on form submit event (i.e. on Enter) which in turn calls onChange event for the input field right away instead of waiting 1 second as defined by `debounce` option.


andrecardoso:
@olegrise Good catch!

olegrise:
@andrecardoso please review the last commit. Thanks.

andrecardoso:
@olegrise looks good. Thanks

PR #1053, opened 2019-06-18
Fix/issue 1042 race condition in rise image settings

santiagonoguez:
This reads the attribute data before updating thumbnail information, so deleted images are not re-added again.

Validation: open any presentation in the link below, and quickly remove images from it, they shouldn't reappear when thumbnail data is read again.
https://apps-stage-9.risevision.com/editor/list?cid=cf85e5c4-9439-40ce-94d6-e5ea6ea2411c

### DIFF
--- a/web/scripts/template-editor/components/directives/dtv-component-image.js
+++ b/web/scripts/template-editor/components/directives/dtv-component-image.js
@@ -104,6 +104,10 @@ angular.module('risevision.template-editor.directives')
             return $scope.getAttributeData($scope.componentId, key);
           }
 
+          function _setAttribute(key, value) {
+            $scope.setAttributeData($scope.componentId, key, value);
+          }
+
           function _getDefaultFilesAttribute() {
             return $scope.getBlueprintData($scope.componentId, 'files');
           }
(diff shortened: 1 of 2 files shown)


web/scripts/template-editor/components/directives/dtv-component-image.js
```diff
@@ -196,14 +200,41 @@ angular.module('risevision.template-editor.directives')
             }).join('|');
           }
 
+          $scope.updateImageMetadata = function (metadata) {
```
andrecardoso: Is this function going to be used in the HTML? If not, why adding it to `$scope`?

santiagonoguez: @andrecardoso I added to $scope just to be able to test that function from the unit tests. I evaluated doing the test starting with show() but setting it up would result in a very convoluted test implementation. Another alternative would be to extract the function as a service, but I don't see much value in doing that as its implementation is now very coupled to the rest of the state in this directive, and I don't think it helps in clarity or other way. So I don't see another nice alternative for the moment. Thoughts ?

andrecardoso: Got it. I'm good with keeping it if it makes testing easier. Thanks.

PR #1055, opened 2019-06-19
Fix for issue 1015 - Cancelled subscription shows "Managed by parent" for subcompany

olegrise:


olegrise:
Fix for https://github.com/Rise-Vision/store-server/issues/1015

PR #1056, opened 2019-06-19
Feature/volume pricing

tejohnso:
Go-Live date for volume pricing is Monday. Here are the apps changes. There isn't much here as most of the changes are in common-header.

Live on [stage 6](https://apps-stage-6.risevision.com)


web/pricing-component.mjs
```diff
@@ -0,0 +1,114 @@
+import {PolymerElement, html} from "https://unpkg.com/@polymer/polymer@3.2.0/polymer-element.js?module"
+import "https://widgets.risevision.com/stable/components/pricing/pricing-data-component.mjs"
+import "https://widgets.risevision.com/stable/components/pricing/pricing-selector-component.mjs"
+import "https://widgets.risevision.com/stable/components/pricing/pricing-summary-component.mjs"
+import "https://widgets.risevision.com/stable/components/pricing/pricing-grid-component.mjs"
```
olegrise: @tejohnso The webpage loads faster when it requests one file instead of multiple files. See [here](https://developers.google.com/web/fundamentals/performance/get-started/httprequests-5). I'm not sure what was the benefit for creating multiple repositories at the first place, but regardless, is there a way to merge those files into one? The Apps is already making way too many requests and load time is not ideal.

tejohnso: The pro/con discussion isn't so clear these days with HTTP/2. However I just realized we're not using it. 

I'm working on a build process to integrate into Gulp which would bundle these similar to how we bundle the other dependencies. Might not be ready for launch.

olegrise: Sounds good. Thanks.

tejohnso: See commit#971f177

andrecardoso:
@tejohnso Nice you added rollup and babel 👍, but I'm curious any reason to not use webpack?

tejohnso:
> @tejohnso Nice you added rollup and babel , but I'm curious any reason to not use webpack?

Had better experience with rollup, plus lit-element recommended it.

web/pricing-component.mjs
olegrise: 👍

web/index.html
```diff
@@ -1,6 +1,8 @@
 <!DOCTYPE html>
 <html ng-app="risevision.apps" itemscope itemtype="http://schema.org/WebApplication" prefix="og: http://ogp.me/ns# fb: http://ogp.me/ns/fb#">
 <head lang="en">
+  <script src="https://unpkg.com/@webcomponents/webcomponentsjs@2.2.10/webcomponents-loader.js"></script>
+  <script src="/pricing-component.js"></script>
```
olegrise: @tejohnso can this be moved to [line 69](https://github.com/Rise-Vision/rise-vision-apps/pull/1056/files#diff-d23fe9001e474c96419c8f2519cbef67R69) so it's included into `website.min.js`?

tejohnso: Okay. See latest commit.

olegrise: @tejohnso I've noticed that when I purchase a Plan for a sub-company that is not in Education industry, I get the discount. What is the requirement for discount - is it based on the industry of the Bill To company or Ship-To company?

olegrise:
@tejohnso I've noticed that when I purchase a Plan for a sub-company that is not in Education industry, I get the discount. What is the requirement for discount - is it based on the industry of the Bill To company or Ship-To company?

tejohnso:
> @tejohnso I've noticed that when I purchase a Plan for a sub-company that is not in Education industry, I get the discount. What is the requirement for discount - is it based on the industry of the Bill To company or Ship-To company?

Updated with https://github.com/Rise-Vision/store-server/pull/1025

PR #1092, opened 2019-07-04
Hide PUD in Schedule for HTML Template

olegrise:
@andrecardoso please review

### DIFF
--- a/test/unit/schedules/controllers/ctr-playlist-item-modal.tests.js
+++ b/test/unit/schedules/controllers/ctr-playlist-item-modal.tests.js
@@ -44,18 +44,14 @@ describe('controller: Playlist Item Modal', function() {
     $provide.service('presentation',function(){
       return {
         get: function(){
-          var deferred = Q.defer();
-          deferred.resolve({item: {productCode: '123'}});
-          return deferred.promise
+          return Q.resolve({item: {productCode: '123'}});
         }
       };
     });
     $provide.service('template',function(){
       return {
         loadBlueprintData: function(){
-          var deferred = Q.defer();
-          deferred.resolve(blueprintData);
-          return deferred.promise
+          return Q.resolve(blueprintData);
         }
       };
     });
--- a/test/unit/schedules/controllers/ctr-playlist-item-modal.tests.js
+++ b/test/unit/schedules/controllers/ctr-playlist-item-modal.tests.js
@@ -6,7 +6,7 @@ describe('controller: Playlist Item Modal', function() {
       return {
         open : function(){
           var deferred = Q.defer();
-          deferred.resolve(['presentationId', 'name']);
+          deferred.resolve(['presentationId', 'name', presentationType]);
 
           return {
             result: deferred.promise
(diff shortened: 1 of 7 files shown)


test/unit/schedules/controllers/ctr-playlist-item-modal.tests.js
```diff
@@ -40,17 +40,46 @@ describe('controller: Playlist Item Modal', function() {
           return 'companyId';
         }
       };
-    })
+    });
+    $provide.service('presentation',function(){
+      return {
+        get: function(){
+          var deferred = Q.defer();
```
andrecardoso: I think you can do simply `return Q.resolve({item: {productCode: '123'}});` here

test/unit/schedules/controllers/ctr-playlist-item-modal.tests.js
```diff
@@ -40,17 +40,46 @@ describe('controller: Playlist Item Modal', function() {
...
+          deferred.resolve({item: {productCode: '123'}});
+          return deferred.promise
+        }
+      };
+    });
+    $provide.service('template',function(){
+      return {
+        loadBlueprintData: function(){
+          var deferred = Q.defer();
```
andrecardoso: Same as above

web/scripts/schedules/controllers/ctr-playlist-item-modal.js
```diff
@@ -21,9 +30,39 @@ angular.module('risevision.schedules.controllers')
...
         modalInstance.result.then(function (presentationDetails) {
           $scope.playlistItem.objectReference = presentationDetails[0];
+          $scope.playlistItem.presentationType = presentationDetails[2];
+          configurePlayUntilDone();
         });
       };
 
+      function configurePlayUntilDone() {
+
+        $scope.playUntilDoneSupported = true;
```
andrecardoso: If the blueprint request fails should we assume that `$scope.playUntilDoneSupported` is true? (non-rhetorical question, I don't know the answer and want to discuss the options)

test/unit/schedules/controllers/ctr-playlist-item-modal.tests.js
olegrise: @andrecardoso I tried that - didn't work.

web/scripts/schedules/controllers/ctr-playlist-item-modal.js
olegrise: @andrecardoso I think it's better than setting `$scope.playUntilDoneSupported` to `false`. I was thinking about different options like show user an error and decided that in this case we can simply assume `$scope.playUntilDoneSupported = true;`. Let me know if you disagree.

test/unit/schedules/controllers/ctr-playlist-item-modal.tests.js
andrecardoso: Weird, but OK. It might be some incompatibility between Q and angular's $q. Thanks for trying.

web/scripts/schedules/controllers/ctr-playlist-item-modal.js
andrecardoso: Isn't it safer to assume it as false? If we assume as true, is the display going to be stuck if the  template doesn't support PUD?

test/unit/schedules/controllers/ctr-playlist-item-modal.tests.js
olegrise: @andrecardoso sorry, you are right, it works. I'll make a change.

olegrise: Here is what I tried before `return Q.defer().resolve()`

web/scripts/schedules/controllers/ctr-playlist-item-modal.js
andrecardoso: Also, if it fails, how are we going to fix the schedule record? Will we need to rely on the user to open the modal again and save it?

PR #1100, opened 2019-07-10
Display Address validation

ezequielc:
Validates Display address when not using Company address.
[stage-1]

### DIFF
--- a/web/scripts/displays/services/svc-display-factory.js
+++ b/web/scripts/displays/services/svc-display-factory.js
@@ -2,9 +2,10 @@
 
 angular.module('risevision.displays.services')
   .factory('displayFactory', ['$rootScope', '$q', '$state', '$modal', '$loading', '$log',
-    'display', 'displayTracker', 'playerLicenseFactory', 'processErrorCode',
+    'display', 'displayTracker', 'playerLicenseFactory', 'processErrorCode', 'addressFactory',
+    'humanReadableError',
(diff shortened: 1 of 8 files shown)


web/scripts/displays/services/svc-display-factory.js
```diff
@@ -126,6 +127,14 @@ angular.module('risevision.displays.services')
         return deferred.promise;
       };
 
+      var _validateAddress = function () {
+        if (factory.display.useCompanyAddress) {
+          return $q.resolve();
+        } else {
+          return addressFactory.isValidOrEmptyAddress(factory.display);
```
alex-deaconu: @ezequielc should we allow empty addresses for Display? To me, if the user un-checks "Use Company Address" they should probably enter an address. Just a though. Thanks.

alex-deaconu:
@ezequielc looks good, a comment inline. Thanks!

web/scripts/displays/services/svc-display-factory.js
ezequielc: @alex-deaconu Very good point. I will update and not allow blank address. Thanks

ezequielc: @alex-deaconu Updated as suggested, please review the last commits. Thanks!

alex-deaconu:
@ezequielc looks good 👍 . Thanks for updating!

Issue #1085, opened 2019-06-28
You can add an HTML Template to an Embedded Presentation item


sheadarlison:
@ezequielc what do you think of this? https://www.screencast.com/t/wNeBbqN9

ezequielc:
@sheadarlison Agreed, Embedded Presentation functionality can't handle HTML Templates so hiding from the list makes sense to me.
@santiagonoguez @fjvallarino fyi. Thanks

alex-deaconu:
@sheadarlison this issue should now be resolved. To confirm, we are filtering out HTML Templates from the Presentations list when selecting an Embedded Presentation. Thanks.

PR #1104, opened 2019-07-11
Exclude HTML Templates from Embedded Presentation add

alex-deaconu:


alex-deaconu:
I struggled to find a proper way to pass the fact that we want to query out HTML Templates to `ctr-presentation-list.js`. It is used for the Presentation list, Schedule selector and a bunch of other places. Passing a Parameter was not possible since sometimes the controller is initialized through `ng-controller` in the partials.

Recommended way to pass variables to Controllers is through services. Using the variable is like a basic service with 1 property.

Using `$timeout` to reset works because the code gets executed after the `$digest` cycle which opens the modal. It is the recommended way to wait for a cycle to end.

alex-deaconu:
@ezequielc that is the exact issue I was facing, transferring data from one controller (ie. `PresentationSelectorModal`) to the `PresentationListController`. The solution I came up with, although not ideal, was the shared "service" constant. I will think about it more and see if I can come up with something better. Thanks.

alex-deaconu:
I was not able to find a more optimal way to pass the additional filter for the Presentations list than what is implemented here. @andrecardoso @olegrise wondering if you could take a look at this PR and assess if the implementation is acceptable. Thanks!

andrecardoso:
@alex-deaconu it is an unusual way to use a constant, but I'm good with it.

alex-deaconu:
@andrecardoso as discussed, I've added a fix for the Display Address validation failure in e2e tests. With that occasion I'm actually testing the functionality to some extent as well. Thanks.

alex-deaconu:
Fixes #1085 

Release checklist:
Validated the fix in the Beta environment to ensure list filters work correctly.

Unit tests coverage has been added for the functionality.

Will test the released functionality in Production once deployed.

In the eventuality of an issue, we will have to revert the PR as per Apps build process.

No documentation updates are needed. PR description provides information regarding the implementation details.

Will inform Support of the change going out after merge.

Thanks.

olegrise:
@alex-deaconu looks good. Better late than never 😉

Issue #1063, opened 2019-06-21
Images with special characters in names or paths not showing in editor preview or player


santiagonoguez:
To reproduce use this company:
https://apps.risevision.com/editor/list?cid=7fa5ee92-7deb-450b-a8d5-e5ed648c575f

I created the following images and folders in storage:
- apache (" !"#$%&'*+,-.:;<=>?@[\]^_`{|}~)|.jpg
- special (" !"#$%&'*+,-.:;<=>?@[\]^_`{|}~)

If in rise-image settings I pick that first image or images in that second directory, they don't show in the preview editor, or in the display after publishing.

stulees:
This is now fixed by deploying #1143 and rise-image changes in https://github.com/Rise-Vision/rise-image/pull/34

PR #1106, opened 2019-07-12
Save Image files value as Array and support backwards compatibility of String value

stulees:
- Fix for issue #1063 
- `files` will be an Array data type in the component, see https://github.com/Rise-Vision/rise-image/pull/24 for those changes. This is to allow file path values to include any special character, i.e. `|`
- Changes in this PR support legacy implementation of String `files` value with | separator. The legacy value is converted to Array for saving to attribute data. 

Validated with a staged version of the component in _example-financial-template_.  Tested both old `files` | based value and also having revised JSON Array value. Also tested a file with all special characters in name and also files residing in a folder with all special characters in name  - https://apps-stage-8.risevision.com/templates/edit/55d88832-c75d-4512-b165-50773322e5c4/?cid=b428b4e8-c8b9-41d5-8a10-b4193c789443

### DIFF
--- a/test/unit/template-editor/components/directives/dtv-component-image.tests.js
+++ b/test/unit/template-editor/components/directives/dtv-component-image.tests.js
@@ -2,9 +2,9 @@
 
 describe('directive: TemplateComponentImage', function() {
   var $scope,
-      element,
-      factory,
-      timeout;
+    element,
+    factory,
+    timeout;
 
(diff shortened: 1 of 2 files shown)


stulees:
@fjvallarino e2e tests have passed, this is good for you to review now. Please also review https://github.com/Rise-Vision/rise-image/pull/24 , thanks!

Issue #1088, opened 2019-07-01
Parenthesis in Display or Schedule search causes it to fail.


colby-maughan:
### What is the expected behavior?

Schedules or displays with a () in their name should load. 

### What is the actual behavior?

When searching for either, an error about schedules or displays failing to load appears. It suggests to reload the page, but as soon as you remove the parenthesis it loads correctly. 

### What steps can be taken to replicate the problem?

Search for something like Test(Display) or Test(Schedule) on even just a single ( or ). This doesn't appear to be an issue with presentations.

alex-deaconu:
@colby-maughan this issue should now be fixed. To confirm, it's been fixed for some other lists as well as Displays/Schedules. Thanks.

PR #1111, opened 2019-07-17
Fix search terms, should use double quotes

alex-deaconu:
[stage-2]

### DIFF
--- a/test/e2e/template-editor/cases/template-editor-add.js
+++ b/test/e2e/template-editor/cases/template-editor-add.js
@@ -19,7 +19,7 @@ var TemplateAddScenarios = function() {
       templateEditorPage = new TemplateEditorPage();
 
       presentationsListPage.loadCurrentCompanyPresentationList();
-      presentationsListPage.createNewPresentationFromTemplate('"Example Financial Template V3"', 'example-financial-template-v3');
+      presentationsListPage.createNewPresentationFromTemplate('Example Financial Template V3', 'example-financial-template-v3');
       templateEditorPage.dismissFinancialDataLicenseMessage();
     });
 
--- a/web/scripts/launcher/directives/dtv-weekly-templates.js
+++ b/web/scripts/launcher/directives/dtv-weekly-templates.js
@@ -18,7 +18,7 @@ angular.module('risevision.apps.launcher.directives')
 
           $scope.search = {
             // sortBy: 'templateReleaseDate DESC',
-            query: 'templateOfTheWeek:1',
+            filter: 'templateOfTheWeek:1',
             category: 'Templates',
             count: 4
           };
--- a/web/scripts/schedules/services/svc-schedule.js
+++ b/web/scripts/schedules/services/svc-schedule.js
@@ -22,7 +22,7 @@ angular.module('risevision.schedules.services')
         var query = '';
 
         for (var i in fields) {
-          query += 'OR ' + fields[i] + ':~\'' + search + '\' ';
+          query += 'OR ' + fields[i] + ':~\"' + search + '\" ';
         }
 
         query = query.substring(3);
--- a/bower.json
+++ b/bower.json
@@ -15,7 +15,11 @@
     "Sortable": "^1.4.2",
     "angular-truncate": "*",
     "angular-slugify": "^1.0.1",
+<<<<<<< HEAD
     "common-header": "https://github.com/Rise-Vision/common-header.git#fix/search",
+=======
+    "common-header": "https://github.com/Rise-Vision/common-header.git#v3.9.7",
+>>>>>>> master
     "oauthio-web": "0.6.2",
     "stretchy": "https://github.com/Rise-Vision/stretchy.git#gh-pages"
   },
--- a/bower.json
+++ b/bower.json
@@ -15,11 +15,7 @@
     "Sortable": "^1.4.2",
     "angular-truncate": "*",
     "angular-slugify": "^1.0.1",
-<<<<<<< HEAD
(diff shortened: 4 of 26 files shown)


alex-deaconu:
Fix for https://github.com/Rise-Vision/rise-vision-apps/issues/1088

As per https://cloud.google.com/appengine/docs/standard/java/search/query_strings and my research, single quotes `'` are not correctly interpreted by the search API.

Had to make some adjustments to the e2e tests and also the templates filter, but overall I think using quotes for the search terms (in `svc-store.js`) is an improvement.

@fjvallarino @olegrise please review. Thanks!

test/e2e/template-editor/cases/template-editor-add.js
```diff
@@ -19,7 +19,7 @@ var TemplateAddScenarios = function() {
       templateEditorPage = new TemplateEditorPage();
 
       presentationsListPage.loadCurrentCompanyPresentationList();
-      presentationsListPage.createNewPresentationFromTemplate('"Example Financial Template V3"', 'example-financial-template-v3');
+      presentationsListPage.createNewPresentationFromTemplate('Example Financial Template V3', 'example-financial-template-v3');
```
olegrise: @alex-deaconu was there a problem with original implementation? The quotes around the name were used specifically to get the exact search.

alex-deaconu: @olegrise the quotes were added to account for the fact that the search was innacurate (it was missing quotes https://github.com/Rise-Vision/rise-vision-apps/blob/master/web/scripts/editor/services/svc-store.js#L86). This PR fixes that by adding quotes around the search term, so this workaround is no longer needed. Thanks.

web/scripts/launcher/directives/dtv-weekly-templates.js
```diff
@@ -18,7 +18,7 @@ angular.module('risevision.apps.launcher.directives')
 
           $scope.search = {
             // sortBy: 'templateReleaseDate DESC',
-            query: 'templateOfTheWeek:1',
+            filter: 'templateOfTheWeek:1',
```
olegrise: @alex-deaconu are you sure? The `$scope.search` is passed to the `productsFactory.loadProducts`, and then [loadProducts](https://github.com/Rise-Vision/rise-vision-apps/blob/af5185332334aabaa1d6e8f56db1556de61a6208/web/scripts/editor/services/svc-products-factory.js#L39) calls `search.query`

olegrise: @alex-deaconu never mind. It seems that with your change it's working correctly. Thanks.

alex-deaconu: @olegrise I don't follow. This filter is for the search api, to filter out `templateOfTheWeek` where value is `1`. The query is doing something else, searching based on a search term.

The line you are quoting is used for other purposes and it is correctly using the query to separate out the results of the unlisted products (long story).

Basically, there's no point to query `templateOfTheWeek:1` in there, for unlisted products. First off, it wouldn't work. Second, unlisted products is specific to Widgets/Content, not Templates.

Either way, to conclude here are some sample search strings:
before
`(visibleTo:ALL OR visibleTo:b428b4e8-c8b9-41d5-8a10-b4193c789443) AND (productTag: Templates OR productTag: HTMLTemplates) AND test template search`

`(visibleTo:ALL OR visibleTo:b428b4e8-c8b9-41d5-8a10-b4193c789443) AND (productTag: Templates OR productTag: HTMLTemplates) AND templateOfTheWeek:1`

after:
`(visibleTo:ALL OR visibleTo:b428b4e8-c8b9-41d5-8a10-b4193c789443) AND (productTag: Templates OR productTag: HTMLTemplates) AND "test template search"`

`(visibleTo:ALL OR visibleTo:b428b4e8-c8b9-41d5-8a10-b4193c789443) AND (productTag: Templates OR productTag: HTMLTemplates) AND templateOfTheWeek:1`

and potentially
`(visibleTo:ALL OR visibleTo:b428b4e8-c8b9-41d5-8a10-b4193c789443) AND (productTag: Templates OR productTag: HTMLTemplates) AND "test template search" AND templateOfTheWeek:1`

Thanks.

olegrise: @alex-deaconu sounds good. Sorry for the confusion. Thanks.

web/scripts/schedules/services/svc-schedule.js
```diff
@@ -22,7 +22,7 @@ angular.module('risevision.schedules.services')
         var query = '';
 
         for (var i in fields) {
-          query += 'OR ' + fields[i] + ':~\'' + search + '\' ';
+          query += 'OR ' + fields[i] + ':~\"' + search + '\" ';
```
fjvallarino: @alex-deaconu not really important, but I don't think we need to escape the `"` since they are inside `'`

test/e2e/template-editor/cases/template-editor-add.js
olegrise: @alex-deaconu I see. Thanks.

web/scripts/schedules/services/svc-schedule.js
alex-deaconu: @fjvallarino I was wondering the same thing, but I kept it like that for consistency with the *only working* list search https://github.com/Rise-Vision/rise-vision-apps/blob/master/web/scripts/editor/services/svc-presentation.js#L22.

Issue #1073, opened 2019-06-24
Several presentations were created when editing presentation title


santiagonoguez:
A weird behavior presented right after creating a presentation on Samsung Galaxy S8 on browserstack.

Context: the simulation was very slow, and several network errors presented during the test, so it seems that the network of the remote test phone was not very good.

After creating a new presentation of the new QA template, the phone opened it but showed a dialog saying it could not be saved ( network error apparently ).

But as the presentation could open, I tried to edit the presentation title. The editor was presenting weird editing behavior, maybe related to the slow interaction with the remote phone, and/or to https://github.com/Rise-Vision/rise-vision-apps/issues/928

The real crappy behavior was that as a result of trying to edit this, several new presentations were created, apparently by auto-save ( trying to insert the presentation as if it were new, it seems ). Look capture below:

![multiple-creation](https://user-images.githubusercontent.com/33162690/60036996-1db1a700-9676-11e9-8fda-e0d85ecdea72.png)

santiagonoguez:
I think the real problem here was that the initial save failed, so every subsequent auto-save did not see a presentation id, and did an insertion instead of an update.

andrecardoso:
I couldn't reproduce this issue, neither Francisco (see https://github.com/Rise-Vision/rise-vision-apps/pull/1115#issuecomment-512570617). Closing it.

PR #1115, opened 2019-07-17
Use updatePresentation instead of save on auto-save

andrecardoso:
- If the presentation id is null auto-save will create a new presentation on every change
- Fix to https://github.com/Rise-Vision/rise-vision-apps/issues/1073

### DIFF
--- a/test/unit/template-editor/controllers/ctr-template-editor.tests.js
+++ b/test/unit/template-editor/controllers/ctr-template-editor.tests.js
@@ -38,6 +38,9 @@ describe('controller: TemplateEditor', function() {
       save: function() {
         return Q.resolve();
       },
+      updatePresentation: function() {
+        return Q.resolve();
+      },
       publishPresentation: function () {
         return Q.resolve();
       }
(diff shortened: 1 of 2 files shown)


andrecardoso:
[[Issue #1073] Several presentations were created when editing presentation title [3]](https://trello.com/c/mXUqug2Z/166-issue-1073-several-presentations-were-created-when-editing-presentation-title-3)

fjvallarino:
@andrecardoso I'm not sure about this. I was never able to reproduce the issue and I understand you were not able either. Looking at the Core, when updating a Presentation the logic is basically the same as when adding (it calls `put` after parsing what the client sent):

- Add: https://github.com/Rise-Vision/core/blob/master/src/main/java/com/risevision/core/api/v1/accessors/PresentationAccessor.java#L188
- Update: https://github.com/Rise-Vision/core/blob/master/src/main/java/com/risevision/core/api/v1/accessors/PresentationAccessor.java#L224

I think if the issue happens again, we will face the same problem with this approach.

I personally think we should discuss if attempting to fix this issue makes sense, considering we were never able to reproduce it (which means we can't validate if our fix corrects the problem)

andrecardoso:
@fjvallarino you are right. I switched back to using `save` and have similar results in my tests. I don't think we need to merge this and as both you and me couldn't reproduce this issue I think we should close it.

fjvallarino:
@andrecardoso agreed. I'll add my thoughts on the issue too

PR #1116, opened 2019-07-18
Update phone number for user rather than company object

alex-deaconu:
[stage-2]


alex-deaconu:
@andrecardoso please review. Thanks!

Issue #1080, opened 2019-06-27
Presentation Editor shows blank bar at the bottom of the page


ezequielc:
As the plans bar was recently removed, calculation for the Editor to fit the entire screen is incorrect.

*What should happen*

Editor should fill the entire window.

ezequielc:
The same happens for Template Editor. Thanks

alex-deaconu:
This issue is now fixed. Thanks.

PR #1118, opened 2019-07-18
Use directive that checks and sets CH height

alex-deaconu:
Sets height based on the actual value
Updates on Company update/change and Plan change

Use `$scope.$on` to prevent event handler from being active after
directive is removed

[stage-2]

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -15,7 +15,7 @@
     "Sortable": "^1.4.2",
     "angular-truncate": "*",
     "angular-slugify": "^1.0.1",
-    "common-header": "https://github.com/Rise-Vision/common-header.git#fix/editor-height",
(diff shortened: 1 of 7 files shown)


alex-deaconu:
Fix for https://github.com/Rise-Vision/rise-vision-apps/issues/1080

alex-deaconu:
Instead of using states to determine the height of the common-header arbitrarily, this directive checks the actual rendered height and updates the CSS variable.

Works for both the Template and regular Editor.

Second fix (using `$scope.$on`) is for an issue I found https://trello.com/c/nUZ2byrs/5119-issue-updating-company-settings-outside-the-template-editor-after-having-opened-the-template-editor-before-causes-a-postmessage.

@fjvallarino @andrecardoso please review. Thanks!

Issue #928, opened 2019-04-30
Template Editor: Template name shifts out of view when it is long 


sheadarlison:
Screenshot of issue happening: https://www.screencast.com/t/5cwN1Yk2wgFQ

Template where it occurs: https://apps.risevision.com/templates/edit/cb997e91-0e21-4081-87f7-b93a6ce90fa4?cid=7fa5ee92-7deb-450b-a8d5-e5ed648c575f

How it should work: template name should appear correctly aligned, [screenshot](https://app.zeplin.io/project/5c58a8a2b74a443081a59e09/screen/5ca038ecc2d2d2052d7590f3)

santiagonoguez:
And in Safari it looks even worse !

santiagonoguez:
Also iPhone X Chrome looks very bad

santiagonoguez:
Also Samsung Chrome sometimes looks very bad

alex-deaconu:
@sheadarlison @santiagonoguez I've addressed some of the issues listed in this PR with 3 fixes in Apps and the Stretchy library. I was not able to replicate any more issues, but if we do discover any we can reopen or log a more specific issues. Closing. Thanks!

santiagonoguez:
@alex-deaconu thanks ! :D

PR #1122, opened 2019-07-19
Fix various issues with Stretchy

alex-deaconu:
When name is using ellipsis, the scrollLeft bulletproof
from Stretchy doesn't work. It shifts the contents of the
input field out of view and calculates a much larger width.
Fixed by PR from the forked repo.

In Safari, stretchy initializes the element before the
text field value is updated. This causes a very short input to
be rendered. Fixing this by forcing stretchy initialization after
directive is rendered.

defaultNameWidth variable is initialized incorrectly. Using element.style
does not obtain the correct width. Using getComputedStyle instead.

[stage-2]

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -17,7 +17,7 @@
     "angular-slugify": "^1.0.1",
     "common-header": "https://github.com/Rise-Vision/common-header.git#v3.9.11",
     "oauthio-web": "0.6.2",
-    "stretchy": "https://github.com/Rise-Vision/stretchy.git#fix/elipsis"
+    "stretchy": "https://github.com/Rise-Vision/stretchy.git#gh-pages"
   },
   "devDependencies": {
     "q": "1.0.1",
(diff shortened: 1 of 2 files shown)


alex-deaconu:
This PR addresses some of the issues I found when investigating https://github.com/Rise-Vision/rise-vision-apps/issues/928. I've created videos to demo the behaviour of each of the issues I fixed.

Fix #1 in Stretchy: https://screencast.com/t/c8YXqyMWKhL

Fix #2 for Safari: https://screencast.com/t/RA9k02XwBYgs

Fix #3 for element.style: https://screencast.com/t/Xqhu3YvjDoa

@stulees @fjvallarino wondering if you could review as you're most familiar with this functionality. Thanks!

alex-deaconu:
As per release checklist:
I've tested the changes on various resolutions in Chrome Desktop, Safari Mac, see various demo videos.

There are no automated tests for this directive so adding new tests for this functionality would require quite an extensive effort.

I will validate the release once in Production to ensure everything works as expected.

For rollback, we have to rely on the Apps build to pass. This is not ideal but this is not a high impact release.

The PR comments should contain enough descriptions of the fixes applied.

I will let Support know of the release once merged.

Thanks.

PR #1142, opened 2019-07-31
Fix/issue 29 image refresh in preview

santiagonoguez:
## Description
Includes the 'time-created' entry in the metadata, so preview can be updated when a file is replaced.

see https://github.com/Rise-Vision/rise-image/pull/33
for test and release info.

### DIFF
--- a/web/scripts/template-editor/components/directives/dtv-component-image.js
+++ b/web/scripts/template-editor/components/directives/dtv-component-image.js
@@ -68,6 +68,7 @@ angular.module('risevision.template-editor.directives')
             var newFile = {
               file: filePath,
               exists: true,
+              'time-created': _timeCreatedFor(item),
               'thumbnail-url': _thumbnailFor(file)
             };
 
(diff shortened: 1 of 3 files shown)


web/scripts/template-editor/components/directives/dtv-component-image.js
```diff
@@ -82,12 +84,16 @@ angular.module('risevision.template-editor.directives')
...
             if (item.metadata && item.metadata.thumbnail) {
-              return item.metadata.thumbnail + '?_=' + (item.timeCreated && item.timeCreated.value);
+              return item.metadata.thumbnail + '?_=' + _timeCreatedFor(item);
             } else {
               return DEFAULT_IMAGE_THUMBNAIL;
             }
           }
 
+          function _timeCreatedFor(item) {
+            return item.timeCreated && item.timeCreated.value;
```
fjvallarino: @santiagonoguez for previously uploaded images this will return `undefined` and appended to the url, right? It should work fine, I'm just confirming

fjvallarino: @santiagonoguez disregard, this will be updated every time and the component relies on the `thumbnail-url` we generate here

santiagonoguez: :D

PR #1188, opened 2019-08-16
Load/save RSS attribute data and validation of RSS feed

stulees:
## Description
Loading and saving of attribute data and populating UI

Validation of RSS feed - URL format, checking if feed parser can parse it (i.e. not an actual RSS feed), and checking if valid RSS (via rss [feed validator service](https://validator.w3.org/feed/check.cgi?url=))

## Motivation and Context
Required as part of our RSS component goal, so users can edit the RSS feed and customize max items. 

## How Has This Been Tested?
Tested using https://apps-stage-8.risevision.com/templates/edit/1aa312a7-410c-4317-8723-6a0092e30c45/?cid=b428b4e8-c8b9-41d5-8a10-b4193c789443

## Release Plan:
  - Manually tested
  - Automated unit tests included
  - Will validate in production upon deployment
  - Given no users are using RSS in templates, rollback unlikely, however if needed it will require re-running previous master deployment and reverting code changes

- Support and documentation skipped


ezequielc:
@stulees Re. @fjvallarino comment on how to avoid `xmlToJSON.js`, what about using our feed-parser as the validator? We could simply make a request to it and in case of error, we can assume that is it not a valid feed. It also has the benefit of really validating if feed-parser supports it. Thanks!

stulees:
@ezequielc This is already occurring. These settings first validate the url format, upon being valid then a request to feed-parser is made to validate feed-parser is able to parse the feed, upon that being valid then a request is made to the feed validator service to check for rss validation errors. This last step is where `xmlToJSON` is required because the response from the service is XML and we need to check the `validity` node.

ezequielc:
@stulees Ok, I am good in that case. Thanks!

Issue #1167, opened 2019-08-09
[Image Component] [Rise Storage] Buttons when there are no images in Rise Storage


jasoncaldwell:
**Defect:** When there isn’t any content in Rise Storage, the “Add Selected” button shouldn’t be visible:

![image26](https://user-images.githubusercontent.com/1440943/62787523-80c29600-ba8a-11e9-8961-a97b0abe57e9.jpg)

santiagonoguez:
@jasoncaldwell this has been fixed [here](https://github.com/Rise-Vision/rise-vision-apps/pull/1195).

You can validate it working in any presentation, such as these:

https://apps.risevision.com/templates/edit/b5df98d6-b534-4d62-99f1-b899dde2949d/?cid=7fa5ee92-7deb-450b-a8d5-e5ed648c575f

Please confirm we can close this. thanks !

jasoncaldwell:
@santiagonoguez Looks good!

PR #1195, opened 2019-08-19
Fix/issue 1167 hide add selected button on empty storage

santiagonoguez:
## Description
Hide 'add selected' button when there are no files that can be selected in a folder.

## Motivation and Context
See issue 1167 https://github.com/Rise-Vision/rise-vision-apps/issues/1167

## How Has This Been Tested?
See https://github.com/Rise-Vision/common-header/pull/1038
for this part and the release checklist.

### DIFF
--- a/web/scripts/template-editor/services/svc-template-editor-utils.js
+++ b/web/scripts/template-editor/services/svc-template-editor-utils.js
@@ -64,6 +64,14 @@ angular.module('risevision.template-editor.services')
         });
       };
 
+      svc.hasRegularFileItems = function (folderItems) {
+        var regularFiles = _.filter(folderItems, function(item) {
+          return !svc.isFolder(item.name);
+        });
+
+        return regularFiles.length > 0;
+      };
+
       svc.findElement = function (selector) {
         return document.querySelector(selector) && angular.element(document.querySelector(selector));
       };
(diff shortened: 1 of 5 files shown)


santiagonoguez:
@willread updated to latest master ! :D

web/scripts/template-editor/services/svc-template-editor-utils.js
```diff
@@ -64,6 +64,14 @@ angular.module('risevision.template-editor.services')
         });
       };
 
+      svc.hasRegularFileItems = function (folderItems) {
```
stulees: @santiagonoguez This is minor, but just wondering about this function name, why _Regular_ is in the name? Is this to differentiate it from something else?

santiagonoguez: @stulees yes, it's to differentiate regular files from folders ( selectedFiles in contrast contains both folders and regular files ). As with HTML templates we still don't support folder selection, showing the 'add selected' button when there are only folders and no regular files makes no sense. Thanks !

stulees: @santiagonoguez Okay thanks for clarifying

PR #1205, opened 2019-08-21
Request user confirmation before removing logo

ezequielc:
## Description
Request user confirmation before removing logo

## Motivation and Context
JTBD from Branding epic.

## How Has This Been Tested?
Manually tested removing a logo and cancel/confirm clicks.
Also confirm the modal doesn't show for regular images.

https://apps-stage-1.risevision.com

## Release Plan:

**Note**: The button color does not match mockup due to a recent styling change that introduced this defect, also present in Presentation Delete modal. Reported here: https://github.com/Rise-Vision/common-header/pull/1039#discussion_r316344775

It will get automatically fixed once styling is reverted. We are also performing QA to confirm this is not shipped to users. 


web/scripts/template-editor/components/directives/dtv-component-image.js
```diff
@@ -262,7 +262,9 @@ angular.module('risevision.template-editor.directives')
           };
 
           $scope.removeImageFromList = function (image) {
-            _setSelectedImages(imageFactory.removeImage(image, $scope.selectedImages));
+            imageFactory.canRemoveImage(image).then(function(){
+              _setSelectedImages(imageFactory.removeImage(image, $scope.selectedImages));
```
alex-deaconu: @ezequielc `imageFactory.canRemoveImages()` be part of `imageFactory.removeImage()`? ie. not add another factory function. `removeImage()` could just return the old images list if the user wants to keep the image. 

Also, does `removeImage` require the `image` parameter? I don't see it being used.

Thoughts? Thanks.

ezequielc: @alex-deaconu I was struggling to have a good solution given the asynchronous call but I like your suggestion. I will update the code.
And I agree, there is no need to pass `image` as a parameter. Thanks!

web/scripts/template-editor/components/services/svc-base-image-factory.js
```diff
@@ -31,9 +31,7 @@ angular.module('risevision.template-editor.services')
         var metadata =
           fileMetadataUtilsService.metadataWithFileRemoved(currentMetadata, image);
 
-        if (metadata) {
```
alex-deaconu: @ezequielc do we not need this check anymore? Thanks.

ezequielc: @alex-deaconu It doesn't look like this case can happen as the user will click Remove in a file from the metadata list.

I would expect the `metadataWithFileRemoved` to return the original metadata if file is not found, but it looks like there is no return value in such case. 
To be safe I will add a conditional and resolve to an empty metadata. I don't want to change the behaviour of the other method as it is being used elsewhere. Do you agree?

alex-deaconu: @ezequielc agreed, thanks!

ezequielc: @alex-deaconu Updated, please review. Thanks!

PR #1207, opened 2019-08-21
Fix/template delete confirmation style

santiagonoguez:
See https://github.com/Rise-Vision/common-header/pull/1040

### DIFF
--- a/web/partials/template-editor/components/component-image/image-list.html
+++ b/web/partials/template-editor/components/component-image/image-list.html
@@ -35,7 +35,11 @@
 
 <div class="file-component-list-action-button-bar">
   <div class="upload-files">
-    <label id="image-list-uploader-label" class="btn btn-block" for="image-list-uploader" ng-disabled="isUploading">
+    <label id="image-list-uploader-label"
+      class="btn btn-not-primary btn-block"
+      for="image-list-uploader"
+      ng-disabled="isUploading"
+    >
       <strong>Upload Images</strong>
     </label>
   </div>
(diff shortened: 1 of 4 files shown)


web/partials/template-editor/components/component-image/image-list.html
```diff
@@ -35,7 +35,11 @@
 
 <div class="file-component-list-action-button-bar">
   <div class="upload-files">
-    <label id="image-list-uploader-label" class="btn btn-block" for="image-list-uploader" ng-disabled="isUploading">
+    <label id="image-list-uploader-label"
+      class="btn btn-not-primary btn-block"
```
ezequielc: @santiagonoguez I think you should try to use `btn-default` instead of defining a new class. 
If the styles don't match, I would still use `btn-default` and apply the required changes with a more specific selector. Thoughts? Thanks!

santiagonoguez: @ezequielc sounds good, I'll do that then, thanks !

santiagonoguez: @ezequielc @stulees it's now working with btn-default, so I removed the new style I introduced. Validation also here:
https://apps-stage-9.risevision.com/templates/edit/6eba4d1b-bd19-435a-a2a9-eef3feeb47a0/?cid=cf85e5c4-9439-40ce-94d6-e5ea6ea2411c

Please review again this and https://github.com/Rise-Vision/common-header/pull/1040

PR #1290, opened 2019-10-04
Add checkout scenario and remove it from template-add

alex-deaconu:
#### Don't purchase a plan for Editor tests
Created a separate sub-company with a plan
#### Don't create subcompanies for unsubscribed tests
Use a generic subcompany instead

## Motivation and Context
Move checkout step to brand new company and separate flow
Optimize tests to reduce sub-company creations

## How Has This Been Tested?
No actual code/functionality changes in this PR
Tested e2e suite locally and on Circle CI

## Release Plan:

N/A


alex-deaconu:

test/e2e/common-header/cases/checkout.js
```diff
@@ -0,0 +1,80 @@
...
+                
+      before(function (){
+        commonHeaderPage = new CommonHeaderPage();
+        homepage = new HomePage();
+        signInPage = new SignInPage();
+        purchaseFlowModalPage = new PurchaseFlowModalPage();
+        pricingComponentModalPage = new PricingComponentModalPage();
+
+        //TODO Allow time for the trial subscription to be enabled
+        browser.sleep(30000);
```
ezequielc: @alex-deaconu We should probably add an improvement to our list to implement a retry mechanism here. It failed a few times on my branch so I added this sleep as a workaround, but it is not an ideal solution. Thanks

test/e2e/common-header/cases/checkout.js
```diff
@@ -0,0 +1,80 @@
...
+
+      it("should purchase",function() {
+        //TODO use credit card
+        purchaseFlowModalPage.purchase(false);
+
+        //TODO move this inside retry mechanism of purchase() above
+        browser.sleep(30 * 1000);       
+        browser.driver.navigate().refresh();
+        browser.sleep(10000);
+        helper.waitDisappear(commonHeaderPage.getLoader(), 'CH Spinner Loader');
```
ezequielc: @alex-deaconu Also, as we no longer use purchase elsewhere, we could change its `_waitForPlanUpdate()` to check for the `Subscribe Now` button instead of `See Our Plans` link and we can remove this code from here. Thanks!

alex-deaconu: @ezequielc good call. Updated the check to validate that the Plan Bar is not present in the Purchase function and removed this retry. Thanks.

alex-deaconu:
@ezequielc addressed on of your comments, and also switched to use a CC for the purchase (not sure if it will require further fixes). Thanks.

test/e2e/common-header/cases/checkout.js
```diff
@@ -56,16 +56,9 @@
       });
 
       it("should purchase",function() {
-        //TODO use credit card
-        purchaseFlowModalPage.purchase(false);
+        purchaseFlowModalPage.purchase();
```
ezequielc: @alex-deaconu To use credit card I think you need to specify `.purchase(true)`. But anyway, it can be addressed in another PR. Thanks!

ezequielc:
@alex-deaconu Thanks for updating the retry!  👍

test/e2e/common-header/cases/checkout.js
alex-deaconu: @ezequielc oh, good catch. Will fix in my next PR, thanks!

PR #1294, opened 2019-10-07
Apply title for counter instance based on instance 'type' attribute

stulees:
## Description
Format the title for rise-data-counter instances by constructing it to concatenate the `type` attribute value of the instance to differentiate between _up_ and _down_ for matching the common-header local values

## Motivation and Context
rise-data-counter UI settings development. We need to differentiate labels due to the component being able to be configured to operate as a _down_ type or _up_ type

## How Has This Been Tested?
Locally and on Apps staging

https://apps-stage-8.risevision.com/templates/edit/851ba81e-1113-4788-994c-9d11a1f0c9a6/?cid=b428b4e8-c8b9-41d5-8a10-b4193c789443

## Release Plan:


web/scripts/template-editor/directives/dtv-attribute-editor.js
```diff
@@ -66,6 +66,14 @@ angular.module('risevision.template-editor.directives')
               $scope.directives[component.type].iconType : '';
           };
 
+          $scope.getComponentTitle = function (component) {
+            if (component.type === 'rise-data-counter') {
+              return 'template.' + component.type + '-' + component.attributes.type.value;
+            }
```
fjvallarino: @stulees maybe we can do something similar to what we do with `onBackHandler`? This means, we can have an extra method in the object we provide to `registerDirective` (something like `getTitle`) which will be used if provided.

```
$scope.getComponentTitle = function (component) {
  if (directive.getTitle) {
    return directive.getTitle();
  }

  return 'template.' + component.type;
}
```

And the directive would have the corresponding logic for the title. To clarify, we don't need to modify the other directives to account for this method; failing to provide it will use the default one.

stulees: @fjvallarino Yes I like that idea, I'll make the change, thanks

stulees:
@fjvallarino Missed leveraging this when navigating into the instance and its a non-panel title, now updated, see last commit.

fjvallarino:
@stulees nice!

PR #1313, opened 2019-10-21
Include user phone and mailSyncEnabled during registration

tejohnso:
## Description
Include telephone and newsletter optin during registration.

## Motivation and Context
Fields from registration are not being passed to the api call.

## How Has This Been Tested?
Via staging.

### DIFF
--- a/web/scripts/common-header/services/svc-account.js
+++ b/web/scripts/common-header/services/svc-account.js
@@ -52,10 +52,10 @@
     .factory('registerAccount', ['$q', '$log',
       'addAccount', 'updateUser',
       function ($q, $log, addAccount, updateUser) {
-        return function (userFirst, userLast, companyName, companyIndustry) {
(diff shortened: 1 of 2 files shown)


web/scripts/common-header/services/svc-account.js
```diff
@@ -67,15 +67,17 @@
...
           var deferred = $q.defer();
           riseAPILoader().then(function (riseApi) {
             var request = riseApi.account.addWithDetails({
               userFirst: userFirst,
               userLast: userLast,
               companyName: companyName,
-              companyIndustry: companyIndustry
+              companyIndustry: companyIndustry,
+              mailSyncEnabled: mailSyncEnabled,
+              userPhone: userPhone
```
alexey-rise: @tejohnso to make it consistent with the Core definition I would move `userPhone` parameter to the 3rd position.

tejohnso: Updated

PR #1335, opened 2019-10-30
HTML Templates Announcement modal

ezequielc:
## Description
- Shows a modal with a video to promote HTML Templates for companies that haven't used them recently (account older than 15 days and not created an HTML Presentation in the past 30 days).
- Require user to provide  feedback to close the modal (Thumbs Up/Down) - no close button.
- Logs 'Templates Announcement Thumbs Up/Down' events to Segment and BQ.

## Motivation and Context
In-App Templates Announcement epic

## How Has This Been Tested?
Tested locally and on stage-1.

Company that shows modal: https://apps-stage-1.risevision.com/editor/list?cid=7fab085e-a773-4351-b710-a2e40da6b04e

Company that does not show modal (has recent HTML Presentation): https://apps-stage-1.risevision.com/editor/list

## Release Plan:

Given its impact on multiple e2e tests and conditionals, we decided to disable this functionality when running e2e tests to prevent them from failing eventually.

### DIFF
--- a/web/scripts/common/directives/dtv-in-app-messages.js
+++ b/web/scripts/common/directives/dtv-in-app-messages.js
@@ -1,8 +1,8 @@
 'use strict';
 
 angular.module('risevision.apps.directives')
-  .directive('inAppMessages', ['inAppMessagesFactory', 'templatesAnnouncementFactory', 'CHECK_TEMPLATES_ANNOUNCEMENT',
-    function (inAppMessagesFactory, templatesAnnouncementFactory, CHECK_TEMPLATES_ANNOUNCEMENT) {
+  .directive('inAppMessages', ['inAppMessagesFactory', 'templatesAnnouncementFactory',
+    function (inAppMessagesFactory, templatesAnnouncementFactory) {
       return {
         restrict: 'E',
         scope: {},
@@ -11,9 +11,7 @@ angular.module('risevision.apps.directives')
           $scope.inAppMessagesFactory = inAppMessagesFactory;
           inAppMessagesFactory.pickMessage();
 
-          if (CHECK_TEMPLATES_ANNOUNCEMENT === 'true') {
-            templatesAnnouncementFactory.showAnnouncementIfNeeded();
-          }
+          templatesAnnouncementFactory.showAnnouncementIfNeeded();
         }
       };
     }
--- a/web/scripts/common/directives/dtv-in-app-messages.js
+++ b/web/scripts/common/directives/dtv-in-app-messages.js
@@ -11,7 +11,7 @@ angular.module('risevision.apps.directives')
           $scope.inAppMessagesFactory = inAppMessagesFactory;
           inAppMessagesFactory.pickMessage();
 
-          templatesAnnouncementFactory.showTemplatesAnnouncementIfNeeded();
+          templatesAnnouncementFactory.showAnnouncementIfNeeded();
         }
       };
     }
--- a/web/scripts/common/directives/dtv-in-app-messages.js
+++ b/web/scripts/common/directives/dtv-in-app-messages.js
@@ -1,8 +1,8 @@
 'use strict';
 
 angular.module('risevision.apps.directives')
-  .directive('inAppMessages', ['inAppMessagesFactory',
-    function (inAppMessagesFactory) {
+  .directive('inAppMessages', ['inAppMessagesFactory', 'templatesAnnouncementFactory',
+    function (inAppMessagesFactory, templatesAnnouncementFactory) {
       return {
         restrict: 'E',
         scope: {},
@@ -10,6 +10,8 @@ angular.module('risevision.apps.directives')
         link: function ($scope) {
           $scope.inAppMessagesFactory = inAppMessagesFactory;
           inAppMessagesFactory.pickMessage();
+
+          templatesAnnouncementFactory.showTemplatesAnnouncementIfNeeded();
         }
       };
     }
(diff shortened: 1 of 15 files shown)


web/scripts/common/directives/dtv-in-app-messages.js
```diff
@@ -1,15 +1,19 @@
...
+    function (inAppMessagesFactory, templatesAnnouncementFactory, CHECK_TEMPLATES_ANNOUNCEMENT) {
       return {
         restrict: 'E',
         scope: {},
         templateUrl: 'partials/common/in-app-messages.html',
         link: function ($scope) {
           $scope.inAppMessagesFactory = inAppMessagesFactory;
           inAppMessagesFactory.pickMessage();
+
+          if (CHECK_TEMPLATES_ANNOUNCEMENT === 'true') {
```
alex-deaconu: @ezequielc I wouldn't mix this business logic in this directive. Maybe move to the service? Thanks.

web/scripts/common/services/svc-templates-announcement-factory.js
```diff
@@ -0,0 +1,77 @@
...
+        var creationDate = (company && company.creationDate) ? new Date(company.creationDate) : new Date();
+        return _daysDiff(new Date(), creationDate) > 15;
+      };
+
+      var _isDismissed = function () {
+        return localStorageService.get(dismissedKey) === true;
+      };
+
+      var _daysDiff = function (date1, date2) {
+        return Math.abs((date1 - date2) / 1000 / 60 / 60 / 24);
```
alex-deaconu: Not something to address right now, but this would be a nice util function. Thanks.

web/scripts/common/services/svc-templates-announcement-factory.js
```diff
@@ -0,0 +1,77 @@
...
+angular.module('risevision.apps.services')
+  .factory('templatesAnnouncementFactory', ['localStorageService', 'userState', 'CachedRequest', 'presentation', '$q',
+    '$templateCache', '$modal', 'segmentAnalytics', 'bigQueryLogging',
+    function (localStorageService, userState, CachedRequest, presentation, $q, $templateCache, $modal,
+      segmentAnalytics, bigQueryLogging) {
+      var dismissedKey = 'templatesAnnouncement.dismissed';
+      var factory = {};
+      var search = {
+        sortBy: 'changeDate',
+        reverse: true,
```
alex-deaconu: @ezequielc just want to confirm, order by `changeDate` and reverse `true` returns the latest HTML Template correct? Thanks.

web/scripts/common/services/svc-templates-announcement-factory.js
```diff
@@ -0,0 +1,77 @@
...
+            controller: 'TemplatesAnnouncementModalCtrl',
+            size: 'lg',
+            backdrop: 'static', //prevent closing modal by clicking outside
+            keyboard: false, //prevent closing modal by pressing escape
+          });
+          instance.result.then(function (liked) {
+            localStorageService.set(dismissedKey, true);
+
+            var eventName = 'Templates Announcement';
+            eventName += liked ? ' Thumbs Up' : ' Thumbs Down';
```
alex-deaconu: @ezequielc curious why did you choose to track both values via the name, and not have a key value pair like `Templates Announcements`, `Thumbs up`. Thanks.

web/scripts/common/directives/dtv-in-app-messages.js
ezequielc: @alex-deaconu Good point. I will move it. Thanks

web/scripts/common/services/svc-templates-announcement-factory.js
ezequielc: @alex-deaconu Correct. Thanks

web/scripts/common/services/svc-templates-announcement-factory.js
ezequielc: @alex-deaconu Mainly because it would be easier to use with our tools.
For example in FullStory, we can filter by the event name, so we could easily list all users that gave thumbs up. We don't have the same for key/value. Thanks

alex-deaconu:
@ezequielc couple of things for consideration.

If the User has no Plan, this still shows. I'm wondering if that's intentional since they can't use Templates anyways.
Do we want to show this to non-Education users?
Do we want to refresh the logic for sub-companies (probably not, but just wanted to make sure).

Thanks.

web/scripts/common/services/svc-templates-announcement-factory.js
alex-deaconu: Sounds good, thanks!

ezequielc:
@alex-deaconu About your questions:
> If the User has no Plan, this still shows. I'm wondering if that's intentional since they can't use Templates anyways.

I think we should still show it. It is an opportunity for the user to learn and eventually subscribe.

> Do we want to show this to non-Education users?

You are right, I double checked the JTBDs and it should show only for Education. I will update.
 
> Do we want to refresh the logic for sub-companies (probably not, but just wanted to make sure).

I don't think it is needed. If they used in their main company, they are aware of HTML Templates and don't need to see the message when changing to a sub-company.

Thanks

ezequielc:
@alex-deaconu Please take a look at the latest commit. Builds are failing because of Store version, but you can take a look at staged version. Thanks

alex-deaconu:
@ezequielc looks good 👍 . Thanks for the updates and clarifications!

PR #1348, opened 2019-11-06
Log in user after sign up

alex-deaconu:
## Description
Log in user after sign up

Remove need for confirmation email
Expect Auth token from signUp response
Unify code for login and signup to handle authentication
Some refactoring

Update Password length to 12 characters
But allow 4 characters or more for login (legacy)

Remove cookie usage for registration

[stage-2]

## Motivation and Context
Part of the Sign Up improvements epic.

## How Has This Been Tested?
Tested locally and in the staging environment. Validated via unit and updated e2e tests.

## Release Plan:

No


alex-deaconu:
Will need to revert Core version updates before merge.

web/scripts/components/userstate/services/svc-custom-auth-factory.js
```diff
@@ -7,69 +7,71 @@
...
-              .then(null, function (err) {
-                deferred.reject(err);
-              });
-          } else if (_state.userToken && _state.userToken.token) {
-            gapiLoader().then(function (gApi) {
+          if (_state.userToken && _state.userToken.token) {
+            return gapiLoader().then(function (gApi) {
               gApi.auth.setToken(_state.userToken.token);
 
               // TODO: Validate token?
```
ezequielc: @alex-deaconu Are you planning to add token validation here? Thanks

alex-deaconu: @ezequielc not really, that note as been there since the original implementation.

To confirm, this would mean validating and potentially requesting a new JWT token, something which we have implemented on the server side but never on the client.

Definitely not in the scope of this epic from the best of my knowledge.

Thanks.

ezequielc: @alex-deaconu Ok, I didn't realize it was there from before. Definitely not in the scope. Thanks!

ezequielc:
@alex-deaconu I found a few problems we need to address before releasing this change:

- Password Change form (User Settings) still shows 4 characters in the error message, while it is validated against 12. Both in the input error message and on a popup after submit failure.
- The message "This account is authenticated by Google. Please, use the 'Sign in with Google' button." no longer appears when I try to use a google account on custom sign in. It shows a regular password didn't match error.
- Similar thing happens on Password Reset, it doesn't let me know it is a Google account as in production. And I actually received the password reset email.
- Also, I think we should remove  "A strong password is at least 12 characters, includes uppercase/lowercase letters..." message. It appears only after we already added 12 characters so it becomes irrelevant. Alternatively, we can show it when user types the first character. 

Let me know your thoughts. Thanks!

web/partials/components/userstate/reset-password-confirm.html
```diff
@@ -20,24 +20,14 @@ <h1 class="u_remove-top">Password Confirmation</h1>
...
     <div class="form-group" ng-class="{'has-error': (forms.resetPasswordForm.$submitted && forms.resetPasswordForm.newPassword.$invalid)}" show-errors>
       <label class="control-label">New Password</label>
-      <input type="password" class="form-control" name="newPassword" id="newPassword" ng-model="credentials.newPassword" required minlength="4" focus-me="true">
+      <input type="password" class="form-control" name="newPassword" id="newPassword" ng-model="credentials.newPassword" required minlength="12" focus-me="true">
       <p class="text-danger" ng-show="forms.resetPasswordForm.$submitted && forms.resetPasswordForm.newPassword.$error.minlength">
-        New Password must be at least 4 characters long.
+        New Password must be at least 12 characters long.
       </p>
-      <p class="text-warning" ng-show="forms.resetPasswordForm.newPassword.$valid">
+      <p class="text-warning" ng-show="forms.resetPasswordForm.newPassword.$dirty">
```
alex-deaconu: @ezequielc I opted to just show the warning whenever the User starts typing. We can revise the logic and message later in the epic. Thanks.

alex-deaconu:
@ezequielc I've addressed the validation issues you mentioned. As discussed, I could not replicate the Google account issues. Wondering if you could review the last few commits as well as the staged version. Thanks.

PR #1353, opened 2019-11-08
Add Password Meter

ezequielc:
## Description
- Added Password Meter as a directive for Registration, Password Reset and Password Change forms.
- Used [zxcvbn](https://github.com/dropbox/zxcvbn) library for password scoring
- Added [oclazyload](https://oclazyload.readme.io/) to load zxcvbn only when needed, due to its file size (~800KB).
- Refactored custom sign in e2e tests, moved signup cases to registration scenario.

## Motivation and Context
Sign Up Epic

## How Has This Been Tested?
Manually tested locally and on https://apps-stage-1.risevision.com/createaccount/.
Confirmed that zxcvbn library is only loaded on demand.

## Release Plan:

@alex-deaconu Please review


test/e2e/common/cases/signin-custom.js
```diff
@@ -12,50 +11,16 @@ var SigninCustomScenarios = function() {
...
-
-      expect(signUpPage.getPasswordStrengthWarning().isDisplayed()).to.eventually.be.true;
-    });
-
-    it('should show error when trying to signup', function() {
-      signUpPage.getSignupButton().click();
-
-      helper.waitDisappear(commonHeaderPage.getLoader(), 'CH spinner loader');
-
-      expect(signUpPage.getAlreadyRegisteredError().isPresent()).to.eventually.be.true;
```
alex-deaconu: @ezequielc curious, why are you removing this test? ie. I don't see the Already Registered scenario being copied to registration.js. Thanks.

web/partials/components/password-input/password-input.html
```diff
@@ -0,0 +1,28 @@
+<div class="form-group" ng-class="{ 'has-error' : (ngModelCtrl.$invalid && ngModelCtrl.$dirty) }">
+	<label class="control-label" for="password">
```
alex-deaconu: @ezequielc weird tab length in this partial. Not sure if it's just the Github UI or you necessarily want to address. Thanks.

web/scripts/components/password-input/dtv-password-input.js
```diff
@@ -0,0 +1,53 @@
...
+            'partials/components/password-input/password-input.html'),
+          link: function (scope, element, attrs, ctrl) {
+            scope.ngModelCtrl = ctrl;
+
+            scope.$watch('ngModel', function (newValue, oldValue) {
+              if (newValue !== oldValue) {
+                scope.ngModelCtrl.$setDirty(true);
+              }
+              if (scope.isCreating) {
+                $ocLazyLoad.load(ZXCVBN_PATH).then(function () {
```
alex-deaconu: @ezequielc wondering if we should preload when the directive initializes. Thanks.

web/partials/components/password-input/password-input.html
```diff
@@ -0,0 +1,28 @@
...
+	     class="form-control"
+	     placeholder="Enter Password"
+	     ng-model="ngModel"/>
+
+	<ng-messages role="alert" for="ngModelCtrl.$error" ng-if="ngModelCtrl.$invalid && ngModelCtrl.$dirty">
+		<p ng-message="required" class="text-danger">Oops, don't leave this blank.</p>
+		<p ng-message="minlength" class="text-danger">Password needs to be {{ isCreating ? 12 : 4}} or more characters in length.</p>
+	</ng-messages>
+
+	<div id="passwordMeter" ng-if="isCreating && scorePercentage && ngModelCtrl.$dirty">
```
alex-deaconu: @ezequielc wondering if we should still show the Password Meter even in the `$invalid` state? The double error message seems confusing. Thanks.

web/partials/common-header/user-settings-modal.html
```diff
@@ -71,26 +71,12 @@ <h2 id="user-settings-label" class="modal-title">
             Current Password is not valid.
           </p>
         </div>
-        <div class="form-group"
```
alex-deaconu: @ezequielc should we add the `<password-input>` for the Current Password field above? Thanks.

test/e2e/common/cases/signin-custom.js
ezequielc: @alex-deaconu Good catch, probably a copy/paste error. I will add it. Thanks!

web/partials/components/password-input/password-input.html
ezequielc: @alex-deaconu Ok, I will double check. Thanks

web/scripts/components/password-input/dtv-password-input.js
ezequielc: @alex-deaconu That is already the case. The ngModel $watch gets triggered on directive initialization so it gets loaded right form the start. Thanks

web/partials/components/password-input/password-input.html
ezequielc: @alex-deaconu I think we should leave as is. If we hide it on $invalid, users will not benefit from the meter as they will type 12 characters before we show it and they won't see the meter hints. 
Also, if they change the password, the meter could be showing/hiding, which is not a good experience as well. 
Thanks

web/scripts/components/password-input/dtv-password-input.js
alex-deaconu: Missed that, thanks!

web/partials/components/password-input/password-input.html
alex-deaconu: Sounds good, thanks!

web/partials/common-header/user-settings-modal.html
ezequielc: @alex-deaconu I consider that but it was not straightforward because of the `currentPasswordNotValid` error so I want to work on that in a separated PR if good with you.   Thanks

alex-deaconu: Sounds good. The one thing I would mention is that the current password field's `placeholder` text is missing, so it looks inconsistent. Thanks!

ezequielc: @alex-deaconu Good call, I added placeholder for now at least. Thanks

PR #1359, opened 2019-11-13
Log uncaught errors and signup failures to BQ

alex-deaconu:
## Description
Log uncaught errors to BQ
Log errors for registration and signup

[stage-2]

## Motivation and Context
Track if users are encountering uncaught exceptions
Also log registration and signup failures to BQ

## How Has This Been Tested?
Was not actually able to replicate any uncaught errors. But tested with an error to validate BQ insert.
Tested various errors being logged.
Updated unit tests

## Release Plan:

No


alex-deaconu:

web/scripts/components/userstate/controllers/ctr-login.js
```diff
@@ -93,7 +93,7 @@ angular.module('risevision.common.components.userstate')
               if (err && err.status === 409) {
```
ezequielc: @alex-deaconu I think we should also log the duplicate email error. Thoughts? Thanks

web/scripts/common-header/dtv-common-header.js
```diff
@@ -181,6 +181,17 @@ angular.module('risevision.common.header', [
     }
   ])
 
+  .factory('$exceptionHandler', ['$log', '$injector', function ($log, $injector) {
+    return function customExceptionHandler(exception, cause, caught) {
+      // Prevents circular reference
+      // https://stackoverflow.com/questions/22332130/injecting-http-into-angular-factoryexceptionhandler-results-in-a-circular-de
+      var bigQueryLogging = $injector.get('bigQueryLogging');
+
+      bigQueryLogging.logException(exception, cause, caught);
```
ezequielc: @alex-deaconu I am wondering if we should also send the event to Segment. It might be useful in the future. Not strong on this. Thoughts? Thanks

web/scripts/components/userstate/controllers/ctr-login.js
alex-deaconu: @ezequielc interesting thought. I tend to agree as it will help us identify how many times users actually use the wrong form to login. Thanks.

web/scripts/common-header/dtv-common-header.js
alex-deaconu: @ezequielc that's a good point. I will update to reflect that. Thanks.

web/scripts/components/userstate/controllers/ctr-login.js
alex-deaconu: @ezequielc to confirm, I've updated this logging but it turns out we have an e2e test that validates the 409 error, so this will generate log entries from this. Based on this, do you think we should retain this? Thanks.

ezequielc: @alex-deaconu I think that is ok. Ideally we should be able to filter test vs production errors and events, but that is probably a future improvement. Thanks

alex-deaconu:
@ezequielc I've updated the PR as per your 2 comments. Also the last commit contains a fix for parsing/logging response errors (Core or otherwise). Wondering if you could review. Thanks!

ezequielc:
@alex-deaconu Looks good. Thanks for addressing them.

PR #1363, opened 2019-11-15
Feature/call revised user get

santiagonoguez:
## Description
Calls 'account.get' API with email parameter for new users before presenting dialog with newsletter signup.

## Motivation and Context
Profiles for new users doesn't retrieve the flag that indicates if it already has opted in for newsletter subscription. So we need to explicitly ask for that value using the updated account.get API call with a parameter.

## How Has This Been Tested?
Manually tested locally by changing the 'isSubscribed' flag of NewsletterSubscriber entities and clearing memcached.

The ICP modal was also manually tested by manipulating state and forcing it to appear.

It was also tested remotely using the same procedure here:
https://apps-stage-9.risevision.com/

Automated tests were updated to consider the new changes.

## Release Plan:
     - To be released on Monday
     - Simple rollback is needed if there are issues, this release is low risk as it does not modify user data.

No need to notify support, or to update documentation.

### DIFF
--- a/test/unit/common-header/controllers/ctr-registration-modal-spec.js
+++ b/test/unit/common-header/controllers/ctr-registration-modal-spec.js
@@ -10,12 +10,12 @@ describe("controller: registration modal", function() {
       return {
         _closed: false,
         close: function(reason) {
-          expect(reason).to.equal("success");          
+          expect(reason).to.equal("success");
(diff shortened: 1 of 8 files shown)


santiagonoguez:
@alexey-rise please review this, thanks !

PR #1369, opened 2019-11-19
Load hubspot script and call identity at registration

tejohnso:
## Description
Loads hubspot tracking script from `js.hs-scripts.com/` and then calls the identify action through the script. This is to ensure that users are created in hubspot via the script, and any previous activity from a previous risevision.com session is captured in their hubspot user data.

## Motivation and Context
Improved functionality for marketing.  See [doc](https://drive.google.com/open?id=1xGOVuslmtSXk5VZuPbZDRHPwM6AUZ_JZ4so592cJocM)

## How Has This Been Tested?
E2E + manual testing to confirm user is created in hubspot and updated with company id via backend.

## Release Plan:

### DIFF
--- a/web/scripts/components/hubspot/svc-hubspot.js
+++ b/web/scripts/components/hubspot/svc-hubspot.js
@@ -1,0 +1,35 @@
+(function (angular) {
+
+  'use strict';
+
+  angular.module('risevision.common.components.hubspot', [])
+
+    .factory('hubspot', ['$window', 'HUBSPOT_ACCOUNT',
+      function ($window, hubspotAccount) {
+        var service = {};
+        var loaded = false;
+
+        var _hsq = window._hsq = window._hsq || [];
+
+        service.loadAs = function (email) {
+          if (!loaded) {
+             _hsq.push(['identify',{ email: email }]);
+
+            var e = document.createElement('script');
+            e.type = 'text/javascript';
+            e.defer = !0;
+            e.async = !0;
+            e.id = 'hs-script-loader';
+            e.src = '//js.hs-scripts.com/' + hubspotAccount + '.js';
+            var n = document.getElementsByTagName('script')[0];
+            n.parentNode.insertBefore(e, n);
+
+            loaded = true;
+          }
+        };
+
+        return service;
+      }
+    ]);
+
+})(angular);
--- a/test/e2e/registration/cases/registration.js
+++ b/test/e2e/registration/cases/registration.js
@@ -53,6 +53,7 @@
         PASSWORD = commonHeaderPage.getPassword();
         NEW_COMPANY_NAME = commonHeaderPage.addStageSuffix("Public School");
 
+        console.log('Checking login for ' + EMAIL_ADDRESS);
         detectAndFixUserAlreadyRegistered();
 
         signUpPage.get();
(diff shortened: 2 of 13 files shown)


tejohnso:
@ezequielc this is the change for the identity call that we discussed on Thursday

web/scripts/components/hubspot/svc-hubspot.js
```diff
@@ -0,0 +1,35 @@
...
+        var _hsq = window._hsq = window._hsq || [];
+
+        service.loadAs = function (email) {
+          if (!loaded) {
+             _hsq.push(['identify',{ email: email }]);
+
+            var e = document.createElement('script');
+            e.type = 'text/javascript';
+            e.defer = !0;
+            e.async = !0;
```
santiagonoguez: curious, why put !0 instead of true ?

web/scripts/components/hubspot/svc-hubspot.js
```diff
@@ -0,0 +1,35 @@
...
+  'use strict';
+
+  angular.module('risevision.common.components.hubspot', [])
+
+    .factory('hubspot', ['$rootScope', '$window', 'userState', 'HUBSPOT_ACCOUNT',
+      function ($rootScope, $window, userState, hubspotAccount) {
+        var service = {};
+        var loaded;
+
+        var _hsq = window._hsq = window._hsq || [];
```
santiagonoguez: what does _hsq stands for ? could we use a clear name here ?

web/scripts/components/hubspot/svc-hubspot.js
tejohnso: Following the style used here `https://github.com/Rise-Vision/rise-vision-apps/blob/master/web/scripts/components/segment-analytics/svc-segment-analytics.js#L71`

web/scripts/components/hubspot/svc-hubspot.js
tejohnso: This comes from their example [here](https://github.com/Rise-Vision/rise-vision-apps/blob/master/web/scripts/components/segment-analytics/svc-segment-analytics.js#L71). The script file expects that variable name.

web/scripts/components/hubspot/svc-hubspot.js
santiagonoguez: weird, that programmer seems to code in minified mode, hahahahaha, but I'm not against that.

web/scripts/components/hubspot/svc-hubspot.js
santiagonoguez: I see, not nice to mess with window properties, but if that's what the injected script needs, then there's nothing to do about that I guess.

web/scripts/components/hubspot/svc-hubspot.js
```diff
@@ -0,0 +1,35 @@
...
+
+        service.loadAs = function (email) {
```
ezequielc: @tejohnso Just nitpicking here, but I would rename this function to `identify()` to match the Hubspot call. Thanks

tejohnso: It doesn't just make the hubspot call, it pre-configures the hsq variable with the identity call, then it **loads** the hubspot script and the script makes the call.  That's why it's called **load**As

tejohnso:
> @tejohnso Code and functionality look good. I am wondering if you could add unit tests for the `hubspot` service. Thanks!

Unit test added.  Confirms script loads and identity is configured in hsq.  Also added a check to make sure it only loads once, just to get the :100:

PR #1388, opened 2019-11-26
Add email confirmation notification

alex-deaconu:
## Description
Add email confirmation notification
Add error handling for confirmation email

Create confirm-account page
Show user confirmation or error to confirm account
Fix wording/partial for unsubscribe

Check notifications without loading list

Clean up request-confirmation-email page (no longer used)

[stage-2]

## Motivation and Context
Notify user that they need to confirm their email account when signing up with custom authentication

## How Has This Been Tested?
Validated changes in the local environment. Updated unit tests.

## Release Plan:

No


alex-deaconu:

web/partials/components/userstate/confirm-account.html
```diff
@@ -0,0 +1,10 @@
+<div ng-show="apiError" class="alert alert-danger col-xs-12 u_margin-md-top" role="alert">
+  <strong>Oops, an error occurred while trying to confirm your account.</strong>
+  {{apiError}}
+</div>
+
+<div ng-hide="apiError" class="alert alert-warning col-xs-12 u_margin-md-top">
```
ezequielc: @alex-deaconu I think it would make more sense to use `alert-info` here. Showing as yellow gives the impression something is not right. Thoughts? Thanks

ezequielc:
@alex-deaconu I think there is a small issue on account confirmation. 
After I resent email confirmation and confirmed my account, I clicked to proceed to the App and the 'email not confirmed' message still appeared. It disappeared after page reload.
I am wondering if it is because my user was already signed in when I clicked on the confirmation link? Maybe it didn't refresh the user data as I din't have to sign in again. Thanks!

web/partials/components/userstate/confirm-account.html
alex-deaconu: @ezequielc I've used the same style for the confirmation as the unsubscribe warning. But I agree, I think it should be `info`. Do you think I should change both or just here? Thanks.

alex-deaconu:
@ezequielc interesting. I will take a look. Thanks.

ezequielc:
@alex-deaconu Code looks good, apart from the issue of my previous comment. Thanks!

web/partials/components/userstate/confirm-account.html
ezequielc: @alex-deaconu I would change just here. The other one, unsubscribing from monitoring can be unintentional, so I think it makes sense to be a warning. But not strong on this, up to you. Thanks!

alex-deaconu: @ezequielc agreed. Will update just in this instance. Thanks.

alex-deaconu:
@ezequielc I've updated the class as suggested.

Also, now once the account is confirmed, the result triggers an update for that boolean field locally. It fixes the issue, and I think it's sufficient to update locally, rather than loading from the API.

Thanks.

PR #1397, opened 2019-12-03
Show onboarding for new education customers

alex-deaconu:
## Description
Add separate route for launcher onboarding
Remove auth routes from launcher
Move signin controller inline

Detect if Company should show onboarding
Determine onboarding step based on Templates, Displays presence

[stage-2]

## Motivation and Context
Initial framework for the new onboarding flow

## How Has This Been Tested?
Tested changes in the local environment. Updated unit tests.

## Release Plan:

No


alex-deaconu:

web/scripts/launcher/services/svc-company-assets-factory.js
```diff
@@ -0,0 +1,64 @@
+'use strict';
+
+angular.module('risevision.apps.launcher.services')
+  .factory('companyAssetsFactory', ['$q', 'CachedRequest', 'presentation', 'display',
```
ezequielc: @alex-deaconu I think we need to add a `$rootScope.$on('risevision.company.selectedCompanyChanged',...)` to reset this factory. Otherwise, when we change companies it will return information about the wrong company. Thanks!

web/scripts/launcher/services/svc-onboarding-factory.js
```diff
@@ -0,0 +1,68 @@
...
+        });
+      };
+
+      var _reset = function () {
+        factory.onboardingStep = undefined;
+        
+        _getOnboardingStep(true);
+      };
+
+      $rootScope.$on('risevision.company.selectedCompanyChanged', function () {
```
ezequielc: @alex-deaconu I thought more about this and I am not sure if it is a good idea to keep the state in localstorage. I am worried that it would add too much complexity.
For example, here we are resetting it on company changed, but it won't reset if I enter a new company id from the URL. Also if I open two subcompanies in different tabs, they will conflict with each other.
One way to fix is to map the state by company id, but that increases complexity.
Also, given that we are caching the companyAssetsFactory requests, I think we are ok in terms of reducing requests. Thoughts?

PR #1405, opened 2019-12-10
Save time and date formats to attribute data

stulees:
## Description
Initializes the time format and date format UI selection from attribute data, otherwise initializes them with a default selection.

Saves the time format and date format selections to attribute data. Ensures to only save time format if type is _time_ and only saves date format if type is _date_. Otherwise saves both when type is _timedate_.

## Motivation and Context
Component UI development

## How Has This Been Tested?
Tested locally and on staging with example template

https://apps-stage-8.risevision.com/templates/edit/7bba17fa-b8cf-4221-9fea-3f9270a7e2a2/?cid=b428b4e8-c8b9-41d5-8a10-b4193c789443

To test from a new instance of template, search for _Example Time and Date_

## Release Plan:

n/a

### DIFF
--- a/web/scripts/template-editor/components/directives/dtv-component-time-date.js
+++ b/web/scripts/template-editor/components/directives/dtv-component-time-date.js
@@ -8,7 +8,7 @@ angular.module('risevision.template-editor.directives')
         scope: true,
         templateUrl: 'partials/template-editor/components/component-time-date.html',
         link: function ($scope, element) {
-          var DATE_FORMATS = [ 'MMMM DD, YYYY', 'MMM DD YYYY', 'MM/DD/YYYY', 'DD/MM/YYYY' ];
+          var DATE_FORMATS = ['MMMM DD, YYYY', 'MMM DD YYYY', 'MM/DD/YYYY', 'DD/MM/YYYY'];
 
           $scope.factory = templateEditorFactory;
           $scope.dateFormats = DATE_FORMATS.map(function (format) {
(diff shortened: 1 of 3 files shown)


web/scripts/template-editor/components/directives/dtv-component-time-date.js
```diff
@@ -36,11 +35,36 @@ angular.module('risevision.template-editor.directives')
 
           $scope.load = function () {
             var type = $scope.getAvailableAttributeData($scope.componentId, 'type');
+            var timeFormat = $scope.getAvailableAttributeData($scope.componentId, 'time');
+            var dateFormat = $scope.getAvailableAttributeData($scope.componentId, 'date');
+            var timeFormatVal = timeFormat || 'Hours12';
+            var dateFormatVal = dateFormat || $scope.dateFormats[0].format;
```
santiagonoguez: @stulees just a question, will be there later considerations for cases where a designer sets invalid values for 'type', 'time', 'date' ? going back to a default maybe, logging an error.

stulees: @santiagonoguez Good point. There will be a further task to implement the timezone UI and saving of timezone. In those tasks we can introduce validating the `type`. I think if it is invalid, we will log an error. I'm not sure if we should also force a default `type` in this scenario, I think it might be better not to so the designer is aware immediately and fixes the template. Thoughts?

stulees: @santiagonoguez Also, the web component will definitely be applying validation of `type`, so that will be another indication for the designer to correct the template.

santiagonoguez: sounds good, thanks !

PR #1422, opened 2019-12-31
Persist onboarding completed in user settings

ezequielc:
## Description
Persist onboarding completed in user settings.

## Motivation and Context
Onboarding epic.

## How Has This Been Tested?
Manually tested locally and on stage-19.

## Release Plan:

### DIFF
--- a/web/scripts/launcher/services/svc-onboarding-factory.js
+++ b/web/scripts/launcher/services/svc-onboarding-factory.js
@@ -29,8 +29,9 @@ angular.module('risevision.apps.launcher.services')
       'productCode': '601ee80f3fe2950b7e4f15c57d88bf0963efb57a'
     }
   ])
-  .factory('onboardingFactory', ['$q', '$localStorage', 'userState', 'companyAssetsFactory', 'updateUser', '$rootScope', 'segmentAnalytics',
-    function ($q, $localStorage, userState, companyAssetsFactory, updateUser, $rootScope, segmentAnalytics) {
+  .factory('onboardingFactory', ['$q', 'userState', 'companyAssetsFactory', 'updateUser', '$rootScope',
+    'segmentAnalytics', '$exceptionHandler',
+    function ($q, userState, companyAssetsFactory, updateUser, $rootScope, segmentAnalytics, $exceptionHandler) {
       var factory = {};
       var onboarding = {
         currentStep: -1,
@@ -93,10 +94,6 @@ angular.module('risevision.apps.launcher.services')
         onboarding.currentStep = -1;
         onboarding.activeStep = -1;
         onboarding.tabsCompleted = {};
-
-        $localStorage.onboarding = angular.extend({
-          completed: false
-        }, $localStorage.onboarding);
       };
 
       var _setCurrentStep = function (step) {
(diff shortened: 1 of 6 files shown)


web/scripts/launcher/services/svc-onboarding-factory.js
```diff
@@ -235,22 +231,35 @@ angular.module('risevision.apps.launcher.services')
...
+        factory.loading = true;
+
+        return updateUser(userState.getUsername(), {
+            'mailSyncEnabled': signupToNewsletter,
+            'settings': {
+              'onboardingCompleted': 'true'
+            }
+          })
+          .then(function (resp) {
+            userState.updateUserProfile(resp.item);
```
alex-deaconu: @ezequielc I would consider this a good refactoring if we moved the `userState.updateUserProfile()` call to the `updateUser` service. Or have a User factory we can call that does both. Thanks.

web/scripts/launcher/services/svc-onboarding-factory.js
```diff
@@ -235,22 +231,35 @@ angular.module('risevision.apps.launcher.services')
...
+
+            $rootScope.$emit('onboardingCompleted');
```
alex-deaconu: @ezequielc as per the above point, I would rather have a generic `userUpdated` event rather than this specific one. Thanks.

web/scripts/launcher/services/svc-onboarding-factory.js
```diff
@@ -235,22 +231,35 @@ angular.module('risevision.apps.launcher.services')
...
+            _setCurrentStep('promoteTraining');
+            _completeTabsUpTo(3);
+
+            segmentAnalytics.track('Onboarding Step 3 Completed', {
```
alex-deaconu: @ezequielc wondering if we should be more specific with the event names. `Onboarding Newsletter Signup Completed` maybe, in case we: decide to add more steps, or change the logic for step 3. Thoughts? Thanks.

web/scripts/launcher/services/svc-onboarding-factory.js
ezequielc: @alex-deaconu Yeah, I figured it would be useful but I didn't want to modify `updateUser` as it could have side effects. 
But instead of a new factory with similar functionality, maybe it is worth spending time and improve `updateUser` directly. Thoughts? Thanks

web/scripts/launcher/services/svc-onboarding-factory.js
ezequielc: @alex-deaconu Good suggestion, I will work on that. Thanks

web/scripts/launcher/services/svc-onboarding-factory.js
ezequielc: @alex-deaconu It makes sense. I wasn't very happy with the event name as well. I will update it. Thanks!

web/scripts/launcher/services/svc-onboarding-factory.js
alex-deaconu: @ezequielc I wouldn't worry about it for this PR, rather a note for refactoring. Thanks.

ezequielc: @alex-deaconu Makes sense. Thanks!

PR #1455, opened 2020-01-08
Hide fields if Player is not installed

alex-deaconu:
## Description
Hide resolution field
Hide Player Version field and don't highlight

Use playerProAuthorized value for license check
it is more accurate than the offlineSubscription

[stage-18]

## Motivation and Context
Showing those fields can be confusing. Resolution holds default values which are not an indication of the real resolution of the Display. The license information may be incorrect as it appears based on the Player report (which may not have happened). Also, the fact that it says "Player Version not Current" can be misleading to the User.

## How Has This Been Tested?
Validated all the scenarios with various Active/Inactive and Licensed/Unlicensed Displays.

## Release Plan:

No


alex-deaconu:
Will revert date change prior to merge.

Issue #1483, opened 2020-01-23
Assigning specific Displays in Placeholder distribution is removed when you switch between HTML and Design view


robb-price:
### What is the expected behavior?
Any changes you make to Placeholder Distribution in Design View should be saved when you press Save in HTML view.

### What is the actual behavior?
Changes you make to Placeholder Distribution in Design View are disregarded when you press Save in HTML view.

### What steps can be taken to replicate the problem?
1. In Design view, Set the distribution of a Placeholder to a specific Display.
2. Switch over to the HTML view, click SAVE, and switch back to the Design view. 
3. Notice the single Display is gone from the Distribution, and it now shows 'No Display Assigned'.

Darius brought this up in this Community conversation: https://help.risevision.com/hc/en-us/community/posts/360056411732-I-think-there-is-a-bug-to-report-1

alex-deaconu:
@robb-price this issue has been resolved. Thanks!

PR #1491, opened 2020-01-29
Fix selected displays not updating

alex-deaconu:
## Description
Fix selected displays not updating

When updating Placeholder distribution, switching
to HTML view, and saving, the Distribution displays
are lost

updateDistribution takes Distribution from the Placeholder list and
updates the presentation.distribution. However, when in
HTML mode, the Placeholder list is refreshed (cleared) based
on the HTML, so the Distribution is lost. This can be corrected
via parseDistribution, but that call is not needed
in this case because it's called after the update.

Also, fixing unit tests. Promise results do not count as failures
since they are not linked to the done sent in the timeout. Moving
timeout to Promise results.

[stage-2]

## Motivation and Context
Fix for #1483 

## How Has This Been Tested?
Replicated the issue in the current version. Debugged the source. Tested fix in the local environment. Validated with unit tests, noticed tests were not failing where expected so I had to do a bit of refactoring to the unit tests to get them to accurately test the functionality.

## Release Plan:

No

### DIFF
--- a/test/unit/editor/services/svc-editor-factory.tests.js
+++ b/test/unit/editor/services/svc-editor-factory.tests.js
@@ -78,24 +78,14 @@ describe('service: editorFactory:', function() {
         validatePresentation: function(presentation) {
           return true;
         },
-        parsePresentation: function(presentation) {
-          presentation.parsed = true;
-          return {};
-        },
-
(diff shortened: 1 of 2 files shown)


alex-deaconu:
@fjvallarino please review. Thanks!

PR #1510, opened 2020-02-07
Replace Segment with Google Tag Manager

ezequielc:
## Description

- Replace Segment initialization script with Google Tag Manager snippet.
- Remove GA linker code, moved to GTM.

For simplification, the directive was not renamed on this PR. A future PR will address that refactoring.

## Motivation and Context
New analytics architecture per Hubspot Simplification epic 

## How Has This Been Tested?
Manually validated on stage-1 and reviewed the events in [Segment debugger](https://app.segment.com/rise-vision/sources/apps_staging/debugger) and Google Analytics. 

## Release Plan:


web/scripts/config/prod.js
```diff
@@ -24,7 +24,9 @@
     .value('VIEWER_URL', 'http://preview.risevision.com')
     .value('ALERTS_WS_URL',
       'https://rvaserver2.appspot.com/alerts/cap')
-    .value('SEGMENT_API_KEY', 'pFutwqoxdVfhEF0J948V28YuzNLIP3MY')
+    .value('TAG_MANAGER_CONTAINER_ID', 'GTM-MMTK3JH')
+    .value('TAG_MANAGER_AUTH', null)
+    .value('TAG_MANAGER_ENV', null)
```
alex-deaconu: @ezequielc wondering if you could clarify why the Container Id is the same as staging, but the Auth and Env values are `null`? Actually, maybe it would help if you described what each variable was used for. Thanks.

web/scripts/components/logging/svc-segment-analytics.js
```diff
@@ -3,78 +3,72 @@
...
+              }
+            }
+          });
+        };
 
-            trackPageviews();
+        service.page = function (properties) {
+          $window.dataLayer.push({
+            event: 'analytics.page',
+            eventName: 'page viewed',
```
alex-deaconu: @ezequielc to confirm, this event name is not the same as described in the Architecture doc: `page viewed` vs `page view`. Thanks.

ezequielc: @alex-deaconu I updated the documentation. The correct name is what we have here 'page viewed'. Good catch! Thanks!

web/scripts/config/prod.js
ezequielc: @alex-deaconu These are some nuances of Tag Manager environments integration.
TAG_MANAGER_CONTAINER_ID is the identifier of the Tag Manager account, similar to Google Analytics tracking code.
TAG_MANAGER_AUTH and TAG_MANAGER_ENV are two url parameters to specify which environment of Tag Manager we want to use. From my investigation, it seems that each environment has its specific AUTH, so I could have used a single variable with all url params, but it is more flexible and readable this way.

The 'Live' environment of on the other hand, doesn't need TAG_MANAGER_AUTH and TAG_MANAGER_ENV, that is why they are null in production and beta.

Finally, I am also using TAG_MANAGER_CONTAINER_ID to turn on/off GTM integration. In Test environment we don't need tracking so TAG_MANAGER_CONTAINER_ID is set to null.

Thanks!

alex-deaconu: @ezequielc makes sense, thanks for the clarifications!

web/scripts/components/logging/app.js
```diff
@@ -9,8 +9,8 @@
 
   angular.module('risevision.common.config')
     .value('ENABLE_EXTERNAL_LOGGING', true)
-    // Production API key
-    .value('SEGMENT_API_KEY', 'pFutwqoxdVfhEF0J948V28YuzNLIP3MY');
+    // Google Tag Manager Stage environment
+    .value('TAG_MANAGER_CONTAINER_ID', 'GTM-MZJ2NQX');
```
alex-deaconu: @ezequielc wondering if we should add the other 2 variables here for completeness. Thanks!

ezequielc: @alex-deaconu Good point. I will add them.

web/scripts/components/logging/app.js
```diff
@@ -9,8 +9,10 @@
 
   angular.module('risevision.common.config')
     .value('ENABLE_EXTERNAL_LOGGING', true)
-    // Production API key
-    .value('SEGMENT_API_KEY', 'pFutwqoxdVfhEF0J948V28YuzNLIP3MY');
+    // Google Tag Manager Stage environment
```
alex-deaconu: @ezequielc I think this comment should be updated to reflect the `null` AUTH/ENV variables, ie. Production GTM env. Thanks.

ezequielc: Good catch. Updated. Thanks!

PR #1580, opened 2020-03-11
Use multiline flag to conditionally show text area for Text Component

ezequielc:
## Description
Use `multiline` flag to conditionally show text area for Text Component.

## Motivation and Context
Apps UX Improvements epic.

## How Has This Been Tested?
As per https://github.com/Rise-Vision/rise-text/pull/34.

## Release Plan:

### DIFF
--- a/web/partials/template-editor/components/component-text.html
+++ b/web/partials/template-editor/components/component-text.html
@@ -1,9 +1,9 @@
 <div class="attribute-editor-component">
   <div class="attribute-editor-row">
     <label>Text:</label>
-    <input ng-if="!isMultiline" id="text-component-input" type="text" class="form-control" ng-model="value" placeholder="Enter your text" ng-change="save()">
-    <p ng-if="isMultiline" class="text-sm">Hit enter to add a line break.</p>
-    <div ng-if="isMultiline" class="row">
+    <input ng-hide="isMultiline" id="text-component-input" type="text" class="form-control" ng-model="value" placeholder="Enter your text" ng-change="save()">
+    <p ng-show="isMultiline" class="text-sm">Hit enter to add a line break.</p>
+    <div ng-show="isMultiline" class="row">
       <div class="col-xs-12">
         ​<textarea id="text-component-input-multiline" class="form-control" rows="2" ng-model="value" placeholder="Enter your text" ng-change="save()"></textarea>
       </div>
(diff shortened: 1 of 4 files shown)


web/partials/template-editor/components/component-text.html
```diff
@@ -1,9 +1,11 @@
 <div class="attribute-editor-component">
   <div class="attribute-editor-row">
     <label>Text:</label>
-    <div class="row">
+    <input ng-hide="isMultiline" id="text-component-input" type="text" class="form-control" ng-model="value" placeholder="Enter your text" ng-change="save()">
```
alex-deaconu: @ezequielc wondering if we should use `ng-if` to show/hide these components. I'm concerned re: `ng-model` & `ng-change` triggering double updates because both elements are rendered & trigger events. Just a thought. Thanks.

web/scripts/template-editor/components/directives/dtv-component-text.js
```diff
@@ -15,6 +15,7 @@ angular.module('risevision.template-editor.directives')
           };
 
           function _load() {
+            $scope.isMultiline = $scope.getBlueprintData($scope.componentId, 'multiline');
             var value = $scope.getAvailableAttributeData($scope.componentId, 'value');
```
alex-deaconu: @ezequielc curious, what's the difference between `getAvailableAttributeData` and `getBlueprintData`? ie. why did you use the second function instead of what the other variables are using? Thanks.

web/partials/template-editor/components/component-text.html
ezequielc: @alex-deaconu I originally had `ng-if` but I had to change to this because as we load the component data on a later moment, these elements were not properly handling the model and were failing to update the attribute values, i.e. not saving nor updating the template preview.
If you think it is important I can try to spend more time into this. Thanks

alex-deaconu: @ezequielc I'm good with what you have. Thanks.

web/scripts/template-editor/components/directives/dtv-component-text.js
ezequielc: @alex-deaconu `getAvailableAttributeData` gives you the data that was entered by the users in the UI. So for example, if users change the font size, it will return the last entered font size.
`getBlueprintData` will always return the value from blueprint file. In our case, `multiline` is not an attribute we want users to modify, so I am always grabbing the blueprint value. Thanks

alex-deaconu: @ezequielc thanks for the clarification!

web/partials/template-editor/components/component-text.html
ezequielc: Sounds good. Thanks!

PR #1618, opened 2020-03-31
Use case: if template is PUD then it is added to the schedule as PUD

olegrise:
## Description
Implementation of the use case "As a User, when I schedule an HTML Template that contains one or more HTML Templates configured for Play Until Done, I can see that the duration of the scheduled playlist item is set to Play Until Done, so that I know my content will play all the way through before moving onto the next item"

## Motivation and Context
This is an Epic requirement.

## How Has This Been Tested?
Updated unit test and also manually verified that if a template is PUD then it is added to the schedule as PUD.

## Release Plan:


olegrise:
@stulees please review

PR #1704, opened 2020-05-25
Use selected company id to check for Free Displays when adding new Schedule

### LINKED ISSUE #1703: Missing Display License appears when saving a new Schedule with only Licensed Displays
### What is the expected behavior?
If all displays on a Schedule are licensed, no message is shown when it is saved.

### What is the actual behavior?
Missing Display License appears when saving a new Schedule even if all Displays are licensed.

### What steps can be taken to replicate the problem?
1. Create a schedule.
2. Name it.
3. Add 1 Display that is licensed.
4. Save.

https://drive.google.com/file/d/1ypXe39Nfq3qXDJaLAicxoOzbWSIvH52o/view?usp=drivesdk

### Some additional notes
- If I save without any displays, no error appears.  If I then add the licensed display, no error appears.
- If I select all displays, I get the error every time.

ezequielc:
## Description
Use selected company id to check for Free Displays when adding new Schedule

## Motivation and Context
Fix #1703.
When adding a new Schedule in a subcompany, the company Id was missing, causing the request to check for Free Displays in the parent company. 

## How Has This Been Tested?
Manually tested locally and on stage-1.

## Release Plan:

### DIFF
--- a/web/scripts/schedules/services/svc-schedule-factory.js
+++ b/web/scripts/schedules/services/svc-schedule-factory.js
@@ -2,9 +2,9 @@
 
 angular.module('risevision.schedules.services')
   .factory('scheduleFactory', ['$q', '$state', '$log', '$rootScope', 'schedule', 'scheduleTracker',
-    'processErrorCode', 'VIEWER_URL', 'HTML_PRESENTATION_TYPE', 'display', 'plansFactory',
+    'processErrorCode', 'VIEWER_URL', 'HTML_PRESENTATION_TYPE', 'display', 'plansFactory', 'userState',
     function ($q, $state, $log, $rootScope, schedule, scheduleTracker, processErrorCode,
-      VIEWER_URL, HTML_PRESENTATION_TYPE, display, plansFactory) {
+      VIEWER_URL, HTML_PRESENTATION_TYPE, display, plansFactory, userState) {
       var factory = {};
       var _hasSchedules;
       var _scheduleId;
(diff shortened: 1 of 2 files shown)


web/scripts/schedules/services/svc-schedule-factory.js
```diff
@@ -21,6 +21,7 @@ angular.module('risevision.schedules.services')
         _scheduleId = undefined;
 
         factory.schedule = {
+          companyId: userState.getSelectedCompanyId(),
```
alex-deaconu, line 24: @ezequielc wondering if this could cause issues elsewhere, and we should just check for an empty Company Id at `display.hasFreeDisplays(factory.schedule.companyId...)` (Line 120). Thoughts? Thanks.

ezequielc: @alex-deaconu I thought about that option, but I decided to add to the empty schedule to prevent future issues, in case we miss this case again. 
I checked that companyId is not picked by [SCHEDULE_WRITABLE_FIELDS](https://github.com/Rise-Vision/rise-vision-apps/blob/master/web/scripts/schedules/services/svc-schedule.js#L6) so it won't affect server updates. Also, I took a look and it is not being referenced by other services, so I think it is safe. Thanks!

alex-deaconu: @ezequielc makes sense, thanks!

PR #1757, opened 2020-06-19
Saving on change made and backwards compatible [stage-3]

duckbeartrap:
## Description
Implemented saving changes and compatibility with existing presentations.

## Motivation and Context
https://trello.com/c/czQqbQ7Y/65-add-tinymce-to-apps-2

## How Has This Been Tested?
Tested visually.

## Release Plan:

### DIFF
--- a/web/scripts/template-editor/components/directives/dtv-component-text.js
+++ b/web/scripts/template-editor/components/directives/dtv-component-text.js
@@ -25,9 +25,12 @@ angular.module('risevision.template-editor.directives')
               'removeformat code'
           };
 
+          $scope.richText = '';
+
           function _load() {
             $scope.isMultiline = $scope.getBlueprintData($scope.componentId, 'multiline');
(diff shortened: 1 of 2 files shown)


web/scripts/template-editor/components/directives/dtv-component-text.js
```diff
@@ -50,7 +54,7 @@ angular.module('risevision.template-editor.directives')
           }
 
           $scope.save = function () {
-            $scope.setAttributeData($scope.componentId, 'value', $scope.value);
```
olegrise, line 53: @willread I just realized that it's probably better to clear the 'value' than not saving it.  
```
$scope.setAttributeData($scope.componentId, 'value', undefined);
```

duckbeartrap: Yes, but is it a good idea to set it to undefined everytime something changes. We can set it to undefined on load, after we set the old value into the richText variable. What do you think?

PR #1837, opened 2020-07-08
Add To Schedule modal behaviour

ezequielc:
## Description
Show modal if Presentation not assigned to Schedules.
Disable modal's Publish button while not assigned to Schedules.
Handle loading status/spinners.
Skip modal if Onboarding (i.e. no schedules in the company).
Refactored circular reference.
Only show Schedules attributes list error after modal is dismissed

## Motivation and Context
Editor Improvements

## How Has This Been Tested?
Manually tested locally.
[stage-1]

## Release Plan:


web/partials/template-editor/attribute-list.html
```diff
@@ -22,11 +22,11 @@
 </div>
 
 <div class="attribute-row attribute-row-hover schedules-attribute-row mt-2" require-role="cp">
-  <div class="col-xs-10 pl-0 attribute-desc" ng-class="{'danger' : !schedulesComponent.hasSelectedSchedules }">
+  <div class="col-xs-10 pl-0 attribute-desc" ng-class="{'danger' : !factory.isAssignedToSchedules }">
     <span>
       Schedules
     </span>
-    <div class="error-message mb-3" ng-show="!schedulesComponent.hasSelectedSchedules">
+    <div class="error-message mb-3" ng-show="!factory.isAssignedToSchedules">
```
alex-deaconu, line 29: @ezequielc I think we can remove the dependency to `scheduleSelectorFactory` from this directive since we don't need `hasSelectedSchedules` anymore. ie. we can move the load from the attribute-list to the component open, and probably deprecate `getSchedulesComponent`. I can make a PR for that. Thoughts? Thanks.

web/scripts/schedules/services/svc-schedule-selector-factory.js
```diff
@@ -155,6 +155,28 @@ angular.module('risevision.schedules.services')
           .then(_loadSelectedSchedules);
       };
 
+      factory.checkAssignedToSchedules = function() {
+        return companyAssetsFactory.hasSchedules().then(function(hasSchedules) {
```
alex-deaconu, line 159: @ezequielc in the main branch you were using `scheduleFactory.hasSchedules`. Curious what is the difference? Thanks.

web/scripts/template-editor/services/svc-template-editor-factory.js
```diff
@@ -229,20 +230,11 @@ angular.module('risevision.template-editor.services')
...
-        }
-        var modalInstance = $modal.open({
-          templateUrl: 'partials/template-editor/add-to-schedule-modal.html',
-          controller: 'AddToScheduleModalController',
-          windowClass: 'madero-style centered-modal',
-          size: 'sm'
-        });
-        return modalInstance.result;
+        return scheduleSelectorFactory.checkAssignedToSchedules()
+          .then(_publish)
```
alex-deaconu, line 234: @ezequielc wondering if we should still `publishBranding`, ie. move this functionality to `_publishPresentation`? Thoughts? Thanks.

web/partials/template-editor/attribute-list.html
ezequielc: @alex-deaconu I noticed we have a defect here. Once we set `factory.isAssignedToSchedules`, it is not being updated if we assign to a schedule from the attributes list. The error message will only be cleared if we click Publish. 
So, I agree with your suggestion, but we need to take a fix for that into account when we remove the dependency. 
Up to you if you want to work on this. I could also do it once I finish unit tests. Thanks

web/scripts/schedules/services/svc-schedule-selector-factory.js
ezequielc: @alex-deaconu [`scheduleFactory.hasSchedules()`](https://github.com/Rise-Vision/rise-vision-apps/blob/b0927dca7c95fcc0b0b4ba76761c645aa5e0eb1b/web/scripts/schedules/services/svc-schedule-factory.js#L74) starts an asynchronous call but returns immediately. It wasn't working 100% of the time for my case, so I replaced it with this asynchronous function. Thanks

web/scripts/template-editor/services/svc-template-editor-factory.js
ezequielc: @alex-deaconu Interesting thought. I think it makes more sense to keep both Branding and Presentation as draft if we could not proceed with the Publish button action, so I would keep it as is. But let me know if you disagree. Thanks!

web/partials/template-editor/attribute-list.html
alex-deaconu: @ezequielc interesting. I'll defer to you for the fix. Thanks.

web/scripts/schedules/services/svc-schedule-selector-factory.js
alex-deaconu: @ezequielc sounds good, thanks!

web/scripts/template-editor/services/svc-template-editor-factory.js
alex-deaconu: @ezequielc I agree. Just wanted to confirm this is what we expect the behaviour to be if there are no unsaved changes to the Template and we only have Branding changes. Thanks.

web/partials/template-editor/attribute-list.html
ezequielc: @alex-deaconu Sounds good. Thanks!

PR #1915, opened 2020-07-29
Add callout/tooltip to branding override

olegrise:
## Description
- Change `tooltip-overlay` directive to allow to have more than one directives/callout/tooltip per page.
- Add callout/tooltip to the branding override directive

Couple of thing to note:
- The callout is implemented using two different libraries: [hone](https://github.com/benmarch/hone) and [bootstrap](https://github.com/angular-ui/bootstrap/tree/baf0a567f9f4641848a9a4699695f9cde5f28f70/src/tooltip). The hone is used for highlighting selected element and adding backdrop to the page. The bootstrap is used for showing the tooltip window with a message. In the HTML snippet below the `tooltip-overlay` and `tooltip-overlay-key` attributes are handled by `tooltip-overlay` directive while all other attributes that start with `tooltip-*` are handled by the bootstrap library.
```
  <div class="attribute-row"
  tooltip-overlay tooltip-overlay-key="BrandingOverrideTooltip"
  tooltip-template="'partials/template-editor/branding-override-tooltip.html'" tooltip-placement="bottom" tooltip-class="madero-style tooltip-guide" tooltip-append-to-body="true">
``
- The problem with the `tooltip-overlay` directive was that it watches the `$scope.tooltipKey` which makes it hard to have more than one directive in the `attribute-list.html`. This is solved by adding the `tooltip-overlay-key` attribute. If not set, then the behaviour of the `tooltip-overlay` directive is the same as before. If set, then tooltip only shows if `tooltip-overlay-key` matches the `$scope.tooltipKey`.

## Motivation and Context
Epic requirements

## How Has This Been Tested?
Visually and with unit tests.

## Release Plan:

### CODE REVIEW

**Comment by olegrise** [comment]
@alex-deaconu FYI

**Comment by olegrise** [comment]
@stulees btw if you need to reset/restart the callout count, edit User.settings in the datastore. You might need to clear the GAE memcache too.

**Comment by ezequielc** [comment]
@olegrise @stulees Another suggestion to reset the counter is to set [this line](https://github.com/Rise-Vision/rise-vision-apps/blob/c2afcc3b8e9a6088c8d6e744cbbe94095de513e5/web/scripts/common/services/svc-tour-factory.js#L22) to return 0 in our local Apps, as it will persist and update the datastore. Thanks!

File: web/scripts/common/directives/dtv-tooltip-overlay.js [hunk 1]
```diff
@@ -42,6 +42,12 @@ angular.module('risevision.apps.directives')
                   return;
                 }
 
+                //tooltip-overlay-key is optinal. It allows to have more than directive on a page.
```
alex-deaconu, line 45: @olegrise typo here. Thanks.

web/scripts/common/directives/dtv-tooltip-overlay.js
```diff
@@ -42,6 +42,12 @@ angular.module('risevision.apps.directives')
...
+                if (element.attr('tooltip-overlay-key') &&
+                  element.attr('tooltip-overlay-key') !== $scope.tooltipKey) {
```
alex-deaconu, line 47: @olegrise curious why add a new parameter and not use `tooltip-overlay="key"`? Thanks.

olegrise: @alex-deaconu I'm good either way. My logic was that having a separate "dedicated" parameter makes it clear what this parameter is for. Also I was thinking about `rv-spinner` and how it has a separate 'rv-spinner-key' parameter.

olegrise:
@alex-deaconu I like yur suggestion re. set the tooltipKey to the array. I'm going to take a look at it.

web/scripts/common/directives/dtv-tooltip-overlay.js
alex-deaconu: @olegrise good point, agree to leave as is. Thanks!

olegrise:
@alex-deaconu actually, I'm not sure if that's going to work or simplify anything. Let's say we set 
```
$scope.tooltipKey = ['ScheduleSelectorTooltip', 'BrandingOverrideTooltip']
```
in [here](https://github.com/Rise-Vision/rise-vision-apps/pull/1915/files#diff-fee5749d9b084d30629c9d054ee7c889R34) , then `$watch` will be triggered in both defectives, and that in turn will create a race condition who calls the factory first.  I don't like it. Am I missing something?

alex-deaconu:
@olegrise to confirm, as it is written right now, the order right is the order of the array, right? So if `tourFactory` is sent an array, it will choose the first key, show it, and if it's not supposed to show, show the second key. Thanks.

olegrise:
@alex-deaconu re. "order is the order of the array" - correct. 
The different is that right now we call `tourFactory.findActiveKey()` only once. If I understand your suggestion correctly, then the `findActiveKey` will be called twice potentially returning different results every time.

stulees:
@olegrise Looks good to me 👍 , but you and @alex-deaconu are closer to this so I'll defer to @alex-deaconu for an approval

alex-deaconu:
@olegrise just a heads up, it looks like your stage is overwritten. I cannot get the tooltip to show. Thanks.

olegrise:
@alexey-rise restaged. Thanks.

alexey-rise:
@olegrise I am guessing you meant this message for someone else? :-)

olegrise:
@alexey-rise please disregard. @alex-deaconu restaged.

olegrise:
@stulees there is a problem with this PR. Basically the Schedules tooltip does not disappear. See the [video](https://drive.google.com/file/d/1qTcel4FBbhRiUdKd_8U4QlFw26mkMLTi/view?usp=sharing). I'm pretty sure that it was working fine before. The strange thing is that showing the tooltip works, but closing doesn't. I double checked the code and googled the problem, but no luck. At the moment I don't have any ideas how to fix it. My feeling is that this is the kind of issue that may consume a lot of time. I think the best option is to release without callout especially if we are going to replace it with something else in the near future.

stulees:
@olegrise @alex-deaconu FYI decision was made in SoS to not pursue this any further given we are abandoning the use of this tooltip component soon (Onboarding epic). @olegrise I'll leave this open for now and let you close it when you return.

olegrise:
Closing this PR. See https://github.com/Rise-Vision/rise-vision-apps/pull/2004

PR #1920, opened 2020-07-30
Optimize address validation

### LINKED ISSUE #1888: Apps calling avalara api on page save
Apps is calling the avalara api every time the company settings page is saved to validate the company address. We are also calling the api when the display settings page is saved if the display address is being used for weather. 

In both cases we should only call the api if the address is changed.

ezequielc:
## Description
Perform address validation only if address fields were changed.
Add this behaviour to Company Settings, New Sub-Company and Display Address forms.

## Motivation and Context
Fixes #1888.

## How Has This Been Tested?
Locally and on stage-1.

## Release Plan:

### DIFF
--- a/web/scripts/components/purchase-flow/services/svc-address.js
+++ b/web/scripts/components/purchase-flow/services/svc-address.js
@@ -89,5 +89,17 @@ angular.module('risevision.common.components.purchase-flow')
           !addressObject.province;
       };
 
+      this.isAddressDirty = function(addressForm) {
+        console.log("XX addressForm", addressForm);
+        var addressFields = ['street', 'unit', 'city', 'province', 'country', 'postalCode'];
+        var isDirty = false;
+        angular.forEach(addressFields, function(field) {
+          if (addressForm[field] && addressForm[field].$dirty) {
+            isDirty = true;
+          }
+        });
+        return isDirty;
+      };
+
     }
   ]);
--- a/web/scripts/components/purchase-flow/services/svc-address.js
+++ b/web/scripts/components/purchase-flow/services/svc-address.js
@@ -89,8 +89,10 @@ angular.module('risevision.common.components.purchase-flow')
           !addressObject.province;
       };
 
-      this.isAddressDirty = function(addressForm) {
-        console.log("XX addressForm", addressForm);
+      this.isAddressFormDirty = function(addressForm) {
+        if (!addressForm) {
+          return false;
+        }
         var addressFields = ['street', 'unit', 'city', 'province', 'country', 'postalCode'];
         var isDirty = false;
         angular.forEach(addressFields, function(field) {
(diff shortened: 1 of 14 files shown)


web/scripts/components/purchase-flow/services/svc-address.js
```diff
@@ -89,5 +89,19 @@ angular.module('risevision.common.components.purchase-flow')
           !addressObject.province;
       };
 
+      this.isAddressFormDirty = function(addressForm) {
```
alex-deaconu, line 92: @ezequielc I'd make this a generic service - `checkDirty(form, fieldsArray)`. Thanks.

web/scripts/displays/controllers/ctr-display-details.js
```diff
@@ -177,7 +177,8 @@ angular.module('risevision.displays.controllers')
 
           return $q.reject();
         } else {
-          return displayFactory.updateDisplay();
+          var shouldSkipAddressValidation = !addressService.isAddressFormDirty($scope.displayDetails) && !$scope.displayDetails.useCompanyAddress.$dirty;
```
alex-deaconu, line 180: @ezequielc what are your thoughts of just putting the form on the `display` object? Thanks.

ezequielc: @alex-deaconu I would like to avoid that as it mixes different concerns. Forms are UI elements and I would prefer to avoid them being attached to data/business objects.
I actually started my implementation by passing the form to the displayFactory, but gave up on that for this reason. Thanks

web/scripts/components/purchase-flow/services/svc-address.js
ezequielc: @alex-deaconu That is an interesting suggestion. If I need to make other changes to this PR I will do it, but if not, I would leave this refactoring for a future PR, if good with you. Thanks!

web/scripts/displays/controllers/ctr-display-details.js
alex-deaconu: @ezequielc right, but `addressFactory.validateAddressIfChanged($scope.forms.companyForm, $scope.company)` above seems to do that. Thanks.

web/scripts/components/purchase-flow/services/svc-address.js
alex-deaconu: @ezequielc sounds good, just thinking this could be useful for other things. Thanks.

ezequielc:
@alex-deaconu Good catch, I will see how can I fix it. Thanks

web/scripts/displays/controllers/ctr-display-details.js
alex-deaconu: @ezequielc to confirm, I don't feel strongly about this point. Overall, it just seems we have 3 different mechanisms for validating addresses (display, company settings, purchase flow) all with slightly different logic. Which makes it hard to track and fix issues. Just some thoughts. Thanks.

ezequielc: @alex-deaconu That is true. Maybe I should move it to the controllers and avoid duplication somehow. Let me know if you have nay suggestions. Thanks

PR #2062, opened 2020-09-25
Use CSS to set draggable mirror offset

### LINKED ISSUE #2053: Clicking and dragging playlist items in a schedule on mobile doesn't work
iPhone / iOS 14 / Safari 

1. Open a schedule on mobile
2. Add multiple playlist items 
3. Attempt to click and drag

Video: https://drive.google.com/file/d/1PBxaBdtvqeofPKmScCKviHIr5RL3CvtO/view?usp=sharing

The icons for copy and trash are also misaligned. @jasoncaldwell how should these icons be aligned on mobile?

ezequielc:
## Description
Due to a limitation of the Draggable library, scrolling offset was not taken into account when dragging on a touch device, causing the mirror to be positioned incorrectly.
Instead of using cursorOffsetY, we are now using CSS top property to position the mirror element.

## Motivation and Context
Fix #2053.

## How Has This Been Tested?
Locally and validated on a real Android/Chrome device. [stage-1]

## Release Plan:

### DIFF
--- a/test/e2e/schedules/cases/add-distribution.js
+++ b/test/e2e/schedules/cases/add-distribution.js
@@ -64,6 +64,7 @@ var AddDistributionScenarios = function() {
 
         it('should show a table for listing displays', function () {
           helper.waitDisappear(distributionModalPage.getDistributionListLoader()).then(function () {
+            helper.wait(distributionModalPage.getDistributionListTable(), 'Distribution Table');
             expect(distributionModalPage.getDistributionListTable().isDisplayed()).to.eventually.be.true;
           });
           
--- a/test/e2e/schedules/cases/add-url.js
+++ b/test/e2e/schedules/cases/add-url.js
@@ -54,6 +54,7 @@ var AddUrlScenarios = function() {
           before(function () {
             scheduleAddPage.getAddUrlItemButton().click();
             helper.wait(playlistItemModalPage.getPlaylistItemModal(), 'Add URL Item');
+            browser.sleep(500);
           });
 
           it('should open the Add URL Modal', function () {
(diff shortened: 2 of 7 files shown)


ezequielc:
@alex-deaconu Do you have a physical  iOS device to validate this fix? 
I tried to use BrowserStack but it seems the touch event is not working 100% there. I tested with [an example](https://shopify.github.io/draggable/examples/simple-list.html) from the library and it didn't work as well, so I think it might be something with BrowserStack. Thanks!

test/e2e/schedules/cases/add-distribution.js
```diff
@@ -64,6 +64,7 @@ var AddDistributionScenarios = function() {
 
         it('should show a table for listing displays', function () {
           helper.waitDisappear(distributionModalPage.getDistributionListLoader()).then(function () {
+            helper.wait(distributionModalPage.getDistributionListTable(), 'Distribution Table');
```
alex-deaconu, line 67: @ezequielc thoughts on moving this line up before waiting for the spinner to show. Thanks.

alex-deaconu:
@ezequielc tested on my iphone, Chrome and Safari, both Template Editor and Schedule page. Looks good. Thanks for fixing!

alex-deaconu (approved):

test/e2e/schedules/cases/add-distribution.js
ezequielc: @alex-deaconu We only show that list if there are displays: https://github.com/Rise-Vision/rise-vision-apps/blob/f7282bf4ac9d804d4e2287fc628d9b34375611a5/web/partials/components/distribution-selector/distribution-list.html#L17, so I don't think it is present before the spinner ends. Thanks

ezequielc:
@alex-deaconu Thanks for testing this 👍

test/e2e/schedules/cases/add-distribution.js
alex-deaconu: @ezequielc sounds good, thanks.

PR #2077, opened 2020-10-01
Add batching to operations

alex-deaconu:
## Description
Add batching to operations

Batch size of 5, configurable
This means only 5 active requests at one time

Technically multiple batches of different operations
can be queued

[stage-19]

## Motivation and Context
Sending all operations at once would overrun the server, decided to batch so we only have a limited number of operations being executed at one time.

## How Has This Been Tested?
Tested changes locally with different batch sizes. Updated unit test coverage.

## Release Plan:

No

### DIFF
--- a/web/scripts/components/scrolling-list/svc-batch-operations.js
+++ b/web/scripts/components/scrolling-list/svc-batch-operations.js
@@ -1,0 +1,71 @@
+'use strict';
+
+angular.module('risevision.common.components.scrolling-list')
+  .factory('BatchOperations', ['$q', '$timeout',
+    function ($q, $timeout) {
+      return function () {
+        var svc = {};
+        svc.queueLimit = 5;
+
+        var queue;
+
+        var _reset = function() {
+          queue = [];
+
+          svc.isActive = false;
+          svc.totalItemCount = 0;
+          svc.completedItemCount = 0;
+        };
+
+        _reset();
+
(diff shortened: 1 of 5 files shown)


alex-deaconu:
Set an arbitrary batch number of 5 for now.

web/scripts/components/scrolling-list/svc-batch-operations.js
```diff
@@ -0,0 +1,71 @@
...
+            if (currItem < items.length) {
+              while (queue.length < svc.queueLimit && currItem < items.length) {
+                var item = items[currItem++];
+
+                queue.push(item);
+
+                _executeOperation(item, method);
+              }
+
+              $timeout(loadBatch, 500);
```
ezequielc, line 59: @alex-deaconu Instead of using a timer here, what about relying on .finally() of _executeOperation() to call loadBatch again? Thanks

ezequielc (approved):
@alex-deaconu Looks good 👍 
One suggestion inline. Thanks

web/scripts/components/scrolling-list/svc-batch-operations.js
alex-deaconu: @ezequielc this is inherited from the old code, you suggestion makes sense however, I will implement it. Thanks!

PR #2096, opened 2020-10-06
Prevents downloadable content to be added as URL items

### LINKED ISSUE #2068: URL Schedule Playlist items which link to files cause automatic downloads without permission when previewed
### What is the expected behavior?
Playlist items can be added to a schedule by URL, allowing you to play a web page that is hosted externally on a rise vision display.

### What is the actual behavior?
Playlist items can be added to a schedule by URL, but there is no restriction to the URL used.  Adding a URL that is not a web page triggers an automatic download without prompt when the page is previewed.

### What steps can be taken to replicate the problem?
- Upload a file like a Powerpoint presentation to Rise Storage.
- Obtain the link to the file in Rise Storage.
- Add it to your schedule playlist by URL.
- Copy the shared schedule link for the presentation.
- Paste it into a web browser.

ezequielc:
## Description
Check content-type of entered URL items and only accepts webpages, images or videos.

## Motivation and Context
Fixes #2068

## How Has This Been Tested?
Locally and on [stage-1]

## Release Plan:

### DIFF
--- a/web/partials/components/url-field/messages.html
+++ b/web/partials/components/url-field/messages.html
@@ -3,7 +3,7 @@
 <p ng-message="noPreviewUrl" class="text-danger">Using Preview URLs is not supported. You can place one Presentation into another using Embedded Presentations.</p>
 <p ng-message="responseHeaderValidator" class="text-danger">
 	<span ng-if="ngModelCtrl.responseHeaderValidatorError === 'not-reachable'">The requested URL could not be reached. Please try again.</span>
-	<span ng-if="ngModelCtrl.responseHeaderValidatorError === 'content-type'">Invalid content type. Please provide a webpage, image or video URL.</span>
+	<span ng-if="ngModelCtrl.responseHeaderValidatorError === 'content-type'">Invalid content type. Please provide a webpage, image, audio or video URL.</span>
 	<span ng-if="ngModelCtrl.responseHeaderValidatorError !== 'content-type' && ngModelCtrl.responseHeaderValidatorError !== 'not-reachable'">The owner of the Web Page at the URL provided does not allow the page to be embedded within an iFrame. If possible, please contact the Web Page owner to discuss ( <a target="_blank" href="https://developer.mozilla.org/en-US/docs/Web/HTTP/X-Frame-Options">X-Frame-Options</a> / <a target="_blank" href="https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-ancestors">Content-Security-Policy</a>).</span>
 </p>
 <p ng-message="fileType" class="text-danger">{{ngModelCtrl.customErrorMessage}}</p>
--- a/web/partials/components/url-field/messages.html
+++ b/web/partials/components/url-field/messages.html
@@ -1,6 +1,10 @@
 <p ng-message="required" class="text-danger">This field is required.</p>
 <p ng-message="pattern" class="text-danger">Please provide a valid URL.</p>
 <p ng-message="noPreviewUrl" class="text-danger">Using Preview URLs is not supported. You can place one Presentation into another using Embedded Presentations.</p>
-<p ng-message="responseHeaderValidator" class="text-danger">The owner of the Web Page at the URL provided does not allow the page to be embedded within an iFrame. If possible, please contact the Web Page owner to discuss ( <a target="_blank" href="https://developer.mozilla.org/en-US/docs/Web/HTTP/X-Frame-Options">X-Frame-Options</a> / <a target="_blank" href="https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-ancestors">Content-Security-Policy</a>).</p>
+<p ng-message="responseHeaderValidator" class="text-danger">
+	<span ng-if="ngModelCtrl.responseHeaderValidatorError === 'not-reachable'">The requested URL could not be reached. Please try again.</span>
+	<span ng-if="ngModelCtrl.responseHeaderValidatorError === 'content-type'">Invalid content type. Please provide a webpage, image or video URL.</span>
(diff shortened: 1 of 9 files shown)


web/scripts/components/url-field/svc-response-header-analyzer.js
```diff
@@ -51,7 +56,12 @@ angular.module('risevision.widget.common.url-field.response-header-analyzer', []
           options.push('frame-ancestors');
         }
         return options;
-      }
+      };
+
+      var _isAcceptedContentType = function(contentType) {
+        $log.debug('URL contentType:', contentType);
+        return contentType && contentType.match('(image\/|video\/|text\/).*');
```
alex-deaconu, line 63: @ezequielc would there be there be other content types that the browser can render, like audio or video, xml, csv, etc? Thanks.

web/scripts/components/url-field/svc-response-header-analyzer.js
```diff
@@ -9,37 +9,42 @@ angular.module('risevision.widget.common.url-field.response-header-analyzer', []
...
+            deferred.reject('not-reachable');
+          });
         return deferred.promise;
       };
 
-      factory.getOptions = function (url) {
+      var _requestHead = function (url) {
         return $http({
-          method: 'GET',
+          method: 'HEAD',
```
alex-deaconu, line 42: @ezequielc nice!

web/partials/components/url-field/messages.html
```diff
@@ -1,6 +1,10 @@
...
+<p ng-message="responseHeaderValidator" class="text-danger">
+	<span ng-if="ngModelCtrl.responseHeaderValidatorError === 'not-reachable'">The requested URL could not be reached. Please try again.</span>
+	<span ng-if="ngModelCtrl.responseHeaderValidatorError === 'content-type'">Invalid content type. Please provide a webpage, image or video URL.</span>
+	<span ng-if="ngModelCtrl.responseHeaderValidatorError !== 'content-type' && ngModelCtrl.responseHeaderValidatorError !== 'not-reachable'">The owner of the Web Page at the URL provided does not allow the page to be embedded within an iFrame. If possible, please contact the Web Page owner to discuss ( <a target="_blank" href="https://developer.mozilla.org/en-US/docs/Web/HTTP/X-Frame-Options">X-Frame-Options</a> / <a target="_blank" href="https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-ancestors">Content-Security-Policy</a>).</span>
```
alex-deaconu, line 7: @ezequielc curious why not check for the `X-frame-options` string directly? Thanks.

ezequielc: @alex-deaconu Mainly because the reject reason could be either `X-frame-options` or `frame-ancestors`. But maybe we should combine into a single reason, given we show the same error message. Thoughts? Thanks

web/scripts/components/url-field/svc-response-header-analyzer.js
```diff
@@ -51,7 +56,12 @@ angular.module('risevision.widget.common.url-field.response-header-analyzer', []
           options.push('frame-ancestors');
         }
         return options;
-      }
+      };
+
+      var _isAcceptedContentType = function(contentType) {
+        $log.debug('URL contentType:', contentType);
```
alex-deaconu, line 62: @ezequielc should we be logging here? Thanks.

web/scripts/components/url-field/svc-response-header-analyzer.js
ezequielc: @alex-deaconu Thanks! Yeah, it had downloaded an entire 50MB video file when I added for testing, so I figured we only need a HEAD request.

web/scripts/components/url-field/svc-response-header-analyzer.js
alex-deaconu: Also applies to other new `$log` entries in the file. Thanks.

web/partials/components/url-field/messages.html
alex-deaconu: @ezequielc just saw that there are two reasons in the code; missed it. I don't feel strongly either way, so we can leave as is. Thanks.

web/scripts/components/url-field/svc-response-header-analyzer.js
ezequielc: @alex-deaconu CSV and XML should fall into `text/` (text/csv, text/xml). 
I thought about audio but wasn't sure if we should accept it or not. But since you pointed out, I guess there is no reason to block it, right?

web/scripts/components/url-field/svc-response-header-analyzer.js
ezequielc: @alex-deaconu This is the only one that is new, and I agree we don't really need it. It will be logged as part of `$log.debug('URL headers:', response.headers());` Thanks

web/scripts/components/url-field/svc-response-header-analyzer.js
alex-deaconu: @ezequielc got it, I'm not sure either, just want to be safe. Thanks.

alex-deaconu:
@ezequielc 2 general notes:
- As `url-field` is used in other places, do you think this functionality would impact that?
- What is the response for Blocked files in Storage, just want to ensure that we wouldn't show the "Not found" error in those cases.

Thanks.

web/scripts/components/url-field/svc-response-header-analyzer.js
alex-deaconu: @ezequielc sounds good, I don't feel strongly either way. Thanks.

web/scripts/components/url-field/svc-response-header-analyzer.js
ezequielc: @alex-deaconu I will add, for completeness. Thanks

ezequielc:
@alex-deaconu Other url-fields are not going to be impacted as Playlist Item is the only place where `response-header-validator` is used.

Re Blocked files in Storage, I need to test but I assume it would be considered non-reachable. But having said that, I don't think it is worth investing time to circumvent that as it is going to be a very rare case, and the error message makes it very clear that something is wrong with the file. Let me know if you agree. Thanks!

alex-deaconu:
@ezequielc sounds good on both counts, thanks for the clarifications!

alex-deaconu (approved):
@ezequielc looks good. Thanks!

alex-deaconu (approved):

PR #2113, opened 2020-10-12
Re-enable displays summary

ezequielc:
## Description
Re-enable displays summary

## Motivation and Context
Re-enables displays summary as https://github.com/Rise-Vision/core/issues/810 is fixed.

## How Has This Been Tested?
Visually confirmed https://apps-beta.risevision.com/

## Release Plan:


alex-deaconu (approved):
@ezequielc looks good, however to confirm this is not yet ready to be released, pending Core fixes. Correct? Thanks.

ezequielc:
@alex-deaconu Correct, we thought it was solved yesterday, but Tyler found new issues around it, so not good to be merged yet. Thanks

tejohnso:
The summary is still not working properly.

The *displays* index is incorrectly showing some displays as `not_activated`.
I'm looking at a display having `null` for *lastContentUpdateDate* and `not_activated` in the index, yet messaging service has a last connection date, and the display page shows a screenshot.

If the [DisplayDevice](https://github.com/Rise-Vision/core/blob/master/src/main/java/com/risevision/core/data/DisplayDevice.java#L345) has a `null` *lastActivityDate* and *connectionStatusChangedate* it will indicate the display is `not_activated`. 

And the *lastActivityDate* expects either a *connectionStatusChangeDate* or a *lastContentUpdateDate*.

Displays that haven't recently connected won't have a *connectionStatusChangeDate* and if they also don't have *lastContentUpdateDate* set, the display is indexed as `not_activated`. The *lastContentUpdateDate* seems to come from the old viewer content fetch [resource](https://github.com/Rise-Vision/core/blob/master/src/main/java/com/risevision/core/resources/viewer/DisplayContentResource.java#L228). So it makes sense that will be empty in a lot of cases. 

I'm not sure what approach to take to fix this. We could run a job to update the *connectionStatusChangeDate*s based on all dates available in the messaging service but I'm hoping there is a more efficient solution.

tejohnso:
For the purposes of *display/summary*, can we consider `not_activated` displays as `offline`? As it is, the count shows # online and # offline, and no mention of # not activated.

ezequielc:
@tejohnso If _connectionStatusChangeDate_ is different from messaging service, I think it would be a good idea to run a job and update so that they are consistent in both places. 
But having said that, I am inclined to wait and get Alexey's feedback too, as he has more experience around this and could have better insights. Thoughts? Thanks!

ezequielc:
@tejohnso We can revisit the decision but we asked last week and Byron said that `not_activated` should not count as offline in the summary.
But also, if the status is not accurate, sorting by display status would be incorrect as well, right?

tejohnso:
> @tejohnso We can revisit the decision but we asked last week and Byron said that `not_activated` should not count as offline in the summary.

Oh, okay nevermind then :)

> But also, if the status is not accurate, sorting by display status would be incorrect as well, right?

Good point. Seems we'll have the fix the data / run a job.

tejohnso:
> @tejohnso If _connectionStatusChangeDate_ is different from messaging service, I think it would be a good idea to run a job and update so that they are consistent in both places.
> But having said that, I am inclined to wait and get Alexey's feedback too, as he has more experience around this and could have better insights. Thoughts? Thanks!

Agreed.

PR #2120, opened 2020-10-13
Bulk Set Address

ezequielc:
## Description
Bulk Set Address implementation

## Motivation and Context
Network Management epic

## How Has This Been Tested?
Locally
[stage-14]

## Release Plan:

### DIFF
--- a/web/scripts/displays/services/svc-display-list-operations.js
+++ b/web/scripts/displays/services/svc-display-list-operations.js
@@ -93,6 +93,14 @@ angular.module('risevision.displays.services')
           return _confirmDisplayUpdate(selectedItems, 'Set Address', 
             'partials/displays/edit-address.html', {
               useCompanyAddress: false,
+              addressDescription: '',
+              street: '',
+              unit: '',
+              city: '',
+              country: '',
+              province: '',
+              postalCode: '',
+              timeZoneOffset: null
           });
         };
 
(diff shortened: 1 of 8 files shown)


web/scripts/displays/services/svc-display-list-operations.js
```diff
@@ -89,6 +89,21 @@ angular.module('risevision.displays.services')
           });
         };
 
+        var _confirmSetAddress = function(selectedItems) {
+          return _confirmDisplayUpdate(selectedItems, 'Set Address', 
+            'partials/displays/edit-address.html', {
+              useCompanyAddress: false,
+              addressDescription: '',
```
alex-deaconu, line 96: @ezequielc do we need any of the fields other than `useCompanyAddress`? Thanks.

alex-deaconu (approved):

web/scripts/displays/services/svc-display-list-operations.js
ezequielc: @alex-deaconu We do. If I don't set them they will not be added to the base model and will not update the display, so the display would keep its previous value.
For example, if a display has a unit number and and leave blank on Set Address, it would not set blank unit to the display. Thanks!

alex-deaconu: @ezequielc makes sense, good catch! Thanks.

PR #2217, opened 2020-11-17
Check scroll event target before loading list

alex-deaconu:
## Description
Check scroll event target before loading list

Fixes issue where events from multiple scrolling-lists
overlap and cause reload

[stage-15]

## Motivation and Context
Fix issue with multiple lists.

## How Has This Been Tested?
Tested with local version where there are two scrolling lists on the billing page. Added unit test coverage.

## Release Plan:

No.

### DIFF
--- a/test/unit/components/scrolling-list/directives/dtv-scrolling-list.tests.js
+++ b/test/unit/components/scrolling-list/directives/dtv-scrolling-list.tests.js
@@ -1,0 +1,93 @@
+'use strict';
+describe('directive: scrolling-list', function() {
+  var $scope,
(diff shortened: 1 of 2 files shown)


alex-deaconu:
Fix is pretty straight forward. I was able to replicate the issue on my current branch with the Subscription and Invoice lists. I'm not sure why the event handlers (which are attached to specific elements) trigger - I think it may be because the whole page scrolls. Either way, this fixes the issue and the second list does not reload when scrolling the first.

ezequielc (approved):

PR #2247, opened 2020-11-26
Fix/issue 2241 help button failsafe

santiagonoguez:
## Description
Open help.risevision.com in a new browser tab, if the elev script doesn't load.

I added a 2 second wait to account for a user that immediately clicks the button while the page is still loading; but I can change ( or eliminate ) that delay if considered necessary.

## Motivation and Context
Issue 2241

## How Has This Been Tested?
Created unit tests as part of this PR, and tested manually.

As we don't have control on the third party code, I created a test branch that always opens the help browser tab: https://apps-stage-9.risevision.com/editor

This particular scenario can be manually tested in stage-9

## Release Plan:
To be released on Monday.

No need to update documentation

### DIFF
--- a/test/unit/common-header/services/svc-help-widget-factory-spec.js
+++ b/test/unit/common-header/services/svc-help-widget-factory-spec.js
@@ -80,37 +80,6 @@ describe("Services: helpWidgetFactory", function() {
       $window.document.createElement.should.not.have.been.called;
       $window._elev.on.should.not.have.been.called;
     });
-
-    it("should not open failback help screen if help script was loaded", function(done) {
-      factory.initializeWidget();
-      $window._elev.on.should.have.been.calledWith('load');
-
-      var loadCallback = $window._elev.on.getCall(0).args[1];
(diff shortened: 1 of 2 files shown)


ezequielc:
@santiagonoguez I managed to replicated the issue by blocking requests to `elev.io` from Chrome DevTools. When testing it, it throws an error  `_elev.openHome is not a function`. From what I saw, window._elev object is always present, even if the script fails to load. So I think we need to change `if ($window._elev)` to use your `widgetScriptLoaded` flag instead. 
Also, I don't think we need to add the 2 seconds delay. Thanks!

santiagonoguez:
@ezequielc I'll apply those changes, thanks !

santiagonoguez:
@ezequielc I've updated the code with your suggestions, and it's working ok in stage-9. Thanks a lot for the devtools hint, it works like a charm :D

ezequielc (approved):
@santiagonoguez Looks great 👍 Thanks for updating it!

Issue #2265, opened 2020-12-03
Search not finding relevant templates


osmankarame:
**Reproduction Steps**
1. On the presentation page, I click on **Add From template**
and by example I look for Human rights, I write human ri or right, it doesn't found the template called Human Rights

**Expected Results**
I expect to see the templates before having to complete the word. or adding a "s".
https://www.screencast.com/t/omBbtOgk

**Actual Results**
Not finding any templates.
https://www.screencast.com/t/ElH2M2Mez61j

ezequielc:
@alanclayton @alexey-rise @tejohnso This came up in the past, our search mechanism does not match partial words, only the full word. I believe it is a limitation of Google Search API and considering we want to move away from it, I think we should either close or block this defect until we migrate. Thanks

alanclayton:
Thanks @ezequielc I recall the same. @tejohnso can you confirm?

alexey-rise:
@ezequielc @alanclayton Google Search API does not support partial matches but that is not a limitation of Google Search API, rather it is a "feature" of a full-text search. If we migrate to other full-text search mechanism we will have the same problem.

If we want to support partial matches in a full-text search system we need to tokenize all text for all possible (or all desirable) substrings and add them to the index (e.g. "hello world" -> "he,hel,hell,hello,wo,wor,worl,world").

tejohnso:
I think Elasticsearch would allow partial matches via [wildcard query](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-wildcard-query.html)

alex-deaconu:
We could move the search to the client side since we retrieve the whole list anyways (no paging). It would be instantaneous and we would support partial word search. Thanks.

alanclayton:
@tejohnso @ezequielc @alexey-rise @alex-deaconu what is the conclusion? is this something we can/should pursue right now?

ezequielc:
We talked, we are going to proceed with Alex's suggestion. Thanks

alex-deaconu:
As discussed we have implemented client side search in Apps, which should resolve the partial word search issues. Closing. Thanks.

PR #2272, opened 2020-12-07
Use client side search for store products

alex-deaconu:
## Description
Use client side search for store products

Disable server side search by mocking function
Add search string as filter
Applies to both Templates and Widgets search

[stage-19]

## Motivation and Context
Fix for #2265 

## How Has This Been Tested?
Tested changes locally.

## Release Plan:

No

### DIFF
--- a/web/partials/editor/store-products-modal.html
+++ b/web/partials/editor/store-products-modal.html
@@ -8,12 +8,12 @@
   <div class="modal-body u_padding-lg" stop-event="touchend">
     
     <div class="u_margin-md-bottom" ng-if="!categoryFilters">
-      <search-filter filter-config="filterConfig" search="search" do-search="factory.doSearch"></search-filter>
+      <search-filter filter-config="filterConfig" search="search" do-search="search.doSearch"></search-filter>
     </div>
 
     <div class="row" ng-if="categoryFilters">
       <div class="col-md-8 u_margin-sm-bottom">
-        <search-filter filter-config="filterConfig" search="search" do-search="factory.doSearch"></search-filter>
+        <search-filter filter-config="filterConfig" search="search" do-search="search.doSearch"></search-filter>
       </div>
       <div class="col-md-4 u_margin-md-bottom aligner">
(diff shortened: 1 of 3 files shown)


alex-deaconu:
Looked at the server code and we do search by all fields. I did some comparisons and the results seem to be the same when the full search term is entered. Of course, partial search is now working.

@ezequielc @santiagonoguez please review. Thanks!

web/partials/editor/store-products-modal.html
```diff
@@ -44,7 +44,7 @@ <h3 id="storeModalTitle" class="modal-title">{{ 'editor-app.storeProduct.' + (fa
 
               <ng-include ng-if="!isEducationCustomer" src="'partials/editor/blank-presentation.html'"></ng-include>
 
-              <ng-repeat ng-repeat="product in factory.items.list | filter: getTemplatesFilter()">
+              <ng-repeat ng-repeat="product in factory.items.list | filter:search.query | filter: getTemplatesFilter()">
```
santiagonoguez, line 47: @alex-deaconu curious. is this search.query filter already defined somewhere ?

alex-deaconu: @santiagonoguez `search.query` is the value bound to the search text box ([link](https://github.com/Rise-Vision/rise-vision-apps/blob/master/web/partials/components/search-filter/search-filter.html#L7)). Thanks.

santiagonoguez: oh, I see ! I didn't know one could filter like that in Angular, nice :D

santiagonoguez (approved):
@alex-deaconu looks good, although for me in the new implementation entering or deleting text sometimes feels a little sluggish; I don't experience that in apps version. Do you or @ezequielc also see this, and if so should we worry about that ?

ezequielc:
@santiagonoguez I see it a little slow sometimes, but it is barely noticeable in my case. I think we can go as is 👍 Thanks!

alex-deaconu:
@ezequielc @santiagonoguez most likely caused by the digest cycle filtering through the huge list. I'm not sure we can do much about that. Merging. Thanks.

PR #2277, opened 2020-12-08
Chore/e2e unauthenticated invoice access

santiagonoguez:
## Description
Adds E2E tests to invoice page. 

Similar, access to unpaid invoices pages will be implemented in next PR.

## Motivation and Context
E2E coverage for unauthenticated access.

## How Has This Been Tested?
This is a test.

## Release Plan:
No runtime changes.


ezequielc:
@santiagonoguez @alex-deaconu My initial idea was to open the Unpaid Invoices list directly, without navigating to Billing, and then navigate to an unpaid invoice to test that page.
But on the other hand, we don't have any other tests covering Billing page, so I think we can use this for future tests as well, like Subscriptions list.

My main concern is about the Unpaid Invoice item in the list to move below the visible/scrollable area and the test fail in the future. But since we no longer purchase new subscriptions from this account, it should be fine. Let me know your thoughts. Thanks!

santiagonoguez:
@ezequielc  oh, I see your point. 

What about leaving this test just before checking the existence of the unpaid invoice, so we can check the billing page. And then I create another case for the flow you were thinking of where I check both unauthenticated access using the fixed company id and token. That way the tests won't be dependent on new active subscriptions being added. Thoughts ?

ezequielc:
@santiagonoguez I am good with your suggestion. Thanks!

santiagonoguez:
@ezequielc I left now just the basic billing page now, please take a look.

I'll add the unauthenticated page tests as additional PRs on top of this one. Thanks !

ezequielc (approved):

PR #2291, opened 2020-12-11
Global catcher for 401 gApi errors

ezequielc:
## Description
Wraps all gApi requests and handles 401 status code.

## Motivation and Context
Eliminate 3rd party cookies 

## How Has This Been Tested?
Locally and on [stage-1] by rewriting requests to return 401 errors.

## Release Plan:


ezequielc:
@santiagonoguez @alex-deaconu Please review and let me know if you agree with this approach. Thanks!

web/scripts/components/gapi-loader/svc-gapi.js
```diff
@@ -94,6 +94,33 @@ angular.module('risevision.common.gapi', [
...
                 .then(function () {
                   if (gApi.client[libName]) {
+
+                    var _addGlobalCatcher = function(library) {
+
+                      var _wrapWithCatcher = function (service, func) {
+                        var originalFunc = service[func];
+                        service[func] = function () {
+                          var ret = originalFunc.apply(service, arguments);
+                          ret.then(null, function(resp) {
```
santiagonoguez, line 104: @ezequielc curious, are we certain that the return type is always a Promise, so we can call `then()` over it ?

ezequielc: @santiagonoguez Yes, unless there is something wrong in my logic, they should all be endpoint requests, so they will have `.then()` available. Thanks!

santiagonoguez: maybe adding a protection like ret && ret.then && ret.then(null, ... ), could protect this against unpleasant surprises, I'm not sure. thoughts ?

santiagonoguez (commented):
Speaking of the whole idea, I really don't know if there's a better alternative, so I'm basically ok with implementing the error catcher as this. I'd like to see what @alex-deaconu thinks about this PR, though. 

Thanks !

ezequielc:
@santiagonoguez Good suggestion on adding the protection/verification. Will do that. Thanks!

web/scripts/components/gapi-loader/svc-gapi.js
```diff
@@ -94,6 +94,33 @@ angular.module('risevision.common.gapi', [
...
+                          });
+                          return ret;
+                        };
+                      };
+
+                      _.map(library, function(val,key) {
+                          if (typeof library[key] === 'function') {
+                            _wrapWithCatcher(library,key);
+                          } else if (typeof library[key] === 'object') {
+                            _addGlobalCatcher(library[key]);
```
alex-deaconu, line 117: @ezequielc do you think there's any chance of a circular reference? ie. do we need a catcher for the recursion level for this? Thanks.

web/scripts/components/gapi-loader/svc-gapi.js
```diff
@@ -94,6 +94,33 @@ angular.module('risevision.common.gapi', [
...
+
+                    var _addGlobalCatcher = function(library) {
+
+                      var _wrapWithCatcher = function (service, func) {
+                        var originalFunc = service[func];
+                        service[func] = function () {
+                          var ret = originalFunc.apply(service, arguments);
+                          ret.then(null, function(resp) {
+                            if (resp && resp.status === 401) {
+                              $rootScope.$broadcast('risevision.gapi.unauthorized');
```
alex-deaconu, line 106: @ezequielc curious, why not call `userAuthFactory.signOut()`? Thanks.

alex-deaconu: Disregard!

web/scripts/app.js
```diff
@@ -152,8 +152,21 @@ angular.module('risevision.apps', [
...
+
+      $rootScope.$on('risevision.gapi.unauthorized', function () {
+        if(userState.isLoggedIn()) {
+          messageBox('Authentication failed!',
+            'You have been disconnected. Please sign in again.',
+            'Ok', 'madero-style centered-modal', 'partials/template-editor/message-box.html','sm'
+          ).finally(function () {
+            //clear authentication
+            userAuthFactory.signOut();
+            $state.go('common.auth.unauthorized');
```
alex-deaconu, line 166: @ezequielc is the `$state.go` necessary here? As per below the `risevision.user.signedOut` event handler already redirects to the same page.

Also, why `signOut` after the `messageBox` result and not right away?

Thanks.

alex-deaconu: @ezequielc also, I would suggest somehow saving the state via the URL token so we can return to the same page when they sign back in. Thanks.

alex-deaconu: @ezequielc curious, did you test what happens if we have unsaved changes in the Template and Legacy editors while attempting this redirect? Thanks.

alex-deaconu:
@ezequielc some thoughts and suggestions inline. Sorry for the multiple comments. Thanks.

web/scripts/components/gapi-loader/svc-gapi.js
ezequielc: @alex-deaconu I didn't see that happening in my tests, but given it is an external code, I agree it is a good idea to have some safeguards here. I will try to add that. Thanks

web/scripts/app.js
ezequielc: @alex-deaconu Good point, I will remove `$state.go('common.auth.unauthorized');`. I added the `signOut` after the `messageBox` mainly to show the message in the same page where the user was, instead of immediately redirecting. I think it is more user friendly this way, but let me know if you disagree. 

I haven't tested with Unsaved Changes modals. I will double check and also try to save the url to redirect after signing in. Thanks!

ezequielc:
@alex-deaconu @santiagonoguez Given the errors we were seeing when wrapping gApi functions (duplicated calls, execute() vs then()), I implemented the approach suggested by Alex to detect it in `proccessErrorCode()`.

It works, but I am not happy with the end result for two main reasons: 
1- The main idea was to have a global catcher, to catch even requests not passed to `proccessErrorCode()`, and this is not achieved. On top of that , in the message returned by `proccessErrorCode` we tell the user to reload the page, defeating the purpose of the retry mechanism.
2- The way we present error messages is not consistent. In the scrollable lists we show it inline while in the Editors we show as a popup. When detecting the 401 error, this implementation shows a popup before redirecting to sign in page, causing duplicated popups.

Based on the above and given we already tell users to reload the page, my suggestion is to drop this idea and close this PR. Let me know if you agree. Thanks!

santiagonoguez:
I agree with your points, not a comfortable situation :O

I'm just wondering if it may still be useful to log the cases where there's a 401 error and userState.isLoggedIn() returns true, to at least see how much this condition affects users.

alex-deaconu:
@ezequielc @santiagonoguez I'm good with closing and pursuing at a later date. Thanks.

PR #2293, opened 2020-12-11
Remove 3rd party cookies check and warning message

ezequielc:
## Description
Removes cookies check and warning message.

## Motivation and Context
Apps now works with 3rd party cookies disabled.

## How Has This Been Tested?
Confirm the message is not present with 3rd party cookies disabled [stage-1]

## Release Plan:


ezequielc:
@alex-deaconu @santiagonoguez I removed the whole cookies checking but now I am not sure if it was the best decision. Apps doesn't even load when all cookies are disabled so the message would never appear. But I guess a better approach would have been to try to fix and make it load. Let me know your thoughts. Thanks!

ezequielc:
@alex-deaconu @santiagonoguez  I investigated the issue a little further and it is trigger because window.localStorage is (surprisingly) not present when Chrome is set to disable all cookies. OAuth.io lib (for Twitter authentication) and now oidc-client try to access it and it generates the error.
I noticed that Google doesn't allow signing in on that situation, it shows a warning asking users to enable cookies, and other sites like Trello, also fail to load. I don't think the fix is going to be trivial as they are internal to the libraries, so I am inclined to leave this as is for now. I will have a low priority task and we can take a better look at the end of the epic if we have time. Thoughts? Thanks!

ezequielc:
@santiagonoguez @alex-deaconu As per our chat, I will close this PR and open a new one removing only the third party check. Thanks!

PR #2294, opened 2020-12-14
Feature/send openid bq errors

santiagonoguez:
## Description
Send silent renew errors to BQ. 

[stage-9]

## Motivation and Context
Observability

## How Has This Been Tested?
Tested in stage-9, and then checked errors directly on BQ:

bq_openid_error

Unit tests added.

## Release Plan:
Part of prototype

### DIFF
--- a/web/scripts/components/logging/svc-openid-tracker.js
+++ b/web/scripts/components/logging/svc-openid-tracker.js
@@ -22,7 +22,7 @@ angular.module('risevision.common.components.logging')
           analyticsFactory.track('OpenId Event', eventProperties);
 
           if (openidEventType === 'silent renew error') {
-            bigQueryLogging.logEvent('OpenId silent renew error', eventProperties.errorMessage, null, email);
+            bigQueryLogging.logEvent('OpenId silent renew error', eventProperties.errorMessage, null, userId || email);
           }
         }
       };
--- a/web/scripts/components/logging/svc-openid-tracker.js
+++ b/web/scripts/components/logging/svc-openid-tracker.js
@@ -1,0 +1,21 @@
+'use strict';
+
+angular.module('risevision.common.components.logging')
+  .factory('openidTracker', ['userState', 'analyticsFactory',
+    function (userState, analyticsFactory) {
+      return function (openidEventType, eventProperties) {
+        if (openidEventType) {
+          eventProperties = eventProperties || {};
+          angular.extend(eventProperties, {
+            openidEventType: openidEventType,
+            userId: userState.getUsername(),
+            email: userState.getUserEmail(),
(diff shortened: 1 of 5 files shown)


santiagonoguez:
@ezequielc @alex-deaconu please don't forget this one, thanks ! :D

web/scripts/components/logging/svc-openid-tracker.js
```diff
@@ -1,22 +1,29 @@
...
+            userId: userId,
+            email: email,
             googleUserId: profile.sub,
             companyId: userState.getSelectedCompanyId()
           });
 
           analyticsFactory.track('OpenId Event', eventProperties);
+
+          if (openidEventType === 'silent renew error') {
+            bigQueryLogging.logEvent('OpenId silent renew error', eventProperties.errorMessage, null, email);
```
alex-deaconu, line 25: @santiagonoguez curious why log the email here instead of the username? Thanks.

web/scripts/components/logging/svc-openid-tracker.js
```diff
@@ -1,22 +1,29 @@
...
-            email: userState.getUserEmail() || profile.email,
+            userId: userId,
+            email: email,
             googleUserId: profile.sub,
             companyId: userState.getSelectedCompanyId()
           });
 
           analyticsFactory.track('OpenId Event', eventProperties);
+
+          if (openidEventType === 'silent renew error') {
```
alex-deaconu, line 24: @santiagonoguez we have a BQ logging pattern based on a constant here:
https://github.com/Rise-Vision/rise-vision-apps/blob/master/web/scripts/components/logging/svc-user-tracker.js#L17.

Not sure if you want to follow it. Thanks.

santiagonoguez: @alex-deaconu I saw that, and I was wondering if I would. As those other classes decide on main event type, and this decides on another field I was not sure. But I'll follow your suggestion and do the change so the code is similar to the others, thanks !

web/scripts/components/logging/svc-openid-tracker.js
santiagonoguez: @alex-deaconu I was trying to provide mail as a fallback in case the username couldn't be read here: https://github.com/Rise-Vision/rise-vision-apps/blob/master/web/scripts/components/logging/svc-big-query-logging.js#L11

But now I realize that the email won't be a fallback but the primary source for the username, so I think I'll change this call so username is always the primary source for this value. Thanks !

santiagonoguez:
@alex-deaconu your suggestions are applied now, thanks !

alex-deaconu (approved):
@santiagoarriaga looks good, thanks for the update!

PR #2310, opened 2020-12-18
add authenticate call to promise chain [stage-9]

santiagonoguez:
## Description
Apps doesn't ( always ) redirect to the requested page after google authentication. 

[stage-9]

## Motivation and Context
This was working ok before the changes in latest epic, and works ok for custom authentication. So it's expected behavior.

The issue is a race condition between the redirection after authentication:
https://github.com/Rise-Vision/rise-vision-apps/blob/master/web/scripts/components/userstate/services/svc-google-auth-factory.js#L65

And a redirection to the root page that happens when the hash is cleared here:
https://github.com/Rise-Vision/rise-vision-apps/blob/master/web/scripts/components/userstate/app.js#L58

This happens because the authenticate call goes out of the promise chain here:
https://github.com/Rise-Vision/rise-vision-apps/blob/master/web/scripts/components/userstate/app.js#L55

And then usually one redirection conflicts with the other; not every time with the same results, but I saw that usually the redirection to root page ends up winning and being showed to the user.

This solution simply adds the authenticate logic to the promise chain, so when there's a redirection state it gets applied to the $location.url().resolve() call, and then the hash is stripped out it doesn't redirect to the root page in this case.

## How Has This Been Tested?
Tested both locally and on stage-9. Different authentication scenarios were tested.

## Release Plan:
To be released on Monday. Simple rollback is enough in case of issues.

No need to notify support. No need to update documentation.

### DIFF
--- a/web/scripts/components/userstate/app.js
+++ b/web/scripts/components/userstate/app.js
@@ -52,7 +52,7 @@
 
                 openidConnect.signinRedirectCallback()
                   .then(function(user) {
-                    userAuthFactory.authenticate(true);
+                    return userAuthFactory.authenticate(true);
                   })
                   .finally(function() {
                     window.location.hash = '';


ezequielc (approved):

alex-deaconu:
@santiagonoguez this fix does not work for me. The issue is actually caused by this:
https://github.com/Rise-Vision/rise-vision-apps/blob/master/web/scripts/app.js#L191.
For the state reload, `$state.current` is `apps.home`. Maybe adding a return on the original promise (`openidConnect.signinRedirectCallback()`) is required too?
Thanks.

santiagonoguez:
@alex-deaconu just a quick question / clarification, did you test locally or in stage-9 ?  I think I rewrote stage-9 with other branches later so it's no longer having this behavior ( shouldn't do that, so it could still be tested there ).

alex-deaconu:
@santiagonoguez that was the issue indeed. Fix works for me as intended. Thanks!

alex-deaconu (approved):

PR #2342, opened 2021-01-06
Update preview URL to widgets.risevision.com/viewer [stage-3]

olegrise:
## Description
Update preview URL to widgets.risevision.com/viewer

## Motivation and Context
Origin consolidation epic. Viewer's URL changed.

## How Has This Been Tested?
Tested by running Apps on local machine. Tested Preview for Presentations and URL for Shared Schedule.

## Release Plan:

### DIFF
--- a/web/scripts/config/prod.js
+++ b/web/scripts/config/prod.js
@@ -23,8 +23,8 @@
     .value('STORE_ENDPOINT_URL',
       'https://store-dot-rvaserver2.appspot.com/_ah/api')
     .value('RVA_URL', 'http://rva.risevision.com')
-    .value('VIEWER_URL', 'https://preview.risevision.com')
-    .value('SHARED_SCHEDULE_URL', 'https://preview.risevision.com/?type=sharedschedule&id=SCHEDULE_ID')
+    .value('VIEWER_URL', 'https://widgets.risevision.com/viewer')
+    .value('SHARED_SCHEDULE_URL', 'https://widgets.risevision.com/viewer/?type=sharedschedule&id=SCHEDULE_ID')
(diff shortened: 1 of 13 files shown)


web/scripts/config/prod.js
```diff
@@ -23,8 +23,8 @@
     .value('STORE_ENDPOINT_URL',
       'https://store-dot-rvaserver2.appspot.com/_ah/api')
     .value('RVA_URL', 'http://rva.risevision.com')
-    .value('VIEWER_URL', 'https://preview.risevision.com')
-    .value('SHARED_SCHEDULE_URL', 'https://preview.risevision.com/?type=sharedschedule&id=SCHEDULE_ID')
+    .value('VIEWER_URL', 'https://widgets.risevision.com/viewer')
```
stulees, line 26: @olegrise I'm wondering if it's safer to add the `/` to the end as we know Viewer doesn't load properly unless it has `/` (and update in all of these config files). The only place in this code base where `VIEWER_URL` is used is https://github.com/Rise-Vision/rise-vision-apps/blob/master/web/scripts/editor/services/svc-editor-factory.js#L488 . We would need to remove the `/` there. Thoughts?

olegrise: @stulees I see what you mean. I don't feel strongly that one way is better than other. Considering that there is no place in the code where `VIEWER_URL` is used as full URL, I don't think it matters. Do you feel strongly that this change is needed?

stulees: @olegrise I don't feel strongly. We can keep as is, it is unlikely the value will ever be used as full url. Thanks

stulees (approved):

PR #2365, opened 2021-01-13
Chore/join purchase controller

santiagonoguez:
## Description
Unify code for add and remove controller ( as it was originally )

[stage-9] 

## Motivation and Context
Both controllers ended up being almost identical, as special validation logic wasn't required for removal and the rest of the distinctions won't be part of controller. So a parameter was introduced to account for minor differences ( it will be used to signal service to perform subtraction rather than addition for removal - next PR ).

## How Has This Been Tested?
Tested locally and in stage-9.
Unit tests were updated.

## Release Plan:
To be merged with parent branch.

-


web/scripts/purchase/controllers/ctr-purchase-licenses.js
```diff
@@ -14,9 +14,9 @@ angular.module('risevision.apps.purchase')
 
       $scope.$watch('factory.loading', function (loading) {
         if (loading) {
-          $loading.start('add-licenses-loader');
+          $loading.start($stateParams.purchaseAction + '-licenses-loader');
         } else {
-          $loading.stop('add-licenses-loader');
+          $loading.stop($stateParams.purchaseAction + '-licenses-loader');
```
alex-deaconu, line 19: @santiagonoguez I'm ok with having the same spinner name for both to avoid all this logic. Thanks.

alex-deaconu (approved):
@santiagonoguez looks good, but I don't think we need to add an extra parameter for the action just for the spinner, we can use the same key. Thanks.

web/scripts/purchase/controllers/ctr-purchase-licenses.js
santiagonoguez: good, I'll use the same name then, thanks !

web/scripts/purchase/app.js
```diff
@@ -89,7 +89,10 @@ angular.module('risevision.apps')
           templateProvider: ['$templateCache', function ($templateCache) {
             return $templateCache.get('partials/purchase/add-licenses.html');
           }],
-          controller: 'AddLicensesCtrl',
+          params: {
+            purchaseAction: 'add'
```
alex-deaconu, line 93: @santiagonoguez this variable is still used? Should it not have been cleaned up in the last commit? Thanks.

santiagonoguez: @alex-deaconu yes, I'm using it to define estimate behavior in the next PR.

alex-deaconu: @santiagonoguez okay, thanks.

PR #2368, opened 2021-01-13
Feature/estimate removal

santiagonoguez:
## Description
- Update purchase licenses factory to subtract displayCount when removing.
- Add function to add credit notes to single total.
- Show 3 totals in partial. 
- Estimate and update now working, but update success partial still doesn't exist ( next PR ).

[stage-9]

## Motivation and Context
Part of epic.

## How Has This Been Tested?
Locally and in stage-9.
Automated unit tests updated / created
To test 3 totals try to remove all displays minus one in some subscriptions.

remove_form

## Release Plan:
To be merged to parent branch

-

### DIFF
--- a/web/partials/purchase/remove-licenses.html
+++ b/web/partials/purchase/remove-licenses.html
@@ -31,24 +31,24 @@
 
         <div class="pt-4">
           <p id="prorated-amount-row" class="left-right-aligner mb-4" ng-if="factory.estimate.invoice_estimate">
-            <span class="font-weight-bold">Generated invoice, due now:</span>
+            <span class="font-weight-bold">Generated invoice:</span>
             <span>
               <span class="u_margin-right text-subtle">{{factory.estimate.invoice_estimate.currency_code}}</span>
-              <span class="purchase-total">${{factory.estimate.invoice_estimate.amount_due/100 | number:2}}</span>
+              <span class="purchase-total">${{factory.estimate.invoice_estimate.total/100 | number:2}}</span>
             </span>
           </p>
-          <p id="prorated-amount-row" class="left-right-aligner mb-4">
+          <p id="prorated-credit-row" class="left-right-aligner mb-4">
             <span class="font-weight-bold">Prorated credit added to your account:</span>
             <span>
               <span class="u_margin-right text-subtle">{{factory.estimate.credit_note_estimates[0].currency_code}}</span>
-              <span class="purchase-total">$--</span>
+              <span class="purchase-total">${{factory.getCreditTotal() | number:2}}</span>
             </span>
           </p>
           <p id="next-invoice-row" class="left-right-aligner mb-0">
             <span class="font-weight-bold">Next invoice on {{factory.estimate.subscription_estimate.next_billing_at * 1000 | date:'d-MMM-yyyy'}}:</span>
             <span>
               <span class="u_margin-right text-subtle">{{factory.estimate.next_invoice_estimate.currency_code}}</span>
-              <span class="purchase-total">${{factory.estimate.next_invoice_estimate.amount_due/100 | number:2}}</span>
+              <span class="purchase-total">${{factory.estimate.next_invoice_estimate.total/100 | number:2}}</span>
             </span>
           </p>
         </div>
--- a/web/partials/purchase/remove-licenses.html
+++ b/web/partials/purchase/remove-licenses.html
@@ -30,10 +30,17 @@
         </div>
 
         <div class="pt-4">
+          <p id="prorated-amount-row" class="left-right-aligner mb-4" ng-if="factory.estimate.invoice_estimate">
+            <span class="font-weight-bold">Generated invoice, due now:</span>
+            <span>
+              <span class="u_margin-right text-subtle">{{factory.estimate.invoice_estimate.currency_code}}</span>
+              <span class="purchase-total">${{factory.estimate.invoice_estimate.amount_due/100 | number:2}}</span>
+            </span>
+          </p>
           <p id="prorated-amount-row" class="left-right-aligner mb-4">
             <span class="font-weight-bold">Prorated credit added to your account:</span>
             <span>
-              <span class="u_margin-right text-subtle">{{factory.estimate.invoice_estimate.currency_code}}</span>
+              <span class="u_margin-right text-subtle">{{factory.estimate.credit_note_estimates[0].currency_code}}</span>
               <span class="purchase-total">$--</span>
             </span>
           </p>
--- a/web/partials/purchase/remove-licenses.html
+++ b/web/partials/purchase/remove-licenses.html
@@ -16,15 +16,15 @@
         <h4 class="u_margin-md-bottom mt-0 mb-4">Subscription Details</h4>
 
         <div class="left-right-aligner" ng-class="{'has-error': purchaseLicensesForm.$invalid }">
-          <label for="displayCount" class="control-label">Number of display licenses you want to remove:</label>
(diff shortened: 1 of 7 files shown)


web/partials/purchase/remove-licenses.html
```diff
@@ -30,18 +30,25 @@ <h4 class="u_margin-md-bottom mt-0 mb-4">Subscription Details</h4>
         </div>
 
         <div class="pt-4">
-          <p id="prorated-amount-row" class="left-right-aligner mb-4">
-            <span class="font-weight-bold">Prorated credit added to your account:</span>
+          <p id="prorated-amount-row" class="left-right-aligner mb-4" ng-if="factory.estimate.invoice_estimate">
+            <span class="font-weight-bold">Generated invoice:</span>
```
santiagonoguez, line 34: Not sure of the label here :O

web/partials/purchase/remove-licenses.html
```diff
@@ -30,18 +30,25 @@ <h4 class="u_margin-md-bottom mt-0 mb-4">Subscription Details</h4>
...
             <span>
               <span class="u_margin-right text-subtle">{{factory.estimate.invoice_estimate.currency_code}}</span>
-              <span class="purchase-total">$--</span>
+              <span class="purchase-total">${{factory.estimate.invoice_estimate.total/100 | number:2}}</span>
+            </span>
+          </p>
+          <p id="prorated-credit-row" class="left-right-aligner mb-4">
+            <span class="font-weight-bold">Prorated credit added to your account:</span>
+            <span>
+              <span class="u_margin-right text-subtle">{{factory.estimate.credit_note_estimates[0].currency_code}}</span>
```
santiagonoguez, line 43: I'm assuming all estimates share same currency

web/scripts/purchase/services/svc-purchase-licenses-factory.js
```diff
@@ -19,22 +19,34 @@
...
+
+          return currentLicenses + _getChangeInLicenses();
+        };
+
         var _getTrackingProperties = function () {
           return {
             subscriptionId: currentPlanFactory.currentPlan.subscriptionId,
-            changeInLicenses: factory.purchase.displayCount,
-            totalLicenses: factory.purchase.displayCount + currentPlanFactory.currentPlan.playerProTotalLicenseCount,
+            changeInLicenses: _getChangeInLicenses(),
```
santiagonoguez, line 48: this may show a positive or negative value depending on 'add' or 'remove' parameter. But another option is to always report positive and add another parameter indicating if it was add or remove.

web/scripts/purchase/services/svc-purchase-licenses-factory.js
```diff
@@ -19,22 +19,34 @@
...
           factory.purchase.displayCount = $stateParams.displayCount;
           factory.purchase.couponCode = '';
+          factory.purchase.action = purchaseAction;
 
           factory.getEstimate();
         };
 
+        var _getChangeInLicenses = function() {
+          return factory.purchase.action === 'remove' ?
+            -factory.purchase.displayCount : factory.purchase.displayCount;
```
alex-deaconu, line 36: @santiagonoguez wondering if it were easier to just use a different `ng-model` variable for each of the two text boxes (add and remove). This way it would be irrelevant to the factory which form is selected, and you can avoid the variable. Thanks.

santiagonoguez: @alex-deaconu so we always add one and subtract the other, and they always default to 0 ? 

yes, that could be a good way to avoid the variable. I'll implement that idea, thanks !

alex-deaconu (commented):
@santiagonoguez one refactoring suggestion inline. However, I'll defer the rest of the points to @ezequielc or a team discussion. Thanks.

web/scripts/purchase/services/svc-purchase-licenses-factory.js
```diff
@@ -19,22 +19,35 @@
...
+        factory.init = function (purchaseAction) {
           _clearMessages();
 
+          var isRemove = purchaseAction === 'remove';
+
           factory.purchase = {};
           factory.purchase.completed = false;
-          factory.purchase.displayCount = $stateParams.displayCount;
+          factory.purchase.licensesToAdd = isRemove ? 0 : $stateParams.displayCount;
+          factory.purchase.licensesToRemove = isRemove ? $stateParams.displayCount : 0;
```
alex-deaconu, line 30: @santiagonoguez you can use `ng-value` on the field to set the default value via `displayCount`. Thanks.

santiagonoguez: good, thanks !

alex-deaconu (approved):
@ezequielc looks good, just some notes inline, no need to address unless you want to. Thanks.

alex-deaconu:
@santiagonoguez sorry that was meant for the store-server PR from Ezequiel 🤕

web/scripts/purchase/services/svc-purchase-licenses-factory.js
santiagonoguez: @alex-deaconu I'm experimenting with ng-value, and it indeed sets the default value on the input field; but does not affect the model variable until it's updated by the user. I read a little about it and it seems that ng-value never affects the model. 

In our case this affects the initial estimation as it doesn't get the initial value for add or remove, so it doesn't get calculated right. Any thoughts ?

alex-deaconu: @santiagonoguez let's leave as is. Thanks.

alex-deaconu (approved):
@santiagonoguez looks good thanks for the updates and investigation!

web/scripts/purchase/services/svc-purchase-licenses-factory.js
santiagonoguez: @ezequielc I think I'll leave this for now, we can adjust later before releasing if needed :D

web/partials/purchase/remove-licenses.html
ezequielc: I think you should use the same wording as Add Licenses: `Prorated amount, due now:`.
Also, I think it makes more sense to change the order and show it after the credit, mainly because the credit will be used to pay this prorated amount.

So it would show:
```
Prorated credit added to your account: CAD $523.80
Prorated amount, due now: CAD $7.20
Next invoice on 7-Feb-2021: CAD $108.90
```

Chargebee is a bit more clear, but I think the above is good enough
```
An invoice for CAD 7.20 will be generated immediately.
Refundable credits worth CAD 523.80 will be created, out of which CAD 7.20 will be applied on the immediate invoice.
Subsequent invoice for CAD 108.90 will be generated upon renewal on 07-Feb-2021 14:54 EST.
```

@alex-deaconu @santiagonoguez Are you good with the above? Thanks!

web/scripts/purchase/services/svc-purchase-licenses-factory.js
ezequielc: @santiagoarriaga Yeah, that was the idea when I created the property, to be negative when removing licenses 👍

web/partials/purchase/remove-licenses.html
alex-deaconu: @ezequielc agreed. Thanks.

santiagonoguez: @ezequielc Sounds good, I'll apply these with next UI changes, thanks !

PR #2374, opened 2021-01-15
receive subscription id on remove licenses [stage-8]

santiagonoguez:


santiagonoguez:
@alex-deaconu @ezequielc I just need some confirmation that I'm moving in the right direction with this implementation.

My main concern is:
- if different states / URLs should be used for add license for default subscription and explicit subscription ( so we keep the current behavior with currentPlanFactory ),}
- or we now require always a subscription id to be specified for both add and remove ( so code is simplified and no conditional code is set ... and in this case current plan checks on abstract apps.purchase.licenses super state will still hold, I think ).

web/scripts/purchase/app.js
```diff
@@ -85,6 +85,8 @@ angular.module('risevision.apps')
           }
         })
         .state('apps.purchase.licenses.add', {
+          // TODO: either turn to always receive subscriptionId,
+          // or add additional '/add/:subscriptionId' state
           url: '/add',
```
alex-deaconu, line 90: @santiagonoguez we need the `subscriptionId` parameter for add as well. Thanks.

web/scripts/purchase/services/svc-purchase-licenses-factory.js
```diff
@@ -30,34 +31,48 @@
           factory.purchase.licensesToRemove = isRemove ? $stateParams.displayCount : 0;
           factory.purchase.couponCode = '';
 
-          factory.getEstimate();
+          if ($stateParams.subscriptionId) {
```
alex-deaconu, line 34: @santiagonoguez  I would check if the `$stateParams.subscriptionId` is available. If not, use the `currentPlan.subscriptionId`. And as we discussed, always load the subscription. There's no need to complicate things.

Also, loading the subscription should probably start the spinner in this factory.

Thanks.

web/scripts/purchase/services/svc-purchase-licenses-factory.js
```diff
@@ -30,34 +31,48 @@
...
+              factory.getEstimate();
+            });
+            // TODO: handle save subscription load errors, either here or on controller/partial
+          } else {
+            factory.getEstimate();
+          }
         };
 
         factory.getCurrentDisplayCount = function() {
+          // TODO: use subscriptionFactory or currentPlanFactory
```
alex-deaconu, line 46: @santiagonoguez as mentioned, always load the Subscription and use subscriptionFactory. Thanks.

alex-deaconu:
@santiagonoguez added some notes, thanks.

ezequielc:
@santiagonoguez  I agree with @alex-deaconu 's suggestions. Thanks!

PR #2381, opened 2021-01-19
Change payment method implementation

alex-deaconu:
## Description
Change payment method implementation

Show notification; merge object on success
Change invoice check to use correct field

Clean out error block and refresh code

[stage-19]

## Motivation and Context
Billing friction epic.

## How Has This Been Tested?
Tested changes locally. Updated unit tests.

## Release Plan:

No

### DIFF
--- a/web/partials/billing/subscription.html
+++ b/web/partials/billing/subscription.html
@@ -10,8 +10,8 @@
     <strong>{{subscriptionFactory.apiError}}</strong>
   </div>
 
-  <div id="errorBoxChargebee" ng-show="chargebeeFactory.apiError" class="madero-style alert alert-danger" role="alert">
-    <span>{{chargebeeFactory.apiError}}</span>
+  <div id="errorBox" ng-show="paymentSourcesFactory.apiError" class="madero-style alert alert-danger" role="alert">
+    <strong>{{paymentSourcesFactory.apiError}}</strong>
   </div>
 
   <h4 class="mb-4">
@@ -62,7 +62,7 @@
         <div class="panel-body">
           <label class="mb-3">Payment Method</label>
           <div class="madero-radio align-left" ng-repeat="card in paymentSourcesFactory.existingCreditCards">
-            <input type="radio" ng-change="setPaymentMethod(card.payment_source.id)" ng-model="subscriptionFactory.item.subscription.payment_source_id" ng-value="card.payment_source.id" name="{{card.payment_source.id}}" id="{{card.payment_source.id}}" tabindex="1">
+            <input type="radio" ng-click="subscriptionFactory.changePaymentMethod($event, subscriptionFactory.item.subscription, card)" ng-model="subscriptionFactory.item.subscription.payment_source_id" ng-value="card.payment_source.id" name="{{card.payment_source.id}}" id="{{card.payment_source.id}}" tabindex="1">
             <label for="{{card.payment_source.id}}">
               {{card.payment_source.card | cardDescription}}
               <span class="pull-right"><a class="madero-link u_clickable text-danger" ng-click="paymentSourcesFactory.removePaymentMethod(card)" ng-show="subscriptionFactory.item.subscription.payment_source_id !== card.payment_source.id">Remove</a></span>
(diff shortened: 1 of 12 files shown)


alex-deaconu:

alex-deaconu:

web/scripts/billing/services/svc-subscription-factory.js
```diff
@@ -30,10 +32,36 @@ angular.module('risevision.apps.billing.services')
...
-      factory.reloadSubscription = function () {
-        if (factory.item && factory.item.subscription && factory.item.subscription.id) {
-          factory.getSubscription(factory.item.subscription.id);
-        }
+      var _changePaymentSource = function (subscriptionId, paymentSourceId) {
+        factory.loading = true;
+
+        return billing.changePaymentSource(subscriptionId, paymentSourceId)
+          .then(function (resp) {
+            angular.extend(factory.item, resp.item);
```
ezequielc, line 40: @alex-deaconu Curious, what is the benefit of using `angular.extend` here instead of replacing the item? I am wondering if this could cause any conflicts in case the subscription gets updated elsewhere in between calls. Thanks

web/scripts/billing/controllers/ctr-subscription.js
```diff
@@ -27,7 +27,8 @@ angular.module('risevision.apps.billing.controllers')
       taxExemptionFactory.init();
 
       $scope.isInvoiced = function() {
```
ezequielc, line 29: @alex-deaconu I think we should check if subscription is `auto_collection:off` here instead. `payment_source_id` could be also be blank when the subscription relies on customer's default payment source. 
And in addition to that, customer could also be auto_collection:off, so you might need to check for both cases here. Thanks!
Thanks

web/partials/billing/subscription.html
```diff
@@ -62,7 +62,7 @@ <h1>${{subscriptionFactory.item.subscription.plan_amount / 100 | number:2}}</h1>
         <div class="panel-body">
           <label class="mb-3">Payment Method</label>
           <div class="madero-radio align-left" ng-repeat="card in paymentSourcesFactory.existingCreditCards">
-            <input type="radio" ng-change="setPaymentMethod(card.payment_source.id)" ng-model="subscriptionFactory.item.subscription.payment_source_id" ng-value="card.payment_source.id" name="{{card.payment_source.id}}" id="{{card.payment_source.id}}" tabindex="1">
+            <input type="radio" ng-click="subscriptionFactory.changePaymentMethod($event, subscriptionFactory.item.subscription, card)" ng-model="subscriptionFactory.item.subscription.payment_source_id" ng-value="card.payment_source.id" name="{{card.payment_source.id}}" id="{{card.payment_source.id}}" tabindex="1">
```
ezequielc, line 65: @alex-deaconu When I click on the selected radio, it triggers me to set the selected card as payment method. Maybe you should revert to ng-change here. Thanks!

alex-deaconu: @ezequielc good catch, I will try it with ng-change. Thanks!

web/scripts/billing/controllers/ctr-subscription.js
alex-deaconu: @ezequielc yeah this definitely needs some improving. Thanks.

web/scripts/billing/services/svc-subscription-factory.js
alex-deaconu: @ezequielc to confirm, the response has a `subscription` and `payment_method` object. If I overwrite I will lose the `company` object returned with the original `get` request. Alternatively to this, I could refresh the subscription but I was just saving an API call. Thoughts? Thanks.

ezequielc: @alex-deaconu Got it. I am good with this. Also, I just noticed that `angular.extend()` does a shallow copy, so factory.item.subscription will be entire replaced with the value from rest.item.subscription. Thanks!

alex-deaconu: @ezequielc yeah I don't think we need more than the shallow copy here. If we do have issues we can just refresh as mentioned. Thanks.

alex-deaconu:
@ezequielc the issues highlighted above have been resolved. I'd like to merge this to the main branch if acceptable. Thanks.

ezequielc (approved):
@alex-deaconu Sounds good 👍 Thanks for the updates

Issue #2493, opened 2021-03-30
[Add A Display] Clicking "Cancel" Doesn't Undo Edits After Previous Save


jasoncaldwell:
Please see this video:

https://www.loom.com/share/814da38880c14d6fb84f3ee661ae6c8f

ezequielc:
@jasoncaldwell This has been resolved. Thanks!

jasoncaldwell:
@ezequielc Thanks!

PR #2496, opened 2021-03-31
Toggle display license on Save

ezequielc:
## Description

- No longer license display immediately when toggling license buttons, instead, checks for license change and update it when user Saves the display changes.
- Removed license toggle local spinner and related variables
- Moved license update code to playerLicenseFactory

## Motivation and Context
Fix for #2493

## How Has This Been Tested?
Manually validated multiple use cases. 
Pending unit tests updates on subsequent PRs.
[stage-1]

## Release Plan:

### DIFF
--- a/src/scripts/displays/services/svc-display-factory.js
+++ b/src/scripts/displays/services/svc-display-factory.js
@@ -56,6 +56,7 @@ angular.module('risevision.displays.services')
         display.get(displayId)
           .then(function (result) {
             factory.display = result.item;
+            factory.display.originalPlayerProAuthorized = factory.display.playerProAuthorized;
 
             deferred.resolve();
           })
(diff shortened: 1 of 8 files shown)


ezequielc:
@alex-deaconu I would like your feedback on this before I move to unit tests. Thanks!

src/scripts/displays/services/svc-display-factory.js
```diff
@@ -56,6 +56,7 @@ angular.module('risevision.displays.services')
         display.get(displayId)
           .then(function (result) {
             factory.display = result.item;
+            factory.display.originalPlayerProAuthorized = factory.display.playerProAuthorized;
```
alex-deaconu, line 59: @ezequielc my first feedback is to not create an `original` variable, but instead use the original value and create a `playerProAuthorizedLocal` variable, so match with your `updateDisplayLicenseLocal` function. Thanks.

ezequielc: @alex-deaconu Your idea makes sense, the main problem is that `playerProAuthorized` is used in multiple places so it would be a riskier change. But let's discuss more tomorrow at standup. Thanks!

alex-deaconu (approved):
@ezequielc looks good, thanks for fixing 👍

PR #2544, opened 2021-04-19
Unique filename to upload canva design

ezequielc:
## Description
Use Canva design title and id to generate a unique name for each design.
Uploads to canva/ folder in Storage
In case of conflict, a message asking to overwrite or not is shown.
Also, refactored the download and created a fileDownloader service

## Motivation and Context
Canva Integration

## How Has This Been Tested?
Tested locally
[stage-1]

## Release Plan:

### DIFF
--- a/src/scripts/template-editor/directives/dtv-basic-uploader.js
+++ b/src/scripts/template-editor/directives/dtv-basic-uploader.js
@@ -17,7 +17,6 @@ angular.module('risevision.template-editor.directives')
         },
         templateUrl: 'partials/template-editor/basic-uploader.html',
         link: function ($scope, element) {
-          var confirmOverwriteModal;
           var FileUploader = fileUploaderFactory();
           var inputElement = element.find('input');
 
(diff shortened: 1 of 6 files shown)


src/scripts/template-editor/directives/dtv-basic-uploader.js
```diff
@@ -31,11 +30,8 @@ angular.module('risevision.template-editor.directives')
 
           $scope.$watch('fileList',function(selectedFiles) {
             if (selectedFiles) {
-              $scope.uploadSelectedFiles(selectedFiles).then(function() {
-                $scope.fileList = null;
```
alex-deaconu, line 35: @ezequielc curious, why do we no longer reset the file list? Thanks.

alex-deaconu (approved):
@ezequielc looks good, one question inline. Thanks.

src/scripts/template-editor/directives/dtv-basic-uploader.js
ezequielc: @alex-deaconu Sorry, I should have clarified this in my PR message. This code was incorrect. By setting `$scope.fileList = null;` it would set this reference to null, but not the source `$scope.canvaUploadList`. So it would fail to trigger a new call when creating a second Canva design, and the new design would never be uploaded. Thanks!

alex-deaconu: @ezequielc make sense, thanks for the clarification.

PR #2657, opened 2021-06-10
Migrate Stripe related services to Angular

ezequielc:
## Description
Migrate Stripe related services to Angular.

Refactored stripeLoader to use Promise and improved the logic for using prod vs test api keys.
Also added some suggestions for future improvements [here](https://docs.google.com/document/d/1Lha7Uf7kGjdhOb95k3V1jyW0xk3L6jYzvD-6PUf3MWg/edit#bookmark=id.nqtzwlg52v0u).

## Motivation and Context
Angular Migration

## How Has This Been Tested?
Locally and on [stage-1]

## Release Plan:


src/environments/environment.beta.ts
```diff
@@ -1,5 +1,5 @@
 export const environment = {
-  production: false,
+  production: true,
```
ezequielc, line 2: @alex-deaconu I want your feedback on this one. I set it to true so we can use the same stripe keys as in production. But I am not sure this is clear enough given the name is `production` and could be misleading in the future. I am considering setting an `env_name` variable and use that instead. Thoughts? Thanks

alex-deaconu: @ezequielc as long as setting that way doesn't break anything else, I think this is ok. Thanks.

ezequielc: @alex-deaconu Sounds good. I will leave as is for now. We can revisit in the future if we find it confusing. Thanks

alex-deaconu (approved):
@ezequielc looks good, nice work 👍

PR #2710, opened 2021-06-30
Use redirectTo for redirects

alex-deaconu:
## Description
Use redirectTo for redirects

Use state.params as they are more reliable

Remove bypass reset; route redirect calls handler twice

[stage-11]

## Motivation and Context
Simplify ui-router implementation. Prevent issues caused by the multiple redirects.

## How Has This Been Tested?
Tested locally. Updated unit tests.

## Release Plan:

No

### DIFF
--- a/src/scripts/app.module.ajs.js
+++ b/src/scripts/app.module.ajs.js
@@ -80,14 +80,7 @@ angular.module('risevision.apps', [
 
         .state('apps.home', {
           url: '/',
-          controller: ['$location', '$state', 'canAccessApps',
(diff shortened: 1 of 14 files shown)


alex-deaconu:
I had to modify the flag for the autosave, because the double redirect is causing 2 state changed events. This was happening before, but the resolve promises were making this asynchronous so the handler was not triggered twice.

I don't see a reason for re-enabling the flag anyways..

ezequielc (approved):

PR #2815, opened 2021-07-27
Viewport component for template editor routing

ezequielc:
## Description
Viewport component for template editor routing, to remove the `template` field from state since it is an AngularJS-based field.

## Motivation and Context
Angular migration

## How Has This Been Tested?
Locally and [stage-20]

## Release Plan:


src/app/template-editor/components/template-editor-preview-holder/template-editor-preview-holder.component.ts
```diff
@@ -104,14 +104,14 @@ export class TemplateEditorPreviewHolderComponent implements AfterViewInit, DoCh
     this.subscription.unsubscribe();
   }
 
-  onLoad() {
-    this.iframeLoaded = true;
+  onLoad(event) {
+    if (event.target.src != '') {
```
ezequielc, line 108: @alex-deaconu For some reason, onLoad started being called twice: once when iframe.src is still blank and once after editorPreviewUrl is populated. I could not pinpoint the reason so I am skipping the event for now. Thanks

alex-deaconu: @ezequielc from my findings I think it's expected that iframes send onLoad events when the src is set. Ideally we should be using an event from the template to initialize and send data, but I think this should work as well. Thanks.

alex-deaconu (approved):
@ezequielc 👍

PR #2826, opened 2021-07-28
Migrate Canva Type Picker

ezequielc:
## Description

- Migrate Canva Type Picker
- Created PromiseBasedModalComponent as an abstraction to mach angularjs modals pattern 

## Motivation and Context
Angular migration

## How Has This Been Tested?
Locally
[stage-20]

## Release Plan:


ezequielc:
@alex-deaconu Let me know if you are good with this abstract PromiseBasedModalComponent approach. If so, I will work on updating ConfirmModal to use it as well. Thanks

src/app/template-editor/components/canva-type-picker/canva-type-picker.component.ts
```diff
@@ -0,0 +1,22 @@
...
+  templateUrl: './canva-type-picker.component.html',
+  styleUrls: ['./canva-type-picker.component.scss']
+})
+export class CanvaTypePickerComponent extends PromiseBasedModalComponent {
+
+  designType = 'Presentation';
+
+  constructor(modalRef: BsModalRef) {
+    super(modalRef);
+  }
```
alex-deaconu, line 16: @ezequielc is this redundant or does it have to be called? Thanks.

src/app/components/modals/promise-based-modal.component.ts
```diff
@@ -0,0 +1,34 @@
...
+  
+  constructor(public modalRef: BsModalRef) {
+    this.promise = new Promise((resolve, reject) => {
+      this._resolve = resolve;
+      this._reject = reject;
+    });
+
+    this.modalRef.onHide.subscribe(() => {
+      if (this._resolvedValue) {
+        this._resolve(this._resolvedValue);
```
alex-deaconu, line 18: @ezequielc I'm not sure if this will work. We can call `resolve()` without a value for a positive user interaction 'OK' and this would reject. Maybe `resolve()` could directly resolve the promise? Any other ideas? Thanks.

src/app/components/modals/promise-based-modal.component.ts
```diff
@@ -0,0 +1,34 @@
+import { BsModalRef } from 'ngx-bootstrap/modal';
+
+export abstract class PromiseBasedModalComponent {
```
alex-deaconu, line 3: @ezequielc should we bother with such a complicated name? Maybe `ModalComponent` would do? Since we don't really need to use the promise if we don't want to it's just a feature. Thanks.

alex-deaconu (approved):
@ezequielc nice refactoring. I think it works very well. A few suggestions inline. Thanks!

src/app/template-editor/components/canva-type-picker/canva-type-picker.component.ts
ezequielc: @alex-deaconu This is required. It throws a build error if I don't add it because Angular's dependency injection is not able to inject based on the abstract class constructor. Thanks

src/app/components/modals/promise-based-modal.component.ts
ezequielc: @alex-deaconu Interesting point. I will think more about it and revisit it. Thanks

src/app/components/modals/promise-based-modal.component.ts
ezequielc: @alex-deaconu Ok, good point. I will rename it. Thanks

ezequielc:
@alex-deaconu I addressed your suggestions in the latest commit. I will open new PRs for the remaining work. Thanks

alex-deaconu (approved):
@ezequielc looks good to me, thanks for the changes 👍

PR #2828, opened 2021-07-29
Use attribute for spinner instead of string

alex-deaconu:
## Description
Use attribute for spinner instead of string

Update selectors for presentation and billing

Update all other spinners and e2e selectors
Handle double spinner showing up for slides

[stage-17]

## Motivation and Context
Fix spinner locators for angular e2e

## How Has This Been Tested?
E2e tests pass. Validated that selectors are working correctly.

## Release Plan:

No


alex-deaconu:
There are two issues here:
1. On the local server (`ng serve`), the attribute `ng-reflect-rv-spinner="some-loader"` exists. However, on the deployed stage/prod/test versions (`ng build`) this attribute no longer appears. It is removed by compilation. Setting the `rvSpinner="some-loader"` property directly (as opposed to `[rvSpinner]` ensures the property remains on the element.

2. The selector picks the parent element, which is usually a parent container on the page, and does not actually disappear. In ajs we would select the child element via `spinner-key`. However, in Angular that is no longer available, so we need to select by the `.spinner-backdrop` class.

Verified that the updated spinner selector works by various methods.

Finally, for the Slides component there is a double spinner being shown when setting the attribute. Because we update the input with `randomSlideId2`, it updates that to a URL, sets the `url` which triggers the `ngModel`. Then it runs the validation of the URL. However, once the `ngModel` debouncer completes it triggers a second validation and a second spinner.

ezequielc (approved):

PR #2851, opened 2021-08-16
Update issue templates

alex-deaconu:
[stage-11]


alex-deaconu:
The configurations at the top are part of the template. I'm assuming they would not show when the template is active, we can validate that once released.

Also, there's the ability to add a default title, which we don't have defined, but it may be useful.

ezequielc (approved):

